PR #110, opened 2025-03-16
Fix call get_all_fields API in get_fields

akihiro-fukuchi:
## WHAT

Call [get_all_fields()](https://github.com/atlassian-api/atlassian-python-api/blob/df25ee599ed61663b667b7d3b595111a7ab01c1d/atlassian/jira.py#L688) instead of `fields()`

## WHY

For example, when using the `jira_create_issue` tool with issue_type `Sub-task`, we get an error for something like this.

`ERROR:mcp-jira:Error getting Jira fields: 'Jira' object has no attribute 'fields'`


sooperset:
Great catch! thanks for the contribution

Issue #114, opened 2025-03-17
No way to retrieve custom fields for a Jira issue


iamjackg:
There doesn't seem to be a way to retrieve any issue fields beyond the ones hardcoded in the MCP server. It would be awesome if there was a way to specify extra fields to retrieve, perhaps through an `extra_fields` parameter in `jira_get_issue`.

Amazing project regardless!

sooperset:
Thanks for reporting. I'll look into this

sooperset:
As I currently don't have a test environment set up for different fields, I welcome PRs from those interested in this matter or if you could provide specific test cases, I'll look into them.

vivekvasani:
@sooperset I took a stab at quickly vibe-coding a solution that worked to resolve the issue I was having. I'm happy to change the merge base to a new branch in your repo so you can take over. I unfortunately won't have much time to dig deeper at the larger ask until late April.

sooperset:
Thanks for the contirubtion. hope @vivekvasani 's PR #117  solves this issue. if you have any further problem, please reopen this issue!

iamjackg:
That was quick! This is a step towards solving this issue, but there is still no way for me as a user to access the custom fields. I might take a stab at that later.

TheConnMan:
I'm still not able to query custom fields either by name or ID. I've pulled the latest and cleared my UV cache. Any help is appreciated.

sooperset:
Hi @TheConnMan, sorry to hear you're still having issues with custom fields!
Let's try to debug this together. I need some more details about your specific scenario:

1. What custom fields are you trying to access? (name and/or ID)
2. How are you currently trying to access them? (with `jira_get_issue` or another function?)
3. Can you share an example of your issue data structure with sensitive info redacted?

sooperset:
@TheConnMan Thanks for the detailed explanation. I'm working on this in #157

lawwu:
I'm interested in this functionality. Right now with the 0.3.1 release I can fetch custom fields with `get_issue`:

```
jira_fetcher.get_issue("ISSUE_NO")

# the custom fields show up like this
custom_fields': {'customfield_21400': None,
  'customfield_15800': None,
  'customfield_15801': None,
  'customfield_10900': None,
...
```

[...]

PR #117, opened 2025-03-18
Add dynamic custom field retrieval in JiraIssue

vivekvasani:
- Add _find_custom_field_by_name method to dynamically find custom field values by name patterns, improving adaptability to different Jira configurations.
- Update epic field extraction in from_api_response to utilize the new method for better handling of varying custom field IDs.
- Enhance unit tests to cover various scenarios for custom field retrieval, including case insensitivity and pattern matching.

Contribute to the resolution of #114

### DIFF
--- a/src/mcp_atlassian/models/jira.py
+++ b/src/mcp_atlassian/models/jira.py
@@ -406,6 +406,76 @@ class JiraIssue(ApiModel, TimestampMixin):
         )
         return self.description
 
+    @staticmethod
+    def _find_custom_field_by_name(fields: dict[str, Any], name_patterns: list[str]) -> Any:
+        """
+        Find a custom field value by searching through field names.
+        
+        Since Jira instances can have different custom field IDs for the same
+        field, this method dynamically discovers fields by their names using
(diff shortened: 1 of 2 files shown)


src/mcp_atlassian/models/jira.py
```diff
@@ -406,6 +406,76 @@ def page_content(self) -> str | None:
         )
         return self.description
 
+    @staticmethod
+    def _find_custom_field_by_name(fields: dict[str, Any], name_patterns: list[str]) -> Any:
```
vivekvasani, line 410: name_patterns as a list of string to match is probably overkill, but I left what Claude gave as-is. But let me know if we want to simplify to a single string input

sooperset:
@vivekvasani Thanks for the great PR addressing the custom fields issue! I noticed a pre-commit error when testing your branch. Could you please fix those when you get a chance? I'll merge whenever it's done. Really appreciate your quick contribution on this!

sooperset:
This is trivial so I'll handle it! Thank you again

PR #118, opened 2025-03-18
fix: in confluence, no-ssl-verify doesn't have any effect issue while…

nolleh:
… fetch space

### PROBLEM 
In mcp-attlanssian, there is a command line argument to handle no-*-ssl-verification (in env, CONFLUENCE_SSL_VERIFY), 
for those who use atlassian is installed in there system with self-signed certificates.

But eventhough you set that value as true,
```
--no-confluence-token,
...
# CONFLUENCE_SSL_VERIFY=false                                # Set to 'false' for self-signed certificates
```

when fetch to space, error occured.

### FIX
Made in server.py file,
added overriding env set for `requests`, if the flag enabled.
(I didn't added this logic to jira, because there isn't any prob in jira to use now. - I don't know why. -)

### DIFF
--- a/src/mcp_atlassian/server.py
+++ b/src/mcp_atlassian/server.py
@@ -88,7 +88,7 @@ async def server_lifespan(server: Server) -> AsyncIterator[AppContext]:
             logger.info(f"Confluence URL: {confluence_url}")
             if not confluence.config.ssl_verify:
                 logger.warning(
-                    "SSL verifiaction is disabled for Confluence. This may be insecure."
+                    "SSL verification is disabled for Confluence. This may be insecure."
                 )
 
                 original_merge_settings = requests.Session.merge_environment_settings
--- a/src/mcp_atlassian/server.py
+++ b/src/mcp_atlassian/server.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 from typing import Any
 from urllib.parse import urlparse
 
+import requests
 from mcp.server import Server
 from mcp.types import Resource, TextContent, Tool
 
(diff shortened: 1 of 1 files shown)


src/mcp_atlassian/server.py
```diff
@@ -75,6 +77,11 @@ async def server_lifespan(server: Server) -> AsyncIterator[AppContext]:
     # Get available services
     services = get_available_services()
 
+    def merge_environment_settings(self, url, proxies, stream, verify, cert):
+        settings = old_merge_enviornment_settings(self, url, proxies, stream, verify, cert)
```
sooperset, line 81: Typo in `old_merge_enviornment_settings`

src/mcp_atlassian/server.py
```diff
@@ -75,6 +77,11 @@ async def server_lifespan(server: Server) -> AsyncIterator[AppContext]:
...
+        settings['verify'] = False
```
sooperset, line 82: The function forces `verify=False` for all requests - is that intentional?

src/mcp_atlassian/server.py
```diff
@@ -84,6 +91,12 @@ async def server_lifespan(server: Server) -> AsyncIterator[AppContext]:
         logger.info("Starting MCP Atlassian server")
         if confluence:
             logger.info(f"Confluence URL: {confluence.config.url}")
+            if not confluence.config.ssl_verify:
+                old_merge_enviornment_settings = requests.Session.merge_environment_settings
+                logger.warning(
+                    "SSL verifiaction is disabled for Confluence. This may be insecure."
```
sooperset, line 97: typo: verifiaction might be verification?

src/mcp_atlassian/server.py
```diff
@@ -84,6 +91,12 @@ async def server_lifespan(server: Server) -> AsyncIterator[AppContext]:
...
+                )
+                requests.Session.merge_environment_settings = merge_environment_settings
```
sooperset, line 99: This approach modifies the global requests behavior, which affects all HTTP calls, not just Confluence
We're never restoring the original method, which could be problematic for long-running servers

sooperset (commented):
Thanks for your contribution! This PR fixes an important issue with SSL verification.
The global monkey-patching approach works but is a bit heavy-handed. Would it be possible to just set `verify=False` directly on the Confluence client's session instead?

src/mcp_atlassian/server.py
nolleh: bulls eye !!

src/mcp_atlassian/server.py
nolleh: I've modifed that confusing code more readable

src/mcp_atlassian/server.py
nolleh: Yeap, you are totally right. but the original code already set perfectly ssl_verify options for session, 
So I was have trouble to figure out why is that, (ssl selfsign - maximum tried issue ocurred..) this code is like.. biting the bullet.

after read your comment, I've checked once more to find out 
1. whether atlassian-python-api has some known issue about disabling ssl-verify. (no vail)
2. Tried to add confluence.confluece.session.verify = false (I've seen vaild set has no effect)

I added one more commit, 
https://github.com/sooperset/mcp-atlassian/pull/118/commits/bccf6c613193f530c6f09d2116bea9e6725dd103
to check only monkey patch applied to on confluence by plain url check, 

But this is along with same line with symptomatic therapy, so feel free to change or closing the PR. 
(and if other module - altassian-pyhon-api?- has related issue, then without changing it will come be okay after times pass..)

src/mcp_atlassian/server.py
nolleh: Add some barrier. 
https://github.com/sooperset/mcp-atlassian/pull/118/commits/bccf6c613193f530c6f09d2116bea9e6725dd103#diff-5aed7a9c7a1edd4414259c5f03467a28163473241789f3465bcae85d96ef5d41R108

sooperset:
Thanks for spotting this SSL verification issue with Confluence servers!
I like your approach, but I've tweaked the implementation to make it more maintainable:

1. Switched from monkey patching to transport adapters (the standard way to customize requests)
2. Fixed the URL matching to handle different formats consistently
Same solution at heart, just structured a bit cleaner. I've merged these changes with your PR.

Thanks for your contribution! 👍

PR #130, opened 2025-03-20
Support getting raw page format

sjanota:
I've noticed the macros (like datetime) are not visible in the markdown when using get_page_content. I added a new argument that allows to disable converting to markdown and see the page in it's raw format.

### DIFF
--- a/src/mcp_atlassian/server.py
+++ b/src/mcp_atlassian/server.py
@@ -46,8 +46,7 @@ def get_available_services() -> dict[str, bool | None]:
             confluence_vars = all(
                 [confluence_url, os.getenv("CONFLUENCE_PERSONAL_TOKEN")]
             )
-            logger.info(
-                "Using Confluence Server/Data Center authentication method")
+            logger.info("Using Confluence Server/Data Center authentication method")
     else:
         confluence_vars = False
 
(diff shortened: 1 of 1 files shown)


src/mcp_atlassian/server.py
```diff
@@ -347,6 +347,11 @@ async def list_tools() -> list[Tool]:
                                 "description": "Whether to include page metadata such as creation date, last update, version, and labels",
                                 "default": True,
                             },
+                            "markdown": {
```
sooperset, line 350:
```suggestion
                            "convert_to_markdown": {
```
Consider renaming the parameter from `markdown` to something more descriptive like `convert_to_markdown` (to match the internal parameter) or `use_markdown_format` to make its purpose clearer.

src/mcp_atlassian/server.py
```diff
@@ -347,6 +347,11 @@ async def list_tools() -> list[Tool]:
...
+                                "type": "boolean",
+                                "description": "Whether to convert page to markdown or keep it in raw HTML format. Can help reaviling macros (like dates) and other content not visible in markdown",
```
sooperset, line 352:
```suggestion
                                "description": "Whether to convert page to markdown (true) or keep it in raw HTML format (false). Raw HTML can reveal macros (like dates) not visible in markdown, but CAUTION: using HTML significantly increases token usage in AI responses.",
```
I'd recommend updating the parameter description to include a caution about token usage. These changes would help users understand the tradeoff between seeing the full content with macros versus keeping token usage lower.

sooperset (commented):
Thanks @sjanota for this contribution! This is definitely a useful addition that will help users see Confluence macros that aren't visible in the markdown conversion.

I've noticed the linting check is failing on the build. Could you fix those issues to help us merge this PR?

src/mcp_atlassian/server.py
sjanota: Good idea, will change that.

sooperset:
LGTM! 👍

sooperset (approved):

sjanota:
I fixed the signature. Looks ready to go now 👍🏻

PR #133, opened 2025-03-20
Add support for jira/confluence search filter

lonelymoon87:
Add confluence-spaces and jira-projects filter arguments,
these filters help narrow down search results to the most relevant spaces or projects, improving response quality and performance.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,26 +10,24 @@
 CONFLUENCE_URL=https://your-domain.atlassian.net/wiki
 CONFLUENCE_USERNAME=your.email@domain.com
 CONFLUENCE_API_TOKEN=your_api_token
-
-# Optional Confluence filters (CLI: --confluence-spaces-filter)
 # CONFLUENCE_SPACES_FILTER=DEV,TEAM,DOC                # Optional: Filter Confluence spaces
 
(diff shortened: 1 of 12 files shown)


.env.example
```diff
@@ -5,6 +5,7 @@
 CONFLUENCE_URL=https://your-domain.atlassian.net/wiki  # Your Confluence cloud URL
 CONFLUENCE_USERNAME=your.email@domain.com              # Your Atlassian account email
 CONFLUENCE_API_TOKEN=your_api_token                    # API token for Confluence
+CONFLUENCE_SPACES_FILTER=DEV,TEAM,DOC                  # Optional: Comma-separated list of space keys to filter search results
```
sooperset, line 8: I think it would be clearer to comment out the optional filters since they're not required for basic setup

sooperset (commented):
Thanks for this PR! The filtering feature looks very useful for large Atlassian environments. I left minor comment. and could you add some test cases for these new filtering features?
Otherwise, the implementation looks clean and well-documented. This will definitely help users get more relevant results!

.env.example
lonelymoon87: Fixed, comment out optinal items.

sooperset (approved):
LGMT! Great contribution

PR #157, opened 2025-03-26
Add support for retrieving Jira custom fields via MCP API

sooperset:
## Summary
This PR adds support for retrieving custom fields and specific fields from Jira issues through the MCP Atlassian integration, addressing issue #114. 

## Changes
- Added `fields`, `properties`, and `update_history` parameters to Jira get_issue methods
- Enhanced the JiraIssue model to handle custom fields properly
- Updated server.py to expose these parameters through the MCP API
- Added support for different formats of the fields parameter (string, list, tuple, set)
- Added comprehensive tests for the new functionality
- Updated existing tests to work with the new parameters

## Testing
- Unit tests added for all new functionality
- Added a real API validation test `test_jira_get_issue_with_fields`
- Manually tested with various custom fields and field formats

## Usage Examples
Users can now retrieve custom fields in several ways:

Fixes #114

PR #162, opened 2025-03-27
Parse fixVersions field properly under "fields"

bstasz-graebert:
It seems this field could be parsed too, as it is not custom field.

@sooperset I was thinking https://github.com/sooperset/mcp-atlassian/pull/157 will resolve this problem that I had, but it seems it parses custom fields only, no? Correct me if I'm wrong.

Anyways. I've added this code to parse this missing field, but maybe it would be good if there would be some refactor so instead of fetching all the predefined fields one-by-one it could just save the whole API call and fetch it depending on which `field` under `fields` user requests? Then it'd avoid adding more fields one-by-one in the future.

LMK what you think @sooperset

### DIFF
--- a/src/mcp_atlassian/models/jira.py
+++ b/src/mcp_atlassian/models/jira.py
@@ -470,6 +470,7 @@ class JiraIssue(ApiModel, TimestampMixin):
     url: str | None = None
     epic_key: str | None = None
     epic_name: str | None = None
+    fix_versions: list[str] = Field(default_factory=list)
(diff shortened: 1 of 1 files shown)


sooperset (approved):
I think that's a good idea! This approach would eliminate the need for manual parsing. Would you be able to take on this task in the future?

bstasz-graebert:
@sooperset we will see, but I could try in one of the next PRs that I would do, but hard to say if I will necessarily have time. Other than that I've noted it :)

sooperset:
I did! Thanks for the contribution

PR #168, opened 2025-03-28
Implement possibility to upload attachments to JIRA

bstasz-graebert:
As the title says - Added attachment upload capability to Jira issues:

1. Upload attachments to a Jira issue: jira_update_issue --issue_key "PROJ-123" --attachments "/path/to/file1.txt,/path/to/file2.pdf"

2. Upload attachments while updating other fields: jira_update_issue --issue_key "PROJ-123" --fields '{"summary": "Updated title"}' --attachments "/path/to/file.txt"

### DIFF
--- a/src/mcp_atlassian/server.py
+++ b/src/mcp_atlassian/server.py
@@ -778,6 +778,11 @@ async def list_tools() -> list[Tool]:
                                     "description": "Optional JSON string of additional fields to update. Use this for custom fields or more complex updates.",
                                     "default": "{}",
                                 },
+                                "attachments": {
+                                    "type": "string",
+                                    "description": "Optional JSON string or comma-separated list of file paths to attach to the issue. "
+                                    'Example: "/path/to/file1.txt,/path/to/file2.txt" or "["/path/to/file1.txt","/path/to/file2.txt"]"',
+                                },
                             },
                             "required": ["issue_key", "fields"],
                         },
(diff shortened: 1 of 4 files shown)


src/mcp_atlassian/server.py
```diff
@@ -1445,7 +1450,37 @@ def format_comment(comment: Any) -> dict:
...
+                            attachments = parsed_attachments
+                        else:
+                            # Single file path as a JSON string
+                            attachments = [parsed_attachments]
+                    except json.JSONDecodeError:
+                        # Plain string - single file path
+                        attachments = [arguments.get("attachments")]
+                elif isinstance(arguments.get("attachments"), list):
+                    # Already a list
+                    attachments = arguments.get("attachments")
```
sooperset, line 1471: The schema description for the `attachments` parameter mentions it can be a "JSON string or comma-separated list of file paths". However, the parsing logic here only explicitly handles a JSON array (via `json.loads`) or potentially a single path provided as a plain string (handled by the `except json.JSONDecodeError:`), or a python list. It doesn't explicitly handle a comma-separated string like `"/path/one,/path/two"`.

Could you please clarify the intended input format(s)?
- If comma-separated strings *should* be supported, please add logic here (e.g., an `elif isinstance(attachments_arg, str) and ',' in attachments_arg:` block with `attachments = attachments_arg.split(',')`).
- If only JSON arrays or single plain strings are intended, please update the schema description to remove the "comma-separated list" part for clarity.

This needs to be consistent before merging to ensure users provide arguments in a format the server can actually parse as intended.

tests/unit/jira/test_attachments.py
```diff
@@ -395,3 +408,173 @@ def test_download_issue_attachments_missing_url(self):
...
+            mock_abspath.return_value = "/absolute/path/test_file.txt"
+            mock_basename.return_value = "test_file.txt"
+
+            result = self.client.upload_attachment(
+                "TEST-123", "/absolute/path/test_file.txt"
+            )
+
+            # Assertions
+            assert result["success"] is False
+            assert "Failed to upload attachment" in result["error"]
```
sooperset, line 580: Unit tests are missing for the `upload_attachments` (plural) method introduced in `src/mcp_atlassian/jira/attachments.py`.

Please add tests covering at least these scenarios:
- Successful upload of multiple files.
- A mix of successful and failed uploads (e.g., one file exists, one doesn't) to test error aggregation in the returned dictionary.
- Calling the method with an empty list of file paths.
- Calling the method with no issue key provided.

sooperset (changes requested):
This is a great addition, @bstasz-graebert! Uploading attachments is a much-needed feature. The core logic in `AttachmentsMixin` looks solid, and the integration into `IssuesMixin` seems reasonable.

Two main points still need addressing before this can be merged safely:
1.  The argument parsing for `attachments` in `server.py` needs clarification and potential correction to match the tool schema description.
2.  Unit tests are needed for the `upload_attachments` (plural) method.

Details below. Thanks for the contribution!

tests/unit/jira/test_attachments.py
bstasz-graebert: Done.

src/mcp_atlassian/server.py
bstasz-graebert: Done.

sooperset (approved):
Everything looks good now. 👍  Thanks again for this valuable contribution!

PR #173, opened 2025-03-31
On prem confluence with basic auth is also made available

jeteve:
This will allow basic auth parameters to be used with on-prem confluence

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,7 +14,15 @@ CONFLUENCE_API_TOKEN=your_api_token
 
 # Confluence Server/Data Center (CLI: --confluence-url, --confluence-personal-token, --[no-]confluence-ssl-verify)
 # CONFLUENCE_URL=https://confluence.your-company.com
+
+# For confluence server using a personal access token:
 # CONFLUENCE_PERSONAL_TOKEN=your_personal_access_token
+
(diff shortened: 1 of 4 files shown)


sooperset (commented):
Thanks for this contribution, @jeteve! Adding flexibility for authenticating with on-prem Confluence via Username/API Token is a useful addition.

The change in `get_available_services` looks logically correct for detecting the necessary environment variables.

Before this can be merged, could you please:

1.  Add unit tests specifically for the `get_available_services` function. These tests should cover:
    *   The new on-prem User/Token configuration scenario.
    *   The existing on-prem PAT and Cloud configuration scenarios to check for regressions.
    *   Cases where services are *not* configured.
    *(These tests would likely be the first for this specific function, helping to build coverage.)*
2.  Update Documentation:
    *   Update `README.md` to explain both authentication methods (PAT and User/Token) for Confluence Server/DC.
    *   Update `.env.example` to include `CONFLUENCE_USERNAME`/`CONFLUENCE_API_TOKEN` as a commented-out alternative for Server/DC.

Adding these tests will verify the new logic and safeguard existing configurations. Once the tests and docs are added, this should be ready. Thanks!

jeteve:
Thanks @sooperset , I've made the requested additions.

PR #178, opened 2025-04-01
New tool : Attach content to confluence page

felixbouazza:
Allow LLM to attach content to confluence page via new tool "confluence_attach_content"

### DIFF
--- a/src/mcp_atlassian/confluence/pages.py
+++ b/src/mcp_atlassian/confluence/pages.py
@@ -478,11 +478,6 @@ class PagesMixin(ConfluenceClient):
             raise ConfluenceGetAttachmentsFromContentError(
                 f"Error when trying to connect to Confluence: {str(e)}"
             ) from e
-        except Exception as e:
-            logger.error(f"Unexpected error: {str(e)}")
-            raise ConfluenceGetAttachmentsFromContentError(
-                f"Unexpected error when trying to get attachments from page {page_id}: {str(e)}"
-            ) from e
 
     def attach_content(self, content: bytes, name: str, page_id: str) -> None:
         """
@@ -508,9 +503,4 @@ class PagesMixin(ConfluenceClient):
             logger.error(f"Network error: {e}")
             raise ConfluenceAttachContentError(
                 f"Error when trying to connect to Confluence: {str(e)}"
-            ) from e
-        except Exception as e:
-            logger.error(f"Unexpected error: {str(e)}")
-            raise ConfluenceAttachContentError(
-                f"Unexpected error when trying to attach content to page {page_id}: {str(e)}"
             ) from e
--- a/src/mcp_atlassian/confluence/pages.py
+++ b/src/mcp_atlassian/confluence/pages.py
@@ -449,14 +449,16 @@ class PagesMixin(ConfluenceClient):
             name: The name of the attachment
             page_id: The ID of the page to attach the content to
 
+        Returns:
+            ConfluencePage model containing the updated page's data or None if an error occurred
         """
         try:
-            logger.debug("Attaching content to page %s", page_id)
+            logger.debug("Attaching content %s to page %s", name, page_id)
             self.confluence.attach_content(content=content, name=name, page_id=page_id)
         except ApiError as e:
-            logger.error("Confluence API Error: %s", str(e))
+            logger.error("Confluence API Error when trying to attach content %s to page %s: %s", name, page_id, str(e))
             return None
         except RequestException as e:
-            logger.error("Network error: %s", str(e))
+            logger.error("Network error when trying to attach content %s to page %s: %s", name, page_id, str(e))
             return None
         return self.get_page_content(page_id=page_id, convert_to_markdown=False)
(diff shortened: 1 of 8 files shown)


src/mcp_atlassian/confluence/pages.py
```diff
@@ -435,3 +435,45 @@ def delete_page(self, page_id: str) -> bool:
...
+        """
+        Attach content to a Confluence page.
+
+        Args:
+            content: The content to attach (bytes)
+            name: The name of the attachment
+            page_id: The ID of the page to attach the content to
+
+        Returns:
+            Boolean indicating success (True) or failure (False)
```
lolrenx, line 449: what would be the use case for return `False` and not raising an exception?

felixbouazza:
@lolrenx @sooperset Sorry guys, there is a lot of changes since your last reviews. First, I've improved error handling on attach content tool, thanks for your comments. Second, @lolrenx, your comments https://github.com/sooperset/mcp-atlassian/pull/178#discussion_r2023052509 which propose to return Confluence Page was interesting but there isn't data about attachments in Confluence Page yet, so I've decided to implement the `get_attachments_from_content`, first as a tool in the mcp-server and second as return value for `attach_content` which allow LLMs to know if action has worked.

src/mcp_atlassian/confluence/pages.py
felixbouazza: So I've included Attachments on Confluence Page and I've returned the related Page object after attaching content.

src/mcp_atlassian/confluence/pages.py
```diff
@@ -435,3 +437,38 @@ def delete_page(self, page_id: str) -> bool:
...
+            logger.debug("Attaching content %s to page %s", name, page_id)
+            self.confluence.attach_content(content=content, name=name, page_id=page_id)
+        except ApiError as e:
+            logger.error(
+                "Confluence API Error when trying to attach content %s to page %s: %s",
+                name,
+                page_id,
+                str(e),
+            )
+            return None
```
sooperset, line 465: Returning `None` here means the `call_tool` function only gets a generic failure signal when this specific `ApiError` occurs.
Consider letting this `ApiError` propagate upwards by using `raise` instead of `return None`.

src/mcp_atlassian/confluence/pages.py
```diff
@@ -435,3 +437,38 @@ def delete_page(self, page_id: str) -> bool:
...
+        except RequestException as e:
+            logger.error(
+                "Network error when trying to attach content %s to page %s: %s",
+                name,
+                page_id,
+                str(e),
+            )
+            return None
```
sooperset, line 473: this too

src/mcp_atlassian/server.py
```diff
@@ -1167,6 +1190,54 @@ def format_comment(comment: Any) -> dict:
...
+                        type="text",
+                        text="Error: Missing required parameters: content, name, and page_id are required.",
+                    )
+                ]
+
+            page = ctx.confluence.attach_content(
+                content=content, name=name, page_id=page_id
+            )
+            if page is None:
+                logger.error("Error attaching content to Confluence page %s", page_id)
```
sooperset, line 1221: If `attach_content` is changed to `raise` specific exceptions (like `ApiError`, `RequestException`) instead of returning `None` on failure, this `if page is None:` check should be replaced with corresponding `try...except` blocks to handle those specific errors and format better feedback.

tests/unit/confluence/test_pages.py
```diff
@@ -386,6 +395,69 @@ def test_delete_page_error(self, pages_mixin):
...
+        exception_message = "Network error"
+        pages_mixin.confluence.attach_content.side_effect = RequestException(
+            exception_message
+        )
+
+        # Act/Assert
+        result = pages_mixin.attach_content(content=content, name=name, page_id=page_id)
+        assert result is None
+
+    def test_attach_content_unexpected_error(self, pages_mixin):
```
sooperset, line 447: Consider renaming this test to `test_attach_content_unhandled_exception_propagates` for slightly better clarity.

sooperset (commented):
Thanks for the updates and addressing the feedback from the previous round, @felixbouazza! This is looking much better. I left a few comments.

tests/unit/confluence/test_pages.py
```diff
@@ -386,6 +395,70 @@ def test_delete_page_error(self, pages_mixin):
...
+        exception_message = "Network error"
+        pages_mixin.confluence.attach_content.side_effect = RequestException(
+            exception_message
+        )
+
+        # Act/Assert
+        with pytest.raises(RequestException, match=exception_message):
+            pages_mixin.attach_content(content=content, name=name, page_id=page_id)
+
+    def test_attach_content_unhandled_exception_propagate(self, pages_mixin):
```
lolrenx, line 447:
```suggestion
    def test_attach_content_unhandled_exception_propagates(self, pages_mixin):
```

src/mcp_atlassian/confluence/pages.py
```diff
@@ -435,3 +437,38 @@ def delete_page(self, page_id: str) -> bool:
...
+        """
+        Attach content to a Confluence page.
+
+        Args:
+            content: The content to attach (bytes)
+            name: The name of the attachment
+            page_id: The ID of the page to attach the content to
+
+        Returns:
+            ConfluencePage model containing the updated page's data or None if an error occurred
```
sooperset, line 453:
```suggestion
            ConfluencePage model containing the updated page's data
```

sooperset (approved):
LGTM!

PR #180, opened 2025-04-01
Fix #172: Quote user identifier in list_resources JQL query

### LINKED ISSUE #172: Error fetching Jira projects caused by character '@'
I am encountering the following errors when I use the latest version of mcp-atlassian.

Install:
brew install uv
uvx mcp-atlassian

Usage:
Transport Type: STDIO
Command: uvx
Arguments: mcp-atlassian --confluence-url my_url --confluence-personal-token my_token_1 --jira-url my_url --jira-personal-token my_token_2

Error output from MCP server when connecting:
`ERROR - mcp-atlassian - Error fetching Jira projects: Error in the JQL Query: The character '@' is a reserved JQL character. You must enclose it in a string or use the escape '\u0040' instead. (line 1, character 17)`

Seems main features are still correct, I can invoke the tools in windsurf. please help to address the above error.

sooperset:
Closes #172

Fixes a JQL syntax error in `list_resources` caused by unquoted user identifiers (like emails containing '@') during startup. The `account_id` variable is now quoted in the JQL query in `src/mcp_atlassian/server.py`.

### DIFF
--- a/src/mcp_atlassian/server.py
+++ b/src/mcp_atlassian/server.py
@@ -12,6 +12,7 @@ from mcp.types import Resource, TextContent, Tool
 
 from .confluence import ConfluenceFetcher
 from .jira import JiraFetcher
+from .jira.utils import escape_jql_string
 from .utils.io import is_read_only_mode
 from .utils.urls import is_atlassian_cloud_url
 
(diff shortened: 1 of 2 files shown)


src/mcp_atlassian/server.py
```diff
@@ -150,7 +150,7 @@ async def list_resources() -> list[Resource]:
             account_id = ctx.jira.get_current_user_account_id()
 
             # Use JQL to find issues the user is assigned to or reported
-            jql = f"assignee = {account_id} OR reporter = {account_id} ORDER BY updated DESC"
+            jql = f'assignee = "{account_id}" OR reporter = "{account_id}" ORDER BY updated DESC'
```
jeteve, line 153: Maybe generate the right escaping? See https://confluence.atlassian.com/jiracoreserver081/search-syntax-for-text-fields-970610224.html 

Or is it not needed because account_id will always be safe?

sooperset: @jeteve Good point! You're right, it's more robust, especially considering potential Server/DC key/name values which might contain special characters. I'll update the PR to handle escaping \ and " correctly before quoting the account_id. Thanks for the catch!

Issue #176, opened 2025-04-01
[Feature Request] Add non-custom fields property/method to JiraIssue


cutekibry:
Some of the fields are non-custom, e.g. `/rest/api/3/field`:

```json
[
  {
    "id": "timespent",
    "key": "timespent",
    "name": "Spent time",
    "custom": false,
    "orderable": false,
    "navigable": true,
    "searchable": false,
    "clauseNames": [
      "timespent"
    ],
    "schema": {
      "type": "number",
      "system": "timespent"
    }
  }
]
```

They doesn't seem to appear in `JiraIssue`, but exist in api calling such as `/rest/api/3/issue/<issue_key>` and the result of `JiraClient.jira.jql()`. We can gather the data using code like

```python
# self extends from JiraClient
response = self.jira.jql(jql, limit=limit)
issue = JiraIssue.from_api_response(issue_data, base_url=self.config.url)
issue.custom_fields["duedate"] = issue_data.get("fields", {}).get("duedate", "")
```

I am new to Atlassian and `mcp-atlassian`, so I don't know if it's implemented or not.

If not, you can tell me how can I add the feature and I may have time to implement it and send a PR.

sooperset:
Sure, as we discussed in #162 , we should definitely look into this! Thanks for the request.

PR #182, opened 2025-04-02
Improve custom_fields handling to include all unprocessed fields

WingCH:
https://github.com/sooperset/mcp-atlassian/issues/114

This PR updates the `JiraIssue` model to capture all unprocessed fields into `custom_fields`.  

- Skips known fields
- Ensures support for fields like `timetracking`, `timeestimate`, or any custom fields  
- Improves compatibility with varying Jira field configurations

### DIFF
--- a/src/mcp_atlassian/models/jira.py
+++ b/src/mcp_atlassian/models/jira.py
@@ -735,6 +735,20 @@ class JiraIssue(ApiModel, TimestampMixin):
             fields, ["Epic Name", "epic-name", "epicname"]
         )
 
+        # Extract all fields that aren't explicitly processed into custom_fields
+        known_fields = {
+            "summary", "description", "status", "issuetype", "priority", 
+            "assignee", "reporter", "labels", "components", "comment", 
+            "attachment", "created", "updated", "fixVersions"
+        }
+        
+        for key, value in fields.items():
+            # Skip already processed fields and null values
+            if key in known_fields or key.startswith("customfield_") or value is None:
+                continue
+            # Add all other fields to custom_fields
+            custom_fields[key] = value
+            
         return cls(
             id=issue_id,
             key=key,
@@ -876,8 +890,8 @@ class JiraIssue(ApiModel, TimestampMixin):
                     value = field_mapping[field]()
                     if value is not None:  # Only include non-None values
                         result[field] = value
-                # Handle custom fields
-                elif field.startswith("customfield_") and field in self.custom_fields:
+                # All other fields are custom fields
+                elif field in self.custom_fields:
                     result[field] = self.custom_fields[field]
 
         # Case 3: No specific fields requested - use essential fields
--- a/src/mcp_atlassian/models/jira.py
+++ b/src/mcp_atlassian/models/jira.py
@@ -737,18 +737,29 @@ class JiraIssue(ApiModel, TimestampMixin):
 
         # Extract all fields that aren't explicitly processed into custom_fields
         known_fields = {
-            "summary", "description", "status", "issuetype", "priority", 
-            "assignee", "reporter", "labels", "components", "comment", 
-            "attachment", "created", "updated", "fixVersions"
+            "summary",
+            "description",
+            "status",
+            "issuetype",
+            "priority",
+            "assignee",
+            "reporter",
+            "labels",
+            "components",
+            "comment",
+            "attachment",
+            "created",
+            "updated",
+            "fixVersions",
         }
-        
+
         for key, value in fields.items():
             # Skip already processed fields and null values
             if key in known_fields or key.startswith("customfield_") or value is None:
                 continue
             # Add all other fields to custom_fields
             custom_fields[key] = value
-            
+
         return cls(
             id=issue_id,
             key=key,
--- a/src/mcp_atlassian/models/jira.py
+++ b/src/mcp_atlassian/models/jira.py
@@ -755,7 +755,7 @@ class JiraIssue(ApiModel, TimestampMixin):
 
         for key, value in fields.items():
             # Skip already processed fields and null values
-            if key in known_fields or key.startswith("customfield_") or value is None:
+            if key in known_fields or value is None:
                 continue
             # Add all other fields to custom_fields
             custom_fields[key] = value


src/mcp_atlassian/models/jira.py
```diff
@@ -876,8 +901,8 @@ def to_simplified_dict(self) -> dict[str, Any]:
                     value = field_mapping[field]()
                     if value is not None:  # Only include non-None values
                         result[field] = value
-                # Handle custom fields
-                elif field.startswith("customfield_") and field in self.custom_fields:
+                # All other fields are custom fields
+                elif field in self.custom_fields:
```
sooperset, line 905: Good addition 👍. This ensures that fields captured into `self.custom_fields` during parsing can be correctly retrieved in the output when requested by name.

src/mcp_atlassian/models/jira.py
```diff
@@ -735,6 +735,31 @@ def from_api_response(cls, data: dict[str, Any], **kwargs: Any) -> "JiraIssue":
...
+            "attachment",
+            "created",
+            "updated",
+            "fixVersions",
+        }
+
+        for key, value in fields.items():
+            # Skip already processed fields and null values
+            if key in known_fields or key.startswith("customfield_") or value is None:
+                continue
```
sooperset, line 759: Could you clarify the thinking behind skipping fields starting with `customfield_...` when populating the `custom_fields` dictionary here in `from_api_response`?

My understanding was that `custom_fields` was intended as a "catch-all" container. I expected it to store *all* fields from the Jira API response that don't map directly to a defined attribute like `summary` or `status`. This would typically include fields named `customfield_12345`, etc. (unless they were already processed, e.g., for Epic linking).

Is there a specific reason why these `customfield_...` fields should be treated differently and excluded from this general collection? Understanding the intent would be helpful!

sooperset (commented):
Thanks, @WingCH! This PR improves how `JiraIssue` handles unprocessed fields (like `timetracking`) by capturing them into `custom_fields`, addressing #114 and increasing robustness. The update to `to_simplified_dict` to retrieve these fields is also a good addition.

My main question is about excluding `customfield_...` prefixes in the `from_api_response` capture loop – could you clarify the intent? (See line comment). Overall, a positive step forward!

src/mcp_atlassian/models/jira.py
WingCH: The reason I skipped customfield_... was to avoid overwriting fields already handled earlier (e.g. Epic linking). I realize this means maintaining the known_fields list manually, which isn’t ideal — but it felt safer than risking duplicates or conflicts. Open to cleaner alternatives if you have any!

WingCH: The reason I skipped customfield_... was to avoid overwriting fields already handled earlier (e.g. Epic linking). I realize this means maintaining the known_fields list manually, which isn’t ideal — but it felt safer than risking duplicates or conflicts. Open to cleaner alternatives if you have any!

sooperset: Okay, thanks for clarifying, @WingCH. Avoiding duplicates for already-processed fields like epics definitely makes sense.

I've been thinking about your comment regarding skipping `customfield_` prefixed fields to avoid conflicts. While I understand the caution, I'm concerned that this approach would exclude most custom fields in Jira, since they typically have IDs like `customfield_10001`.

I'd like to suggest an alternative approach that's more targeted. Instead of excluding all `customfield_` fields, we could track exactly which custom fields we've already processed. Here's how we might implement it:

1. Modify `_find_custom_field_by_name` to return both the field ID and the value:
```python
@staticmethod
def _find_custom_field_by_name(
    fields: dict[str, Any] | None, name_patterns: list[str]
) -> tuple[str | None, Any | None]: # Changed return type
    """Finds a custom field value and its ID by searching field names."""
    if not fields:
        return None, None # Return tuple
    # ... (existing logic) ...
    # Inside the loops where a match is found:
    return field_id, fields[field_id] # Return tuple
    # If no match found
    return None, None
```

2. Track processed custom field IDs in `from_api_response`:
```python
# Keep track of custom field IDs we've already processed
processed_custom_field_ids: set[str] = set()

# Extract epic information and get the field IDs used
epic_key_field_id, epic_key = cls._find_custom_field_by_name(
    fields, ["Epic Link", "epic-link", "epiclink", "Parent Link"]
)
if epic_key_field_id:
    processed_custom_field_ids.add(epic_key_field_id)

epic_name_field_id, epic_name = cls._find_custom_field_by_name(
    fields, ["Epic Name", "epic-name", "epicname"]
)
if epic_name_field_id:
    processed_custom_field_ids.add(epic_name_field_id)

# Combine base known fields with processed custom field IDs
all_known_fields = known_fields.union(processed_custom_field_ids)

# In the loop
for key, value in fields.items():
    # Skip already processed fields and null values
    if key in all_known_fields or value is None:
        continue
    # Add all other fields to custom_fields
    custom_fields[key] = value
```

This approach precisely targets the fields we've already processed, rather than making assumptions based on the `customfield_` prefix. It will correctly include all custom fields we haven't explicitly handled elsewhere.

What do you think about this approach? Would you be willing to update your PR to implement this version instead?

WingCH:
Sorry, I don't fully understand your approach.

Let me explain my understanding and considerations:

1. Current issue: If certain fields (like timetracking) are not explicitly defined in the JiraIssue model, data for these fields gets ignored.

2. I noticed that the current code already has a mechanism to handle fields with the `customfield_` prefix, placing them in the `custom_fields` dictionary.

3. My goal is to extend this mechanism to place all unprocessed fields (not just those with customfield_ prefix) into `custom_fields` as well, to avoid losing any data.

4. To achieve this, I need to exclude fields that have already been processed to avoid duplication or conflicts. That's why I used the condition `if key in known_fields or key.startswith("customfield_") or value is None:` in my code.

However, I just realized there is indeed a logical issue in my implementation - excluding all fields with the `customfield_` prefix would filter out most custom fields. I've fixed this issue in my latest commit, now only excluding known fields rather than excluding based on prefix.

Please check my latest commit to see if it meets your expectations. If you have a better implementation approach,

src/mcp_atlassian/models/jira.py
WingCH: Sorry, I don't fully understand your approach.

Let me explain my understanding and considerations:

Current issue: If certain fields (like timetracking) are not explicitly defined in the JiraIssue model, data for these fields gets ignored.

I noticed that the current code already has a mechanism to handle fields with the customfield_ prefix, placing them in the custom_fields dictionary.

My goal is to extend this mechanism to place all unprocessed fields (not just those with customfield_ prefix) into custom_fields as well, to avoid losing any data.

To achieve this, I need to exclude fields that have already been processed to avoid duplication or conflicts. That's why I used the condition if key in known_fields or key.startswith("customfield_") or value is None: in my code.

However, I just realized there is indeed a logical issue in my implementation - excluding all fields with the customfield_ prefix would filter out most custom fields. I've fixed this issue in my latest commit, now only excluding known fields rather than excluding based on prefix.

Please check my latest commit to see if it meets your expectations. If you have a better implementation approach,

sooperset (approved):
LGTM! 👍 🚀

sooperset:
Closes #176

PR #175, opened 2025-03-31
fix getting resource

jeteve:
Getting resource was crashing before this.

It doesn't crash anymore.

PR #195, opened 2025-04-03
Fix read_resource

dhallas:
The signature of the `read_resource` function should take in a `AnyUrl` and return `str`. Make the needed changes to match this.


dhallas:
This PR is similar to #175 but reduced to only the needed changes for `read_resource` to function correctly.

sooperset (approved):
Thanks for your contribution, @dhallas!  The function now properly takes a Pydantic `AnyUrl` parameter and returns a single string, which aligns with the MCP protocol implementation. Good work on this focused fix.

Issue #205, opened 2025-04-05
Passing A URL Parameter to SSE Transport


maxxrdrgz:
Hello,

I'm trying to setup this MCP server as a remote server using SSE Transport. I'm trying to gauge how feasible it might be for me to modify the code to receive a url parameter from the client, and using that in the JiraFetcher (or the JiraClient) object. For example, in the server.py code, you can certainly parse URL parameters like my example below. But I'm having trouble passing the parameter elsewhere in the code. I'd be happy to contribute if someone else can nudge me in the right directions (or let me know if this is not possible).

The client's url might look like this:
          "url": "http://SERVER:8000/sse?jira_api_token=JIRAPERSONALTOKEN",

```
async def run_server(transport: str = "stdio", port: int = 8000) -> None:
    """Run the MCP Atlassian server with the specified transport."""
    if transport == "sse":
        from mcp.server.sse import SseServerTransport
        from starlette.applications import Starlette
        from starlette.requests import Request
        from starlette.routing import Mount, Route

        sse = SseServerTransport("/messages/")

        async def handle_sse(request: Request) -> None:
            # Extract query parameters
            query_params = request.query_params
            api_key = query_params.get("jira_api_token")  # Get the API key from query parameters

            if not api_key:
                print("API key is missing!")
                return

            print(f"Received API Key: {api_key}")

            async with sse.connect_sse(
                request.scope, request.receive, request._send
            ) as streams:
                await app.run(
                    streams[0], streams[1], app.create_initialization_options()
                )
```

maxxrdrgz:
@sooperset , you're the creator right? If you have some time to nudge me in the right direction I'd be happy to help :-).

I'm more of a javascript guy and so I'm having issues passing the parsed URL parameter outside of the run_server() and into get_available_services().

sooperset:
Hey @maxxrdrgz! I think this should be possible, though I haven't tested it myself. You'd need to:

1. In `server.py`, try modifying `handle_sse()` to add your parameter to the initialization options before passing to `app.run()`

2. Then see if you can access those parameters in `server_lifespan()` through `server.initialization_options` when creating the JiraFetcher

The Server class from MCP might pass those initialization options through to your lifespan function, but you'll need to test if this approach works. If it doesn't, there may be other ways to pass context between these components.

One security consideration - URL params can be visible in logs and browser history, so keep that in mind if you're passing sensitive tokens this way.

maxxrdrgz:
[...]

Issue #221, opened 2025-04-07
Support multi user share same sse server


ttjslbz001:
I need setup a server to support multi user in my company, can we support pass token from cursor client, instead of storing it in server side.

dsaad68:
Hey @sooperset,

I believe it would be beneficial to extend the functionality of the MCP Atlassian to support multiple users. 
This enhancement could significantly improve user experience and scalability.

Proposal:
- Before I begin implementing this feature, I would like to confirm if this aligns with your current roadmap. If it does, I am ready to start the implementation and can also provide additional manpower to assist with development of this feature.
- Provide some guideline for hosting the MCP remotely.

Request:
- Please confirm if supporting multiple users is part of the planned roadmap.

If approved, I will proceed with the implementation and coordinate additional resources as needed.

sooperset:
@ttjslbz001 @dsaad68 
Yes, supporting multiple users is definitely on the roadmap. This would be valuable.

Currently, I'm waiting for full implementation of MCP's streamable HTTP to properly support this use case. I'd be interested in hearing your thoughts on potential implementation approaches that could work with the current specification.

If you're willing to contribute to this feature, that would be fantastic! Feel free to share your implementation ideas or even submit a draft PR so we can collaborate on the best approach.

maxxrdrgz:
see my implementation here:

https://github.com/sooperset/mcp-atlassian/pull/224

sooperset:
Closing this issue as it relates to the broader goal of supporting multiple users/instances with dynamic configurations, which is now being tracked centrally in #231.

Please follow #231 for future updates. Thanks for raising this aspect!

PR #224, opened 2025-04-09
Feature/robust sse transport

maxxrdrgz:
By no means am I expecting you to merge this PR. This is more so my contribution to the community. And it resolves my issue #205 . Where clients can pass url parameters to server side using the same command line arguments already established. 

This offers a clean and simple solution that allows others to host the SSE server with minimal work and respects the jira/confluence URL environment variables that would be given by the host.

To use, on the client side, set the url to 
http://SERVER:8000/sse?jira-personal-token=PERSONAL_TOKEN&confluence-personal-token=PERSONAL_TOKEN

works for both cloud and server

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -70,3 +70,7 @@ JIRA_API_TOKEN=your_api_token
 # JIRA_URL=https://jira.your-company.com
 # JIRA_PERSONAL_TOKEN=your_personal_access_token
 # JIRA_SSL_VERIFY=true                 # CLI: --[no-]jira-ssl-verify
+
+# TLS/SSL
+# SSL_KEY=/some/path/to/sitename.key
+# SSL_CRT=/some/path/to/sitename.crt
--- a/src/mcp_atlassian/server.py
+++ b/src/mcp_atlassian/server.py
@@ -4,13 +4,16 @@ import os
 from collections.abc import AsyncIterator, Sequence
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import Any, TypeVar, cast
 
 from atlassian.errors import ApiError
 from mcp.server import Server
 from mcp.types import Resource, TextContent, Tool
 from pydantic import AnyUrl
 from requests.exceptions import RequestException
+
+from mcp_atlassian.confluence import ConfluenceConfig
+from mcp_atlassian.jira import JiraConfig
 
 from .confluence import ConfluenceFetcher
 from .confluence.utils import quote_cql_identifier_if_needed
@@ -21,6 +24,7 @@ from .utils.urls import is_atlassian_cloud_url
 
 # Configure logging
 logger = logging.getLogger("mcp-atlassian")
+T = TypeVar("T")  # A generic type for the FetcherClass
 
 
 @dataclass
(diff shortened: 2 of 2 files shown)


sooperset (commented):
Thanks for working on multi-user support through URL parameters @maxxrdrgz! This addresses issues #205 and #221, which is great.

I have some concerns with the current implementation that need to be addressed before we can merge:

1. The global `app.user_config` modification is problematic. When multiple users connect simultaneously via SSE (which maintains long-lived connections), each new connection would overwrite the global config. This would cause users to accidentally use each other's credentials. We need a per-connection context approach instead.

2. While convenient, passing tokens via URL parameters exposes them in server logs and browser history. Consider documenting this risk or exploring alternatives like HTTP headers.

3. The PR is currently failing CI checks. These need to be fixed.

I like the direction, but we need to rethink how user credentials are stored and accessed. Instead of modifying a global object, consider using request-specific context.

Let me know if you'd like to discuss implementation approaches - happy to collaborate on finding a solution that maintains good isolation between user sessions.

maxxrdrgz:
1. I'll see what I can do and check if MCP can handle this natively.
2. This is a necessary evil. MCP still hasn't matured enough yet and a lot of clients still do not support standard HTTP options like headers.
3. I'll try and get to this if I fix point number 1.

dsaad68:
@maxxrdrgz @sooperset 
I have a solution for point one, I will create PR tomorrow, I need to test my approach again.

dsaad68:
@maxxrdrgz Great job by the way!

maxxrdrgz:
@dsaad68 whatever gets us there faster my friend. If you get the session stuff figured out, it's likely I'll still integrate the query params because some clients only accept a url for sse transport lol

dsaad68:
@maxxrdrgz added getting config parameters from query parameters. Thanks for the tip.

sooperset:
Hi @maxxrdrgz,

Thanks for your contribution towards multi-user support (#205, #221) and for adding SSL configuration options! We appreciate you taking the initiative to explore solutions, especially the idea of passing authentication via URL parameters for certain client constraints.

However, after careful review, we won't be able to merge this PR in its current state due to a couple of critical reasons:

1.  The primary technical blocker is the use of the global `app.user_config` variable to store user-specific configurations within the `handle_sse` function. In a multi-user scenario with potentially many simultaneous SSE connections, this approach will lead to race conditions where the last connected user's configuration overwrites others, causing unpredictable behavior and users operating under incorrect credentials. A per-request or per-connection context is necessary for safe multi-user handling.
2.  More fundamentally, the Model Context Protocol (MCP) standard itself has evolved. The previous HTTP+SSE transport has been superseded by the new **"Streamable HTTP" transport**, which was recently merged into the official specification ([modelcontextprotocol/modelcontextprotocol#206](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/206)). This new transport offers better scalability, state management, and infrastructure compatibility.
3.  Our plan for `mcp-atlassian` is to align with the official MCP standard and implement multi-user/multi-site support based on this new **Streamable HTTP** transport. However, this is currently blocked pending support for it in the core dependency, `modelcontextprotocol/python-sdk` (Tracked here: [modelcontextprotocol/python-sdk#405](https://github.com/modelcontextprotocol/python-sdk/issues/405)). Implementing multi-user support based on the older SSE transport now would require significant rework once the SDK is updated.

For these reasons, while we value your effort and the ideas presented, we believe the best path forward is to wait for the necessary SDK support for Streamable HTTP and then implement multi-user functionality based on the updated standard.

Thank you again for your contribution and understanding. We hope you'll consider contributing to the Streamable HTTP-based implementation in the future!

Issue #211, opened 2025-04-05
Windsurf unable to parse input schema for jira_create_issue


futhrevo:
Hi
Getting as attached in the image, getting an error like
failed to get tools: failed to parse tool input schema for tool jira_create issue:
json: cannot unmarshal array into Go struct field properties of type string

![Image](https://github.com/user-attachments/assets/6cd57541-419a-4a45-86a8-498b0a2f1688)

sooperset:
Hi @futhrevo, Thanks for reporting this issue. This is indeed related to the recent change in PR #208 where we enhanced Jira's component support to accept multiple components as an array.

While [the MCP specification](https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.json#L1986) does support array arguments in theory, it seems Windsurf is having trouble parsing this particular array input schema. I've confirmed that other clients like Claude Desktop and Cursor IDE handle the tool correctly.

Since Jira component names can't contain commas, we can update the implementation to accept a comma-separated string for components instead of an array.

We can make this change in the next update!

futhrevo:
This error doesn't seem to go away. Still facing same error

sooperset:
@futhrevo Could you share more details about the issue you're still facing? I don't have Windsurf Pro so I can't directly debug this. Are you still seeing the same array parsing error after the fix? Any logs or specific steps to reproduce would help.

sburns:
+1 to @futhrevo, I'm seeing the same error on 0.6.3. I still haven't found where windsurf spits out from mcp servers but this is the error in configuration UI

```
mcp-atlassian
Error: failed to get tools: failed to parse tool input schema for tool jira_create_issue: json: cannot unmarshal array into Go struct field .properties of type string. Check your configuration.
```

If you know of better places to look I'm happy to dig in with you.

sburns:
> Reported this to Windsurf team, awaiting for their response

@futhrevo I haven't reported much to Windsurf, is there a URL to your ticket? Happy to +1 it.

futhrevo:
For now, I have made this change locally as a workaround till actual issue is identified or fixed

![Image](https://github.com/user-attachments/assets/5aacf738-bacb-4fe0-9655-4598638f0131)

sooperset:
[...]

PR #228, opened 2025-04-09
fix: windsurf parse error

BEMELON:
I'm not completely confident if this is the appropriate approach to resolve issue #211. Please provide feedback during code review on whether this solution makes sense.


sooperset (approved):
Thanks for your contribution! This fix for the Windsurf parsing error looks good.

PR #229, opened 2025-04-09
feat: add tool jira_search_fields

cutekibry:
Add tool `jira_search_fields` that can search for field's detail (id, key, name, etc.), using [thefuzz](https://github.com/seatgeek/thefuzz), a fuzzy-search python library.

**Help wanted**: need to add some tests.


sooperset (commented):
Thanks for adding this tool @cutekibry! The `jira_search_fields` functionality is exactly what we needed - finding the right field names/IDs in Jira can be tricky, especially with all those custom fields that vary between instances.

I really like how you've implemented the fuzzy search using `thefuzz`. Checking across id, key, name, and clauseNames is smart and will make the tool much more helpful. The sorting by similarity score makes sense too.

One small thought - how does it behave if someone passes an empty keyword? Probably not a big deal either way, but might be worth considering.

The only blocker I see for merging is we need some tests to make sure this works as expected in different scenarios. Could you add unit tests that cover:
- Finding fields with exact and partial matches
- Making sure it searches across different field attributes
- Verifying the limit parameter works
- Confirming case insensitivity works as expected

Once we have test coverage, this is ready to go. This will be super helpful for users working with different Jira instances - nice contribution!

cutekibry:
> Thanks for adding this tool @cutekibry! The `jira_search_fields` functionality is exactly what we needed - finding the right field names/IDs in Jira can be tricky, especially with all those custom fields that vary between instances.
> 
> I really like how you've implemented the fuzzy search using `thefuzz`. Checking across id, key, name, and clauseNames is smart and will make the tool much more helpful. The sorting by similarity score makes sense too.
> 
> One small thought - how does it behave if someone passes an empty keyword? Probably not a big deal either way, but might be worth considering.
> 
> The only blocker I see for merging is we need some tests to make sure this works as expected in different scenarios. Could you add unit tests that cover:
> 
> * Finding fields with exact and partial matches
> * Making sure it searches across different field attributes
> * Verifying the limit parameter works
> * Confirming case insensitivity works as expected
> 
> Once we have test coverage, this is ready to go. This will be super helpful for users working with different Jira instances - nice contribution!

I would love to add test coverage, but when I run `pytest` it tells me to add Jira and Confluence 
environment variables and no example `.env` file is shown in [CONTRIBUTING.md](https://github.com/sooperset/mcp-atlassian/blob/main/CONTRIBUTING.md).

---

Btw, I guess async test plugin for pytest is not listed in `uv.lock` since it throws warning for that:

```
E:\work\main\mcp-atlassian-enhanced\.venv\Lib\site-packages\_pytest\python.py:148: PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - anyio
    - pytest-asyncio
    - pytest-tornasync
    - pytest-trio
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

src/mcp_atlassian/server.py
```diff
@@ -669,6 +670,33 @@ async def list_tools() -> list[Tool]:
...
                 ),
+                Tool(
+                    name="jira_search_fields",
+                    description="Search Jira fields by keyword with fuzzy match",
+                    inputSchema={
+                        "type": "object",
+                        "properties": {
+                            "keyword": {
+                                "type": "string",
+                                "description": "The keyword to search for",
```
sooperset, line 681: How about updating the current description when the default value is given?

```suggestion
                                "description": "Keyword for fuzzy search. If left empty, lists the first 'limit' available fields in their default order.",
```

sooperset (commented):
Hey @cutekibry,

No worries about real credentials or `.env` for these unit tests – just mock the `get_fields` response with sample data. Also, you can likely ignore the async warning and write standard synchronous tests for the search logic.

Hope that helps!

cutekibry:
Update:
* Add tests for `search_fields`
* Move `search_fields` from `server.py` to `FieldMixins`
* Remove the maximum limit of `limit` in `search_fields`

cutekibry:
It is got too messy, I will clean and open a new one PR.

Sorry for the inconvenience.

PR #245, opened 2025-04-10
fix: add update_sprint

qorpus:
### Description
Added update sprint functionality

### Changes
- added update_sprint method to sprints.py
- added test for update_sprint method to test_sprints.py

### DIFF
--- a/src/mcp_atlassian/jira/sprints.py
+++ b/src/mcp_atlassian/jira/sprints.py
@@ -15,7 +15,7 @@ class SprintsMixin(JiraClient):
     """Mixin for Jira sprints operations."""
 
     def get_all_sprints_from_board(
-        self, board_id: str, state: str = None, start: int = 0, limit: int = 50
+        self, board_id: str, state: str | None, start: int = 0, limit: int = 50
     ) -> list[dict[str, Any]]:
         """
         Get all sprints from a board.
@@ -47,7 +47,7 @@ class SprintsMixin(JiraClient):
             return []
 
     def get_all_sprints_from_board_model(
-        self, board_id: str, state: str = None, start: int = 0, limit: int = 50
+        self, board_id: str, state: str | None, start: int = 0, limit: int = 50
     ) -> list[JiraSprint]:
         """
         Get all sprints as JiraSprint from a board.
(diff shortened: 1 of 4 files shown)


src/mcp_atlassian/jira/sprints.py
```diff
@@ -68,3 +68,55 @@ def get_all_sprints_from_board_model(
...
+        Args:
+            sprint_id: Sprint ID
+            name: New name for the sprint (optional)
+            state: New state for the sprint (future|active|closed - optional)
+            start_date: New start date for the sprint (optional)
+            end_date: New end date for the sprint (optional)
+
+        Returns:
+            Updated sprint
+        """
```
sooperset, line 93: The docstring has a couple of minor inconsistencies:
1. It lists `name` in the Args section, but the parameter is `sprint_name`.
2. It's missing the description for the `goal` parameter.

src/mcp_atlassian/server.py
```diff
@@ -912,6 +912,36 @@ async def list_tools() -> list[Tool]:
...
+                            },
+                            "start_date": {
+                                "type": "string",
+                                "description": "Optional: New start date for the sprint",
+                            },
+                            "end_date": {
+                                "type": "string",
+                                "description": "Optional: New end date for the sprint",
+                            },
+                        },
```
sooperset, line 941: The tool's input schema is missing the `goal` parameter, although it's handled in the `sprints.py` method and extracted in the `call_tool` handler below. Please add it to the schema properties.

src/mcp_atlassian/server.py
```diff
@@ -1825,6 +1855,35 @@ def format_comment(comment: Any) -> dict[str, Any]:
...
+
+            sprint = ctx.jira.update_sprint(
+                sprint_id=sprint_id,
+                sprint_name=sprint_name,
+                goal=goal,
+                start_date=start_date,
+                end_date=end_date,
+                state=state,
+            )
+
```
sooperset, line 1877: The `ctx.jira.update_sprint()` call can return `None` if validation fails or if the underlying API call results in an error (HTTPError or other Exception, which are caught and return `None` in `update_sprint`, or if the API response handler returns `None`). Accessing `sprint.to_simplified_dict()` directly will cause an `AttributeError` in these cases.

You need to check if `sprint` is `None` before attempting to access its methods.

```diff
+            if sprint is None:
+                # Handle the error case, e.g., return an error message
+                error_payload = {
+                    "error": f"Failed to update sprint {sprint_id}. Check logs for details."
+                }
+                return [
+                    TextContent(
+                        type="text",
+                        text=json.dumps(error_payload, indent=2, ensure_ascii=False)
+                    )
+                ]
+
+            # If sprint is not None, proceed as before
```

sooperset (changes requested):
Thanks for adding this `jira_update_sprint` functionality. It's a logical extension to the existing sprint tools and the inclusion of tests and documentation is appreciated.

However, there are a few key points to address before merging:

1. As identified previously, the `call_tool` handler in `server.py` needs to check if `ctx.jira.update_sprint()` returns `None`. The provided context confirms that the underlying `atlassian-python-api`'s `_response_handler` *can* return `None` on errors (like invalid JSON responses or other exceptions caught within it), and the PR's `update_sprint` *also* explicitly catches `HTTPError` and `Exception` and returns `None`. Failing to check for `None` in `server.py` will lead to an `AttributeError` when trying to access `sprint.to_simplified_dict()`.
2.  The `goal` parameter is implemented in the `SprintsMixin` and used in the `call_tool` handler but is missing from the `inputSchema` definition for the `jira_update_sprint` tool in `server.py`. This needs to be added for the tool to be correctly advertised and validated.
3. Minor updates are needed in the `update_sprint` docstring for parameter name consistency (`sprint_name`) and to include the `goal` parameter description.

Additionally, a minor suggestion is to consider adding basic ISO 8601 format validation for the `start_date` and `end_date` parameters within the `update_sprint` method for robustness, although the API itself will likely reject invalid formats.

Addressing the server error handling and the missing schema parameter is crucial for merging.

src/mcp_atlassian/jira/sprints.py
qorpus: Thanks!

src/mcp_atlassian/server.py
qorpus: Great, my bad here.

src/mcp_atlassian/server.py
qorpus: Totally makes sense thank you!

PR #270, opened 2025-04-16
feat(search): use siteSearch as default with text search fallback

LorisFriedel:
## Description
Enhances Confluence search functionality by using `siteSearch` instead of `text` search for simple queries. This change leverages the same search mechanisms as the Confluence WebUI, providing more relevant results by default. An automatic fallback to text search ensures compatibility with older Confluence Server/Data Center versions.

Without this, simple searches on Confluence rarely return the most relevant results, making it harder for an Agent to find what's needed.

## Changes
- Updated the search conversion to use `siteSearch ~ "query"` instead of `text ~ "query"` for simple terms.
- Added an automatic fallback mechanism to text search if siteSearch fails.
- Added unit tests for the updated search.
- Updated tool specs to reflect the new search behavior and fallback mechanism.

## Testing
[x] Unit tests added/updated
[x] Integration tests passed
[x] Manual checks performed: Verified simple search terms are properly converted to siteSearch queries and provide much higher quality results.

## Checklist
[x] Code follows project style guidelines (linting passes on the changes).
[x] Tests added/updated for changes.
[x] All tests pass locally.
[x] Documentation updated.

### DIFF
--- a/src/mcp_atlassian/server.py
+++ b/src/mcp_atlassian/server.py
@@ -340,10 +340,12 @@ async def list_tools() -> list[Tool]:
                         "properties": {
                             "query": {
                                 "type": "string",
-                                "description": "Search query - can be either a simple text (e.g. 'project documentation') or a CQL query string. Examples of CQL:\n"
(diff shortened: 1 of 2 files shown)


sooperset (approved):
LGTM!

This looks like a solid improvement. Switching simple searches to `siteSearch` by default should provide much more relevant results, mirroring the web UI experience. The fallback to `text` search is a good safety net for compatibility. Tests cover the new logic well. Nice work!

PR #328, opened 2025-04-23
feat: add support for getting and adding labels for Confluence pages

mroeckl:
## Description

This PR adds support for labels on Confluence pages:
- get all labels on a page
- add a label to a page

Note: I am not very experienced in Python programming. Please excuse any mistakes 🙈

## Changes

- Add `ConfluenceLabel` with the properties as defined in [Atlassian REST API Spec](https://developer.atlassian.com/server/confluence/rest/v900/api-group-content-labels/#api-rest-api-content-id-label-get-response):
  - id (string): unique id of this label
  - name (string): the actual name of the label
  - prefix (string): prefix (default: global)
  - label (string): I could not find a clear definition. It is mentioned in the spec. My Confluence server always returns an empty string.
- Add 2 new Tools `confluence_get_labels` and `confluence_add_label`
- Add unit and real API tests

## Testing

- [✔️] Unit tests added/updated
- [✔️] Integration tests passed
- [✔️] Manual checks performed: Added it to the vscode extension Cline as MCP server and successfully used prompts like _'Which labels are set on the page with "Test" in the title?'_ and _'Add the label "test" to all subpages of the page 12345'_

## Checklist

- [✔️] Code follows project style guidelines (linting passes).
- [✔️] Tests added/updated for changes.
- [✔️] All tests pass locally.
- [✔️ ] Documentation updated (if needed).

### DIFF
--- a/src/mcp_atlassian/confluence/labels.py
+++ b/src/mcp_atlassian/confluence/labels.py
@@ -1,0 +1,87 @@
+"""Module for Confluence label operations."""
+
+import logging
+
+import requests
+
+from ..models.confluence import ConfluenceLabel
+from .client import ConfluenceClient
+
+logger = logging.getLogger("mcp-atlassian")
+
+
+class LabelsMixin(ConfluenceClient):
+    """Mixin for Confluence label operations."""
+
+    def get_page_labels(self, page_id: str) -> list[ConfluenceLabel]:
+        """
+        Get all labels for a specific page.
+
+        Args:
+            page_id: The ID of the page to get labels from
+
+        Returns:
+            List of ConfluenceLabel models containing label content and metadata
+        """
+        try:
+            # Get page info to extract space details
+            page = self.confluence.get_page_by_id(page_id=page_id, expand="space")
+            space_key = page.get("space", {}).get("key", "")
+
+            # Get labels with expanded content
+            labels_response = self.confluence.get_page_labels(page_id=page_id)
+
+            # Process each label
+            label_models = []
+            for label_data in labels_response.get("results", []):
+                # Create the model with the processed content
+                label_model = ConfluenceLabel.from_api_response(
+                    label_data,
+                    base_url=self.config.url,
+                )
+
+                label_models.append(label_model)
+
+            return label_models
+
+        except requests.RequestException as e:
+            logger.error(f"Network error when fetching labels: {str(e)}")
+            return []
+        except (ValueError, TypeError) as e:
+            logger.error(f"Error processing label data: {str(e)}")
+            return []
+        except Exception as e:  # noqa: BLE001 - Intentional fallback with full logging
+            logger.error(f"Unexpected error fetching labels: {str(e)}")
+            logger.debug("Full exception details for labels:", exc_info=True)
+            return []
+
+    def add_page_label(self, page_id: str, name: str) -> list[ConfluenceLabel]:
+        """
+        Add a label to a Confluence page.
+
+        Args:
+            page_id: The ID of the page to update
+            name: The name of the label
+
+        Returns:
+            Label model containing the updated list of labels
+
+        Raises:
+            Exception: If there is an error adding the label
+        """
+        try:
+            logger.debug(f"Adding label with name '{name}' to page {page_id}")
+
+            update_kwargs = {
+                "page_id": page_id,
+                "label": name,
+            }
+            response = self.confluence.set_page_label(**update_kwargs)
+
+            # After update, refresh the page data
+            return self.get_page_labels(page_id)
+        except Exception as e:
+            logger.error(f"Error adding label '{name}' to page {page_id}: {str(e)}")
+            raise Exception(
+                f"Failed to add label '{name}' to page {page_id}: {str(e)}"
+            ) from e
--- a/src/mcp_atlassian/confluence/labels.py
+++ b/src/mcp_atlassian/confluence/labels.py
@@ -2,8 +2,6 @@
 
 import logging
 
-import requests
-
 from ..models.confluence import ConfluenceLabel
 from .client import ConfluenceClient
 
@@ -22,6 +20,9 @@ class LabelsMixin(ConfluenceClient):
 
         Returns:
             List of ConfluenceLabel models containing label content and metadata
+
+        Raises:
+            Exception: If there is an error getting the label
         """
         try:
             # Get labels with expanded content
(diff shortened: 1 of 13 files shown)


src/mcp_atlassian/server.py
```diff
@@ -1169,6 +1203,15 @@ def format_comment(comment: Any) -> dict[str, Any]:
...
 
+        def format_label(label: Any) -> dict[str, Any]:
+            if hasattr(label, "to_simplified_dict"):
+                # Cast the return value to dict[str, Any] to satisfy the type checker
+                return cast(dict[str, Any], label.to_simplified_dict())
+            return {
+                "name": label.get("name", {}),
+                "prefix": comment.get("prefix"),
+            }
+
```
sooperset, line 1214: The `format_label` helper function has a fallback logic (`return {"name": label.get("name", {}), "prefix": comment.get("prefix")}`) which seems incorrect and unnecessary.

1.  The fallback references `comment.get("prefix")`, but the variable is `label`.
2.  The `get_page_labels` tool call always receives `ConfluenceLabel` model instances from the mixin, which will always have the `to_simplified_dict` method. Therefore, the `if hasattr(label, "to_simplified_dict")` check will always be true, and the fallback path is unreachable.

Consider simplifying this helper to directly use `label.to_simplified_dict()`.

```suggestion
            return cast(dict[str, Any], label.to_simplified_dict())
```

tests/unit/confluence/test_labels.py
```diff
@@ -0,0 +1,151 @@
...
+                    "id": "456789124",
+                    "prefix": "my",
+                    "name": "important",
+                },
+                {
+                    "id": "456789125",
+                    "name": "test",
+                },
+            ]
+        }
```
sooperset, line 53: The return value is already set by the labels_mixin fixture using `MOCK_LABELS_RESPONSE`. No need to set it again here.


```suggestion
```

mroeckl:
Thank you very much, @sooperset , for the great feedback and improvements 👍  I incorporated all suggestions and ran the unit and real API tests against my Confluence as well as the manual tests with an MCP client. Everything looks good to me.

src/mcp_atlassian/confluence/labels.py
```diff
@@ -0,0 +1,83 @@
...
+        except requests.RequestException as e:
+            logger.error(f"Network error when fetching labels: {str(e)}")
+            return []
+        except (ValueError, TypeError) as e:
+            logger.error(f"Error processing label data: {str(e)}")
+            return []
+        except Exception as e:  # noqa: BLE001 - Intentional fallback with full logging
+            logger.error(f"Unexpected error fetching labels: {str(e)}")
+            logger.debug("Full exception details for labels:", exc_info=True)
+            return []
```
sooperset, line 52: The error handling here returns an empty list (`[]`) if any exception occurs while fetching labels. This is slightly inconsistent with `add_page_label` which raises exceptions.

While returning `[]` might be acceptable for `get`, raising the exception might provide more context to the caller (`call_tool`) about *why* labels couldn't be fetched (e.g., network error vs. permission error vs. unexpected data format).

Consider standardizing the error handling approach. If you prefer returning `[]` here, maybe add a more specific log message indicating why an empty list is being returned (e.g., "Returning empty label list due to API error: ..."). However, letting exceptions propagate to `call_tool` might be cleaner.

src/mcp_atlassian/models/confluence/label.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class ConfluenceLabel(ApiModel):
+    """
+    Model representing a Confluence label.
+    """
+
+    id: str = CONFLUENCE_DEFAULT_ID
+    name: str | None = None
+    prefix: str = EMPTY_STRING
```
sooperset, line 25: The current default for `prefix` is `EMPTY_STRING`. However, analysing the [Confluence REST API documentation](https://developer.atlassian.com/cloud/confluence/rest/v2/api-group-label/#api-attachments-id-labels-get) (both v1 and v2) indicates that the expected values for the `prefix` field in API responses are typically `global`, `my`, `team`, or `system`. The documentation doesn't suggest that an empty string is a valid value returned by the API for an existing label's prefix. Since `global` is the most common prefix, especially for labels created via the UI or simple API calls, using `"global"` as the default here would better reflect the expected API data structure when a label object is instantiated or if the field were ever missing in a response.

sooperset (changes requested):
Hi @mroeckl, thanks again for this contribution adding Confluence label support! I left a few comments!

src/mcp_atlassian/confluence/labels.py
mroeckl: Thank you, @sooperset , for your comment. I updated `get_page_labels` to throw exceptions when there is an invalid response, no result or any other error like a communications error - exactly the same approach as in `add_page_labels`. See [2ec22a](https://github.com/sooperset/mcp-atlassian/commit/2ec22a8177bfcbeefb494bdac0fee2047ad7102e).
Does it correspond roughly to what you had in mind?

src/mcp_atlassian/models/confluence/label.py
mroeckl: I've now set `global` as default `prefix` in [2c0542c](https://github.com/sooperset/mcp-atlassian/pull/328/commits/2c0542c509621244f5ec35d727ea587828f74547). 

Do you think I should be ever more strict with an enum type like:
```python
class Prefix(Enum):
    GLOBAL = "global"
    SYSTEM = "system"
    TEAM = "team"
    MY = "my"
```

src/mcp_atlassian/models/confluence/label.py
```diff
@@ -0,0 +1,62 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ConfluenceLabel(ApiModel):
+    """
+    Model representing a Confluence label.
+    """
+
+    id: str = CONFLUENCE_DEFAULT_ID
+    name: str | None = None
```
mroeckl, line 24: @sooperset , do you think I should change this line to  `name: str = EMPTY_STRING`?

sooperset: Yes, you're right. Since this is a minor change and some time has passed, I'll go ahead and fix it directly.

sooperset (approved):
@mroeckl Sorry for the delayed review. I've gone ahead and fixed this minor issue directly and pushed the change. Thank you for your contribution!

PR #339, opened 2025-04-25
Correct jira_update_issue example for assignee format

eran-rom:
## Description

This PR corrects the example JSON provided in the `jira_update_issue` tool description within `src/mcp_atlassian/server.py`. The previous example incorrectly showed the `assignee` value as a dictionary (`{"name": "..."}`). This was inconsistent with the implementation in `IssuesMixin.update_issue`, which expects a string identifier (username, email, or account ID) for the assignee.

This incorrect example caused runtime error, `AttributeError: 'dict' object has no attribute 'startswith'`, when the tool was used with the assignee provided as a dictionary. This fix resolves the inconsistency and prevents this error.

## Changes

- Corrected the `assignee` example format in the `jira_update_issue` tool description (`server.py`).

## Testing

- [x] Manual checks performed: Confirmed the fix works when using the `jira_update_issue` tool via Cursor MCP.

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/server.py
+++ b/src/mcp_atlassian/server.py
@@ -944,7 +944,7 @@ async def list_tools() -> list[Tool]:
                                     "description": (
                                         "A valid JSON object of fields to update as a string. "
                                         'Example: \'{"summary": "New title", "description": "Updated description", '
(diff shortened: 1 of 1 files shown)


sooperset (approved):
Thanks for catching this inconsistency and providing the fix!

PR #345, opened 2025-04-28
Refactor: improve code quality and clean up technical debt

cutekibry:
## Description
This PR focuses on comprehensive code quality improvements, with the primary goal of eliminating all linter errors in `/jira/*.py`. The changes include date handling unification, type system enhancements, and code duplication removal.

### 1. Date Handling Unification
- Consolidated all date-related functions into `utils/date.py`:
  - Adopted `dateutil.parser.parse` for more flexible date parsing
  - Changed return type to `datetime | None` for better type safety
  - Eliminated redundant `<Mixin>._parse_date` methods as they were unnecessary for LLM processing
  - Removed duplicate date formatting implementations across different mixins

### 2. Type System Improvements
- Enhanced `protocol.py`:
  - Introduced new Protocol definitions for better interface segregation
  - Added `@abstractmethod` decorators to enforce implementation requirements
  - Removed problematic `@runtime_checkable` decorators (see [Python docs](https://docs.python.org/3/library/typing.html#typing.runtime_checkable))
- Strengthened type system throughout the codebase:
  - Added comprehensive type hints
  - Implemented stricter type checking
  - Initialized `.pylintrc` to express the using of `f-str` in loggers explicitly to linter
- Removed unsafe `hasattr` checks:
  - These checks were unreliable as they wouldn't automatically update with type/name changes

### 3. Code Duplication Elimination
- Consolidated `get_jira_field_ids` implementations:
  - Removed duplicate implementations from `EpicsMixin`, `IssuesMixin`, and `FieldsMixin`
  - Centralized implementation in `FieldsMixin` using `EpicsMixin`'s logic (`JiraFetcher` inherits `EpicsMixin` first)
  - Renamed to `get_field_ids_to_epic` for clarity
- Consolidated `get_epic_issues` implementations:
  - Removed duplicate implementation in `SearchMixin`
  - Keep implementation in `EpicsMixin` unchanged
- Refactored `JiraClient._field_id`:
  - Renamed to `_field_ids_cache` for better naming consistency
  - Modified type from `dict[str, str] | None` to `list[dict[str, Any]] | None` to match API structure

### 4. Test Suite Improvements
- Updated test cases to reflect new implementations
- Eliminated redundant test scenarios
- Removed dangerous `delattr` usage in tests:
  - This practice was particularly risky as deleting attributes are not expected when linter performs type checks
  - Even with linter hints showing attribute existence, deleted attributes would cause runtime errors
  - This change prevents potential fatal issues in attribute access patterns

[...]

PR #346, opened 2025-04-28
Feat: add batch get changelogs

cutekibry:
## Description
**This PR is based on #345**.

Add `jira_batch_get_changelogs` that can fetch all changelogs of given issues repeatly using `nextPageToken`.

## Changes

- Added `JiraClient.get_paged` method for paginated data Jira API fetching using `nextPageToken`
  - It should **NOT** be exposed to users
  - Some of Jira's APIs cannot fetch all datas in one query. For user's convenience, it would be great if we provide user with a option, since MCP tools are not supposed to expose API calling function to the user
- Added `IssuesMixin.batch_get_changelogs` tool for batch fetching changelogs of multiple issues
- Added `int` support to `utils.parse_date`
 
## Testing
- [x] Manual checks performed: can return expected results

## Checklist
- [ ] All tests pass locally.


sooperset (approved):
Looks great, thanks for adding this @cutekibry!

The new `jira_batch_get_changelogs` tool is a solid addition, and the `get_paged` helper method provides a clean way to handle Jira Cloud pagination. I appreciate the careful handling of Cloud vs Server/DC distinctions and the included tests.

Everything looks good to merge. 👍

PR #369, opened 2025-05-03
Use enhanced_jql_get_list_of_tickets to allow retrieving more than 50 issues

swapnilsurdi:
## Description
There was no way to retrieve more than 50 results from jira search. The changes added allows user to fetch more results using the `_enhanced_jql_get_list_of_tickets_` function instead of `enhanced_jql` one.
enhanced_jql uses nextPageToken to paginate the results but the token was not passed in the response.

Fixes: #

## Changes
- Changed the Jira Cloud search implementation to use enhanced_jql_get_list_of_tickets instead of enhanced_jql.
- Converting the list response form the function to a structure compatible with JiraSearchResult
- Setting correct total count values based on actual results
- Preserving the requested fields in the search result model

## Testing

- [ ] Unit tests updated
- [ ] Manual checks performed: `Searched various issues types using JQL and various limits`

## Checklist

https://github.com/user-attachments/assets/9e70c556-3b05-4be7-ba62-186f1fbe842e

- [ ] All tests pass locally (Except for the create issue in test_server)

### DIFF
--- a/src/mcp_atlassian/jira/search.py
+++ b/src/mcp_atlassian/jira/search.py
@@ -83,26 +83,42 @@ class SearchMixin(JiraClient, IssueOperationsProto):
                 fields_param = fields
 
             if self.config.is_cloud:
-                response = self.jira.enhanced_jql(
+                response = self.jira.enhanced_jql_get_list_of_tickets(
                     jql, fields=fields_param, limit=limit, expand=expand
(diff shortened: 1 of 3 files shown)


sooperset (approved):
Great feature! Thank you for the contribution @swapnilsurdi

PR #402, opened 2025-05-09
fix: Enhance Atlassian OAuth Flow, Callback Handling, and Logging

sooperset:
## Description

This PR significantly improves the reliability, debuggability, and user experience of the Atlassian OAuth 2.0 authorization flow. It resolves critical issues such as the "501 Unsupported method ('GET')" error during callback, enhances logging for easier troubleshooting, and makes the overall process more robust.

Fixes: # (If you have an issue number, add it here)

## Changes

- Corrected the HTTP method name in `CallbackHandler` (from `do_get` to `do_GET`) to properly handle Atlassian's OAuth callback, resolving 501 errors.
- Enhanced logging configuration in `scripts/oauth_authorize.py` to default to DEBUG level with line numbers, and ensured specific loggers (`oauth-authorize`, `mcp-atlassian.oauth`) also output detailed logs.
- Improved `CallbackHandler` to specifically filter for `/callback` requests and correctly handle `/favicon.ico` (and other) requests with a 404, reducing log noise.
- Refined logging in `OAuthConfig.exchange_code_for_tokens` for clearer insights into the token exchange process, including request/response details and more specific error handling (network, JSON decode).
- Clarified CSRF state variable naming in `scripts/oauth_authorize.py` for better code readability.

## Testing

- [ ] Unit tests added/updated
- [ ] Integration tests passed
- [x] Manual checks performed:
    - Successfully completed the end-to-end OAuth 2.0 authorization flow.
    - Verified that access and refresh tokens, along with the cloud ID, were correctly obtained and stored.
    - Confirmed that detailed DEBUG logs were generated throughout the process in both `oauth_authorize.py` and `oauth.py`.
    - Ensured that the browser no longer shows a 501 error and instead displays the "Authorization successful" message.
    - Verified that requests to `/favicon.ico` during the callback process are handled with a 404 and do not interfere with the main OAuth flow or produce misleading error logs.

## Checklist

- [x] Code follows project style guidelines (linting passes).
- [ ] Tests added/updated for changes. (Assuming manual tests cover this for now, can be updated if unit tests are added later)
- [x] All tests pass locally. (Based on manual checks)
- [ ] Documentation updated (if needed). (The code comments and logging are self-documenting for these changes)

PR #404, opened 2025-05-09
docs: Temporarily hide OAuth documentation until fully verified

sooperset:
This PR temporarily removes OAuth 2.0 authentication documentation from the README.

Recent testing revealed some issues with the OAuth flow in Docker environments. To prevent user frustration with an unverified setup process, we're hiding these sections until we can fully validate the implementation.

Once testing is complete and any issues are resolved, we'll restore the documentation with any necessary updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,41 +54,6 @@ MCP Atlassian supports three authentication methods:
 2. Click **Create token**, name it, set expiry
 3. Copy the token immediately
 
-#### C. OAuth 2.0 Authentication (Cloud only)
-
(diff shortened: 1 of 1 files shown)


sooperset:
Hi @zparnold,

Thanks again for your work implementing OAuth 2.0 support! 

I'm temporarily hiding the OAuth setup instructions in the README while I address the integration issues we discussed in #402. I want to ensure users have a smooth experience when following our documentation.

I'll restore these sections once I've verified everything works properly across all environments. Your insights would be valuable when I revisit this.

Thanks for understanding!

Issue #478, opened 2025-06-03
[Feature]: Apply Project Filter to the get_jira_issue tool


schaveyt:
### Is your feature request related to a problem? Please describe.

@lonelymoon87 did an amazing job add the project filter feature.

When the project filter is specified, it only restricts the results in the jira_search. This means that while folks are prevented from obtaining information of restricted project when using the jira_search, if someone has the Jira ID, then are still able to obtain the restricted information

### Describe the solution you'd like

Apply project filter logic to the `get_jira_issue`

schaveyt:
I can bake a pull request for this in the next day or so 👍

schaveyt:
@sooperset PR submitted for your review 👍

PR #485, opened 2025-06-05
Feature: update jira_get_issue to honor the project filter

schaveyt:
## Description

This PR adds logic to the `jira_get_issue` to have it honor the project filter configuration such that when a request for a Jira Issue is rejected if it's prefix does NOT match any project filters (if defined). 

The user's request should NOT be able to override this.

Fixes: #478 

## Changes

- Modified the CONTRIBUTING.md to align use standard markdown numbers and key steps I needed to get it the env operational
- Modify jira/get_issue logic to honor the project filter config
- Added a unit test

## Testing

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -76,8 +76,8 @@ class IssuesMixin(
                 if issue_key_project not in projects:
                     # If the project key not in the filter, return an empty issue
                     msg = (
-                        "Issue with with project prefix "
-                        f"'{issue_key_project}' are restricted"
+                        "Issue with project prefix "
+                        f"'{issue_key_project}' are restricted by configuration"
                     )
                     raise ValueError(msg)
 
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -61,6 +61,26 @@ class IssuesMixin(
             Exception: If there is an error retrieving the issue
         """
         try:
+            # Obtain the projects filter from the config.
+            # These should NOT be overridden by the request.
+            filter_to_use = self.config.projects_filter
+
+            # Apply projects filter if present
+            if filter_to_use:
+                # Split projects filter by commas and handle possible whitespace
+                projects = [p.strip() for p in filter_to_use.split(",")]
+
+                # Obtain the project key from issue_key
+                issue_key_project = issue_key.split("-")[0]
+
+                if issue_key_project not in projects:
+                    # If the project key not in the filter, return an empty issue
+                    msg = (
+                        "Issue with with project prefix "
+                        f"'{issue_key_project}' are restricted"
+                    )
+                    raise ValueError(msg)
+
             # Determine fields_param: use provided fields or default from constant
             fields_param = fields
             if fields_param is None:
--- a/tests/unit/jira/test_issues.py
+++ b/tests/unit/jira/test_issues.py
@@ -1522,3 +1522,39 @@ class TestIssuesMixin:
         # Verify result
         assert result.key == "TEST-123"
         assert result.labels == ["bug", "frontend"]
+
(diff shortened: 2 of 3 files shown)


src/mcp_atlassian/jira/issues.py
```diff
@@ -61,6 +61,26 @@ def get_issue(
...
+                projects = [p.strip() for p in filter_to_use.split(",")]
+
+                # Obtain the project key from issue_key
+                issue_key_project = issue_key.split("-")[0]
+
+                if issue_key_project not in projects:
+                    # If the project key not in the filter, return an empty issue
+                    msg = (
+                        "Issue with with project prefix "
+                        f"'{issue_key_project}' are restricted"
```
sooperset: There is a small typo here. and I'd like to suggest rephrasing slightly for clarity

```suggestion
                        "Issue with project prefix "
                        f"'{issue_key_project}' are restricted by configuration"
```

tests/unit/jira/test_issues.py
```diff
@@ -1522,3 +1522,39 @@ def test_create_issue_with_labels_as_string(self, issues_mixin: IssuesMixin):
...
+            "total": 1,
+            "startAt": 0,
+            "maxResults": 50,
+        }
+        issues_mixin.jira.jql.return_value = mock_issues
+        issues_mixin.config.url = "https://example.atlassian.net"
+        issues_mixin.config.projects_filter = "DEV"
+
+        # Mock the API to raise an exception
+        issues_mixin.jira.get_issue.side_effect = Exception("API error")
```
sooperset, line 1552: This `side_effect` is confusing because if the filter logic works correctly, the `jira.get_issue` method should never be called. The test should focus on verifying that the `ValueError` is raised by the filter logic *before* any API call is attempted.

I recommend removing this line and adding `issues_mixin.jira.get_issue.assert_not_called()` after the `pytest.raises` block to make the test's intent clearer and more precise.

tests/unit/jira/test_issues.py
```diff
@@ -1522,3 +1522,39 @@ def test_create_issue_with_labels_as_string(self, issues_mixin: IssuesMixin):
...
+
+        # Call the method and verify it raises the expected exception
+        with pytest.raises(
+            Exception,
+            match=(
+                "Error retrieving issue TEST-123: "
+                "Issue with with project prefix 'TEST' are restricted"
+            ),
+        ):
+            issues_mixin.get_issue("TEST-123")
```
sooperset, line 1562: This test looks good overall, but I think we could strengthen the test coverage a bit. Right now we're only testing the rejection scenario, but it would be great to also verify that the feature works correctly for allowed cases.

Could you add a few more tests to cover:
- A case where the project actually matches the filter (to make sure we're not accidentally blocking legitimate requests)
- Multiple projects in the filter like `"DEV,PROD"` to ensure both projects work
- Filters with extra whitespace like `" DEV , PROD "` since we have that `strip()` logic

sooperset (commented):
@schaveyt Thanks for the excellent work on this PR! It covers all the required changes nicely. I've added a few minor comments to help align the new code with our existing style. Looking forward to getting this merged.

schaveyt:
Noice. Yup, all valid comments. I'll look to address all comments 👍

tests/unit/jira/test_issues.py
schaveyt: Fixed

tests/unit/jira/test_issues.py
schaveyt: Added 👍

src/mcp_atlassian/jira/issues.py
schaveyt: Fixed 👍

schaveyt:
@sooperset All comments now addressed I believe

sooperset (approved):
LGTM! Great work, @schaveyt  🎉

PR #487, opened 2025-06-05
feat: add custom User-Agent header support for firewall compatibility

tomashucek:
## Description
This feature helps enterprise environments bypass firewall restrictions that filter requests based on User-Agent headers.

Fixes: #

## Changes

- Add JIRA_USER_AGENT and CONFLUENCE_USER_AGENT environment variables
- Add concise documentation in README for enterprise environments
- Update .env.example with User-Agent configuration options
- Add Docker port exposure for SSE transport mode

## Testing

- [x] Manual checks performed: `[briefly describe]`

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,10 +9,10 @@
 # REQUIRED: Replace with your Atlassian instance URLs.
 # Example Cloud: https://your-company.atlassian.net
 # Example Server/DC: https://jira.your-internal-domain.com
-JIRA_URL=https://your-company.atlassian.net
+#JIRA_URL=https://your-company.atlassian.net
 # Example Cloud: https://your-company.atlassian.net/wiki
 # Example Server/DC: https://confluence.your-internal-domain.com
-CONFLUENCE_URL=https://your-company.atlassian.net/wiki
+#CONFLUENCE_URL=https://your-company.atlassian.net/wiki
 
 # =============================================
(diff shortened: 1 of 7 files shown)


sooperset:
Hi, this feature seems for bypassing corporate security policies. I think IT should allowlist the tool's legitimate User-Agent string. closing as this is out of scope for this project

tomashucek:
Hi @sooperset ,
I think there's a misunderstanding about the technical problem this solves.

The issue: Your current code sends a generic User-Agent like python-requests/2.31.0. Our firewall blocks this and only allows specific strings that IT has configured.

What you suggested: "IT should allowlist the legitimate User-Agent string"
The problem: There is no "legitimate" User-Agent to allowlist - your code sends random library defaults that change with updates.

This PR solves it: IT can configure the firewall to allow "MyCompany-Integration" (or whatever), and the MCP server can send exactly that string.

Without this feature, there's literally no predictable User-Agent string for IT to allowlist. Your suggestion isn't technically possible with the current implementation.

Happy to fork if needed, but this PR enables basic enterprise firewall compatibility.

khalidyousif-smtc:
@tomashucek indeed, **python-requests/2.31.0** is just too generic if we wanted to audit who is using mcp-atlassian, when, and what for.

PR #493, opened 2025-06-09
feature: Tool to list available jira projects

MacanPN:
## Summary
Adds get_all_projects MCP tool to retrieve all Jira projects accessible to the current user.

## Motivation
Many existing MCP tools require project keys as input (e.g., create_issue, get_project_issues), but there was no way for users to discover what projects they have access to. This tool fills that gap by enabling project discovery before using project-specific tools.

## Implementation
- Uses existing ProjectsMixin.get_all_projects() method
- Includes optional include_archived parameter (defaults to false)
- Returns JSON array of project objects with keys, names, descriptions, etc.
- Follows existing codebase patterns and conventions

## Testing

- [x] Manual checks performed: `[briefly describe]`
 The tool correctly lists projects in my Jira account

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,6 +550,7 @@ Here's a complete example of setting up multi-user authentication with streamabl
 - `jira_update_issue`: Update an existing issue
 - `jira_transition_issue`: Transition an issue to a new status
 - `jira_add_comment`: Add a comment to an issue
+- `jira_get_all_projects`: List all Jira projects accessible to the current user
 - `jira_get_project_versions`: List all fix versions for a project
 - `jira_create_version`: Create a new version (fix version) in a Jira project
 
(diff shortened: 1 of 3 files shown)


schaveyt:
Be sure to update this implemenation to honor the Project Filter list. If the the project filter list is defined, this function should literally only return those list of ptrojects

MacanPN:
@schaveyt I've added support for `jira.config.projects_filter`. IMO this does not help anything since `projects_filter` does not support wildcards (the tool is implemented so the LLM could run project discovery - what projects exist. To use the filter, one needs to already know what projects exist.) In any case, it's implemented now :-)

schaveyt:
Great to hear. There are very real use case where one must blacklist certain projects from --ever-- being discovered. By adding this one can now safely allow the LLM to discover things without admins worrying improper access 👍

sooperset (approved):
LGTM! Thanks for the contribution @MacanPN

PR #494, opened 2025-06-10
feat: implement env variable MCP_LOGGING_STDOUT for streaming logs to STDOUT (instead of STDERR)

dnnspaul:
## Description
The [default `logging.StreamHandler()` writes to `sys.stderr`](https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler). In our cloud environments we experience the problem that the mcp-atlassian service floods our cloud logging with messages that are flagged as errors, even the DEBUG ones. Since we want to keep the logs instead of blocking them from arriving at our cloud logging, we implemented this solution to write the logs to `sys.stdout` instead of `sys.stderr`.

A short insight of our google cloud logging:
image

## Changes
Introducing the environment variable `MCP_LOGGING_STDOUT` that writes the logs to `sys.stdout` instead of `sys.stderr`.

## Testing

- [x] Unit tests added/updated
- [ ] Integration tests passed
- [ ] Manual checks performed: `[briefly describe]`

## Checklist

- [x] Code follows project style guidelines (linting passes).
- [x] Tests added/updated for changes.
- [x] All tests pass locally.
- [x] Documentation updated (if needed).

PR #505, opened 2025-06-13
feat: refactor environment truthy-check

dnnspaul:
Hi sooperset,
like we discussed previously in https://github.com/sooperset/mcp-atlassian/pull/494, here is the refactoring to align with the DRY principle.


sooperset (dismissed):
Great! Thanks for the contribution @dnnspaul

Issue #534, opened 2025-06-19
[Feature]: Be able to instantiate more than one jira in the same server


tprelle:
### Is your feature request related to a problem? Please describe.

Currently, it's not possible to communicate with multiple instance of jira datacenter in same ReActAgent

### Describe the solution you'd like

Adding a function to register tools and be able to instance config using a dictionary

### Use Case

In a big compagny, you can have more than one jira instance sometime one for frontline and one for project but i want to be able to query both it's the same ReAct agent

Issue #231, opened 2025-04-09
Support for multiple Jira/Confluence instances per server process


CopilotPilot:
**Context:**
I'm using the `mcp-atlassian` server to provide tools for interacting with Jira and Confluence from a LangChain agent, using the `langchain-mcp-adapters` library and its `MultiServerMCPClient`.

**Use Case:**
My application needs to connect to multiple, distinct Jira instances and Confluence instances simultaneously.

**Current Approach & Understanding:**
Currently, it seems the `mcp-atlassian` server process is configured via environment variables (e.g., JIRA_URL, JIRA_API_TOKEN, CONFLUENCE_URL, CONFLUENCE_API_TOKEN) which define one specific Jira site and one specific Confluence site per running server process.

To connect to multiple sites (e.g., 5 different Jira instances), the current workaround is to:

Define separate sets of environment variables for each instance (e.g., JIRA_URL_SITE1, JIRA_TOKEN_SITE1, JIRA_URL_SITE2, JIRA_TOKEN_SITE2, ...).
Configure a client (like `MultiServerMCPClient`) to launch and manage multiple, separate `mcp-atlassian` processes, each with the appropriate environment variables for one site.

**Question:**

Is this multi-process approach (one `mcp-atlassian` process per Jira/Confluence site pair) the intended and recommended design for handling connections to multiple distinct Atlassian instances via MCP?
Would you consider adding support for configuring and managing connections to multiple Atlassian sites within a single `mcp-atlassian` server process?

Thank you in advance!

sooperset:
Hi @CopilotPilot,
Yes, multi-site support is planned and related to the multi-user goal in #221.
Currently, MCP's SSE transport limits our ability to do this well in a single process. We need MCP's upcoming streamable HTTP support first.
Once streamable HTTP is fully available in the MCP standard, we intend to add multi-site capability to mcp-atlassian. For now, the multi-process approach is required.

Kirchen99:
Hi, thanks for the great work on this project!

I noticed there's a Pull Request #416 that adds support for multiple users, which is great. However, Atlassian Server/Data Center requires separate Personal Access Tokens (PATs) for Jira and Confluence, while the current PR still assumes using only one PAT.

Just want to confirm if that's the limitation, or if there's any plan to support per-application PATs in the future.

Thanks!

Troubladore:
[...]

PR #535, opened 2025-06-19
feat : Support more than one Jira/Confluence in the same server (#533)

tprelle:
In order to be able to instantiate more than one jira or confluence in the same server, we need a function to register tools and be able to instance config using custom dictionary.

Reported-by: tprelle
Github-Issue: #534

## Description

In order to be able to instantiate more than one jira or confluence in the same server, we need a function to register tools and be able to instance config using custom dictionary.

Fixes: # 533

## Changes

- make config taking a dictionary
- create a register_tools function
- create a build_main_lifespan 

## Testing

## Checklist

- [ ] All tests pass locally.


sooperset:
@tprelle Thanks for the contribution! This is really valuable feature. Would you add examples and tests demonstrating the multi-instance usage before merging?

sooperset:
also relevant issue: #231

tprelle:
> @tprelle Thanks for the contribution! This is really valuable feature. Would you add examples and tests demonstrating the multi-instance usage before merging?

I did the code but i'm struggling with ruff to avoid duplicating code between tests but ruff do not really like that i use import only on pytest.fixture function argument like

`from tests.unit.servers.test_jira_server import mock_jira_fetcher,mock_base_jira_config, mock_request as mock_jira_request
from tests.unit.servers.test_confluence_server import mock_confluence_fetcher,mock_base_confluence_config

@pytest.fixture
def test_multi_mcp(
    mock_base_jira_config,
    mock_base_confluence_config,
):`

tprelle:
@sooperset  i add a test https://github.com/sooperset/mcp-atlassian/pull/535/files#diff-cfd4948ecea88291526df2487954a00e4ff3aab941e2a375b73d45a787a20b7cR363 inspire on how i use it as mounted server inside our big one

sooperset:
@tprelle Thanks for the contribution! I noticed that issue #534 requests support for multiple Jira instances in a single MCP server, but this PR seems to primarily add programmatic configuration capabilities.

For true multi-instance support, wouldn't we need to implement instance-specific tool prefixes (e.g., jira_frontline_create_issue, jira_project_create_issue) instead of the current shared jira_ prefix? This would enable users to interact with multiple Jira instances through a single MCP connection.

Could you clarify how this PR addresses the multi-instance requirement from #534?

tprelle:
> @tprelle Thanks for the contribution! I noticed that issue #534 requests support for multiple Jira instances in a single MCP server, but this PR seems to primarily add programmatic configuration capabilities.
> 
> For true multi-instance support, wouldn't we need to implement instance-specific tool prefixes (e.g., jira_frontline_create_issue, jira_project_create_issue) instead of the current shared jira_ prefix? This would enable users to interact with multiple Jira instances through a single MCP connection.
> 
> Could you clarify how this PR addresses the multi-instance requirement from #534?

@sooperset i did the other way, the server expose multiple connection then the user can choose. Because it's easier to force a agent to choose the right tool (jira 1 or jira 2) than to prefix everything. 
But yes your are right, it's add programmatic configuration capabilities in order to have multiple jira mcp in the same server. Also also way easier to integrate into a bigger mcp server with other endpoint

sooperset:
@tprelle then what is the missing point that oauth can't support? Current MCP server can support multiple users with different authentication based on oauth, right?

tprelle:
Ah i think i understand what you says.
So edit get_confluence_fetcher and get_jira_fetcher fonction ?

sooperset:
@tprelle Looking at the current architecture:
- OAuth implementation supports multiple users but expects a single instance URL
- The `get_jira_fetcher()` creates user-specific configs but still uses the same base URL
- Tools share names regardless of instance

Your approach enables programmatic config but doesn't solve the core issue: tools from different instances would have identical names.

Since you're using Data Center (OAuth support coming in #562), I think the simplest solution is running separate MCP servers for each Jira instance. Each server would have its own URL and authentication.

sooperset:
Thank you for this substantial contribution! Multi-instance support is a commonly requested feature.

The codebase architecture has changed significantly since this was opened (BYOT multi-tenant headers, OAuth refactoring, new dependency injection). We're closing this PR, but the feature request (#533) remains open. We may revisit this with a fresh implementation that aligns with the current architecture. Thanks again for the effort!

PR #540, opened 2025-06-21
fix: allow unassigning issues

volkov:
Unassign fixed


sooperset:
@volkov  Thank you for the contribution!

PR #553, opened 2025-06-23
in JQL, need to quote project name (usually forgiven but not always)

amichai-gloat:
## Description

Quote project name when accessing JIRA. This solves a bug where you can't request stuff from a project whose name is also a reserved keyword in JQL.

## Changes

Quote project name in 2 JQLs.

## Testing

- [X] Manual checks performed: Running it through my agent

## Checklist

- [X] Code follows project style guidelines (linting passes).
- [X] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/projects.py
+++ b/src/mcp_atlassian/jira/projects.py
@@ -283,7 +283,7 @@ class ProjectsMixin(JiraClient, SearchOperationsProto):
         """
         try:
             # Use JQL to count issues in the project
-            jql = f"project = {project_key}"
+            jql = f'project = "{project_key}"'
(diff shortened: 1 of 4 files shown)


src/mcp_atlassian/jira/projects.py
```diff
@@ -283,7 +283,7 @@ def get_project_issues_count(self, project_key: str) -> int:
         """
         try:
             # Use JQL to count issues in the project
-            jql = f"project = {project_key}"
+            jql = f"project = \"{project_key}ֿ\""
```
sooperset: there is an extra Hebrew character (ֿ) after closing quote, which causes a syntax error

sooperset (approved):
Great work on this fix! Thanks @amichai-gloat

PR #565, opened 2025-06-25
Upgrade Python to 3.12 in Dockerfile

weekens:
## Description

In this PR I upgrade Python to `3.12` in Dockerfile. I actually need that to successfully combine `mcp-atlassian` with [mcpo](https://github.com/open-webui/mcpo/blob/main/Dockerfile) and expose Atlassian MCP server with OpenAPI HTTP interface. But I think it's useful overall given that I've tested it and saw it working.

## Changes

The Docker image now uses Python `3.12` instead of Python `3.10`.

## Testing

- [x] Manual checks performed: I've run the MCP Server wrapped into `mcpo`, made requests to connected LLM with questions about Confluence, and received valid responses.

## Checklist

- [ ] Tests added/updated for changes (no changes in tests).
- [ ] All tests pass locally (I did not configure connection to Atlassian in the project).
- [ ] Documentation updated (if needed) (no updates).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Use a Python image with uv pre-installed
-FROM ghcr.io/astral-sh/uv:python3.10-alpine AS uv
+FROM ghcr.io/astral-sh/uv:python3.12-alpine AS uv
 
 # Install the project into `/app`
 WORKDIR /app
@@ -34,7 +34,7 @@ RUN find /app/.venv -name '__pycache__' -type d -exec rm -rf {} + && \
     echo "Cleaned up .venv"
 
 # Final stage
-FROM python:3.10-alpine
+FROM python:3.12-alpine
 
 # Create a non-root user 'app'
 RUN adduser -D -h /home/app -s /bin/sh app


sooperset:
Hi, @weekens How about support CI/CD docker image build with multiple python versions? Changing the base version could have side effects for existing users.

weekens:
Hi @sooperset ! I see no need for supporting multiple Python versions if you tag the image with the version number on each release (I guess that's what happens in GitHub). The best practice for Docker setups is to always refer to a particular image version for exactly the reason you've mentioned. Guys who are using the fixed version (like `ghcr.io/sooperset/mcp-atlassian:0.11.8`) will not have sudden effects, and the ones who are using `latest` probably know their risks (at least the linter would tell them).

If you mean supporting a separate Docker image for the use cases like mine - I would argue that the best option here is to just clone the repo (which is what I did), so, there isn't really a need to mess with that.

Bottom line, I suggest to keep it simple.

sooperset:
@weekens While our codebase should work fine with Python 3.12, I'm concerned about:
- Unexpected upgrade for users pulling `:latest` who might have organizational requirements for Python 3.10
- No opt-in path for users to test compatibility with their specific setups first

How about we build multi-version images instead?
```
ghcr.io/sooperset/mcp-atlassian:0.x.x        # Python 3.10 (default)
ghcr.io/sooperset/mcp-atlassian:0.x.x-py312  # Python 3.12
```

This gives you Python 3.12 immediately while maintaining stability for existing users. We could switch the default in a future major version.

Would you be open to updating the PR for this approach?

weekens:
@sooperset , no, I suggest not to move forward with this approach. For me, it's over-engineering. The change you suggest will pollute your repository with complexity that you don't need. Let's close this PR. I've built my own local image with Python 3.12 anyway, I can live with it, and I can work with the fork going forward, if needed.

Anyway, thanks for this great project! Tremendously helpful!

PR #633, opened 2025-07-29
fix(jira): use issue type id for epic and subtask creation

BenjaminCarton:
This change updates the issue creation logic to use the issue type ID instead of its name when creating Epics, Tasks or Sub-tasks cause jira sucks and fail to create subtasks and tasks using their name at least in french.

## Description

When I use jira v2 api to create issues with tasks and subtasks using their name in issueType field creation fails (in french in my case). But when I try using their ID, it works. 

Fixes: #

## Changes

I modifed and renamed methods _find_epic_issue_type_name and _find_subtask_issue_type_name in order to return the issue id instead of the name.

I handled the new return type in jira issue creation 

I added a new methos _find_task_issue_type_id to duplicate this logic to tasks creation

## Testing

- [x] Manual checks performed: `i succeded in creating subtacks using le same worflow that was previously failing`

## Checklist

- [x] All tests pass locally.
- [x] Documentation updated (no needed).

### DIFF
--- a/tests/unit/jira/test_issue_creation_logic.py
+++ b/tests/unit/jira/test_issue_creation_logic.py
@@ -43,6 +43,12 @@ class ConcreteIssuesMixin(
     def get_required_fields(self, project_key, issue_type_name):
         pass
 
+    def get_issue_forms(self, issue_key):
+        pass
+
+    def get_form_details(self, issue_key, form_id):
+        pass
+
     def prepare_epic_fields(self, fields, summary, kwargs, project_key):
         pass
 


sooperset:
Merged via squash.

- Prepared head SHA: d9f1bbd2bf4e3219d45aa6ff49aa4e190348d58f
- Merge commit: 3893ac6bd7f1dce407c7c3ebed3c9cd508f4ebf3

Thanks @BenjaminCarton!

sooperset:
Thank you @BenjaminCarton for the fix and persistence here. We merged this after rebasing and tightening the implementation for current main.

Issue #698, opened 2025-09-17
[Bug]: Confluence V1 API used with Cloud +  ATLASSIAN_OAUTH_ENABLE=true


nealedj:
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/sooperset/mcp-atlassian/issues) to make sure this bug has not already been reported.
- [x] I have checked the [README](https://github.com/sooperset/mcp-atlassian/blob/main/README.md) for relevant information.

### Bug Description

This is a follow up bug to the fix applied in https://github.com/sooperset/mcp-atlassian/pull/579

If the server is configured with ATLASSIAN_OAUTH_ENABLE then is_cloud is never set to true even if CONFLUENCE_URL is set to the cloud URL.

Also - for ATLASSIAN_OAUTH_ENABLE is be evaluated `CONFLUENCE_URL` and `JIRA_URL` must be empty.

I'll raise a PR

### Steps to Reproduce

Start server with ATLASSIAN_OAUTH_ENABLE=true, CONFLUENCE_URL="", JIRA_URL="" (BYOT)

Execute confluence_get_page

### Expected Behavior

Confluence page returned

### Actual Behavior

```shell
Error returned: com.atlassian.confluence.api.service.exceptions.GoneException: This deprecated endpoint has been removed.
```

### mcp-atlassian Version

0.11.9

### Installation Method

From PyPI (pip install mcp-atlassian / uv add mcp-atlassian)

### Operating System

Linux (Specify distribution below if relevant)

### Python Version

3.13.2

### Atlassian Instance Type

Confluence Cloud

### Client Application

Librechat (using custom bridge to support oauth)

TahaGHZ:
I'm having the same issue, can you tell me more about your solution ?

PR #699, opened 2025-09-17
fix: support BYOT OAuth mode without requiring service URLs

nealedj:
## Description

Fixes: #698 

## Changes

- Allow missing CONFLUENCE_URL/JIRA_URL when ATLASSIAN_OAUTH_ENABLE is set
- Mark services as available in environment detection for BYOT mode
- Preserve original is_cloud behavior based on oauth_config.cloud_id presence

## Testing

- [x] Manual checks performed:  Tested using my integration from Librechat.

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/confluence/config.py
+++ b/src/mcp_atlassian/confluence/config.py
@@ -90,8 +90,12 @@ class ConfluenceConfig:
         oauth_config = get_oauth_config_from_env()
         auth_type = None
 
-        # Use the shared utility function directly
-        is_cloud = is_atlassian_cloud_url(url)
+        if url:
+            is_cloud = is_atlassian_cloud_url(url)
+        else:
+            # When no URL, let the @property is_cloud method handle detection
+            # It will check for OAuth cloud_id or return False
+            is_cloud = False
 
         if oauth_config:
             # OAuth is available - could be full config or minimal config for user-provided tokens
--- a/src/mcp_atlassian/jira/config.py
+++ b/src/mcp_atlassian/jira/config.py
@@ -90,8 +90,12 @@ class JiraConfig:
         oauth_config = get_oauth_config_from_env()
         auth_type = None
 
-        # Use the shared utility function directly
-        is_cloud = is_atlassian_cloud_url(url)
+        if url:
+            is_cloud = is_atlassian_cloud_url(url)
+        else:
+            # When no URL, let the @property is_cloud method handle detection
+            # It will check for OAuth cloud_id or return False
+            is_cloud = False
 
         if oauth_config:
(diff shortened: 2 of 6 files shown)


TahaGHZ:
Does this also fix the depricated Confluence v1 REST calls ?
I also ran into your same problem today with BYOT and confluence get page function tool

nealedj:
Assuming that you're running Confluence Cloud with BYOT then yes this should solve it.

kelvinwijaya:
@sooperset , I am supportive for this PR

Currently if we create API Token, we are forced to use v1 API (unscoped) which may get deprecated

Scope vs Unscoped PATH

Feature | Scoped PAT (Recommended) | Unscoped PAT (Legacy)
-- | -- | --
Security | Provides granular, least-privilege access by specifying exactly what actions the token can perform. | Grants all API permissions that the user who created the token has.
Permissions | Determined by specific scopes selected during creation (e.g., read:confluence-content). | Inherited from the token creator, with no option to limit access.
API Endpoints | Uses the central API endpoint format: https://api.atlassian.com/ex/confluence/{cloudId}/.... | Uses older, site-specific endpoint formats: https://{your-site-name}.atlassian.net/....
Expiration | Can be created with an optional expiration date (1–365 days) for added security. | Does not automatically expire unless configured by an administrator.
Creation | Created via the "Create API token with scopes" option in your Atlassian account's API tokens page. | Created via the "Create API token" option in your Atlassian account's API tokens page.
Status | Recommended and supported. | Considered legacy and is being deprecated.

nealedj:
Duplicate PR raised in alternative project https://github.com/SharkyND/mcp-atlassian/pull/13 since the maintainer of this one doesn't appear to be active any longer.

TahaGHZ:
In my current approach I'm not creating an API token (unscoped as you have mentioned), instead I'm using an Oauth2 app with granular and classic scoped to get a token and a refresh token, then these tokens are used along the MCP function tool calls.
This might be a better approach for you if you're authenticating multiple ATLASSIAN sites with your MCP instead of a 1on1 usage.

sooperset:
Thank you for the contribution @nealedj! Your approach for BYOT OAuth without requiring service URLs has been implemented.

This was merged in PR #956, which:
- Moves `ATLASSIAN_OAUTH_ENABLE` check from an `elif` branch to a `if not _is_setup` fallback, so it fires regardless of URL presence
- Adds 21 new tests covering all BYOT scenarios (no URL, with URL, cloud_id, value variations)

The config `from_env()` already handled the no-URL case correctly — the fix was in the service detection layer which was gating on URL presence.

Closing as superseded by #956. Thanks for identifying this!

PR #821, opened 2026-01-04
Fix high CPU usage on Windows by using SelectorEventLoop

### LINKED ISSUE #820: [Bug]: 1.2 Core CPU load on Idle
### Prerequisites

### Bug Description

When the MCP is idle the python process is consuming 1.2 CPU cores. With multiple copies of claude running it consumes 20-30% of the CPU on a 32 core Threadripper.

### Steps to Reproduce

1. run claude
2. check the cpu usage.

### Expected Behavior

CPU should be 0-.1% at idle.

### Actual Behavior

```shell
CPU is showing 4% at idle.
```

### mcp-atlassian Version

0.12.0

### Installation Method

### Operating System

Windows

### Python Version

3.12

### Atlassian Instance Type

Jira Cloud

### Client Application

Claude

### Additional Context

_No response_

nhouck:
Fixes #820

## Problem

On Windows, the MCP server consumes approximately 3-5% CPU per process even when completely idle. With multiple Claude Code instances running, this adds up to significant CPU usage (e.g., 4 instances = ~5 CPU cores constantly active).

## Root Cause

Python's default `ProactorEventLoop` on Windows uses IOCP (I/O Completion Ports), which can cause busy-waiting when combined with synchronous HTTP libraries like `requests` (used by `atlassian-python-api`) that internally use `select()` for socket operations.

Profiling with `py-spy` showed the issue in the Windows asyncio event loop polling:
```
ZwWaitForSingleObject (ntdll.dll)
select (WS2_32.dll)
windows_events.py:322
```

## Solution

Switch to `WindowsSelectorEventLoopPolicy` on Windows, which uses the traditional selector-based approach. This is more efficient when the server is idle and is compatible with libraries that expect a selector-based event loop.

```python
if sys.platform == "win32":
    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```

## Results

- **Before**: ~3-5% CPU usage per idle MCP server process
- **After**: Near 0% CPU usage when idle

## Testing

Tested on Windows 11 with Python 3.13. The fix is applied early in module initialization and only affects Windows systems.

### DIFF
--- a/src/mcp_atlassian/__init__.py
+++ b/src/mcp_atlassian/__init__.py
@@ -7,6 +7,13 @@ from importlib.metadata import PackageNotFoundError, version
 import click
 from dotenv import load_dotenv
 
+# Fix high CPU usage on Windows - use SelectorEventLoop instead of ProactorEventLoop
+# ProactorEventLoop uses IOCP which can cause busy-waiting when combined with
+# synchronous libraries (like requests) that use select() for socket operations.
+# This reduces idle CPU usage from ~3-5% per process to near zero.
+if sys.platform == "win32":
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+
 from mcp_atlassian.utils.env import is_env_truthy
 from mcp_atlassian.utils.lifecycle import (
     ensure_clean_exit,


sooperset:
Thank you for investigating #820! After further analysis, we found the root cause is in `SSLIgnoreAdapter.init_poolmanager()` using `block=False`, which causes urllib3 to busy-wait when checking pool availability.

PR #825 addressed this by forcing `block=True` and has been merged. Since that's the correct fix, I'm closing this PR.

Your investigation was valuable in bringing attention to this issue - the Windows-specific event loop approach would have been a reasonable hypothesis, but profiling showed the issue was in the SSL connection pooling layer rather than the asyncio event loop.

sooperset:
Reopened this PR after reviewing PR #825.

**Issue with PR #825:** The `SSLIgnoreAdapter` fix only applies when `ssl_verify=False`. Since the original reporter uses Jira Cloud with default settings (SSL enabled), that code path was never executed.

**Before merging this PR**, could you confirm:
1. Did you test this fix on the actual issue (not just profiling)?
2. Did CPU usage drop from ~3-5% to ~0% after applying this change?

cc @Dragonfire3900 (original issue reporter) - could you test this branch and confirm it resolves the high CPU usage?

nhouck:
> Reopened this PR after reviewing PR #825.
> 
> **Issue with PR #825:** The `SSLIgnoreAdapter` fix only applies when `ssl_verify=False`. Since the original reporter uses Jira Cloud with default settings (SSL enabled), that code path was never executed.
> 
> **Before merging this PR**, could you confirm:
> 
>     1. Did you test this fix on the actual issue (not just profiling)?
> 
>     2. Did CPU usage drop from ~3-5% to ~0% after applying this change?
> 
> 
> cc @Dragonfire3900 (original issue reporter) - could you test this branch and confirm it resolves the high CPU usage?

I updated a local copy of the code with this fix and have been running it the past 1.5 days and it's working great. 0-.1% cpu usage now. Not sure if that's what you meant by #1.

Thanks SO Much for making this project which fixes the constant Oauth token expiration with the official Atlassian version of this MCP. With the API auth approach this is working amazing!

sooperset:
Thanks @nhouck for testing and confirming the fix! Merged.

PR #848, opened 2026-01-12
Add Jira Cloud search pagination (nextPageToken)

solganik:
Problem: The existing jira_search tool exposes start_at/limit like the legacy offset-based API, but Jira Cloud /rest/api/3/search/jql ignores startAt and requires nextPageToken. With no page token exposed (and the tool’s limit capped), callers could not iterate past the first page, so plain search was not enough to fetch all matching issues.

Fix: Add a page-based search tool that returns next_page_token/is_last (and accepts page_token/page_size) so clients can reliably page through all results on Jira Cloud.

## Description

Fixes: #

## Changes

-
-
-

## Testing

## Checklist

- [ ] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/search.py
+++ b/src/mcp_atlassian/jira/search.py
@@ -18,6 +18,123 @@ logger = logging.getLogger("mcp-jira")
 class SearchMixin(JiraClient, IssueOperationsProto):
     """Mixin for Jira search operations."""
 
+    def search_issues_page(
+        self,
+        jql: str,
+        fields: list[str] | tuple[str, ...] | set[str] | str | None = None,
+        page_token: str | None = None,
+        page_size: int = 50,
+        expand: str | None = None,
+        projects_filter: str | None = None,
+    ) -> JiraSearchResult:
+        """
+        Search for issues using JQL and return a single page.
+
+        Cloud environments use the v3 enhanced JQL search endpoint that returns a
+        `nextPageToken` for token-based pagination. Server/DC environments use
+        offset-based pagination.
+
+        Args:
+            jql: JQL query string
+            fields: Fields to return (comma-separated string, list, tuple, set, or "*all")
+            page_token: Cloud pagination token returned by the previous call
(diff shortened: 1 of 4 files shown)


src/mcp_atlassian/jira/search.py
```diff
@@ -18,6 +18,123 @@
...
+                    response,
+                    base_url=self.config.url,
+                    requested_fields=fields_param,
+                )
+                return search_result
+
+            # Server/DC: offset-based paging
+            page_size = min(max(page_size, 1), 50)
+            response = self.jira.jql(
+                jql, fields=fields_param, start=0, limit=page_size, expand=expand
```
sooperset, line 109: `start=0` is hardcoded here, which means Server/DC pagination does not actually work.

The `page_token` parameter is accepted but ignored for Server/DC, so users will always receive the first page regardless of pagination attempts.

**Suggestions:**
- Add a `start_at` parameter for Server/DC offset-based pagination, or
- Make this method Cloud-only and return a clear error/warning for Server/DC

sooperset, line 109: `start=0` hardcoded - Server/DC pagination broken.

Add `start_at` param or make this Cloud-only.

src/mcp_atlassian/servers/jira.py
```diff
@@ -247,11 +257,170 @@ async def search(
...
+
+    # fetch_all: internally page through all results
+    all_issues: list[dict[str, Any]] = []
+
+    if jira.config.is_cloud:
+        # Cloud: token-based pagination
+        page_token: str | None = None
+        while True:
+            page = jira.search_issues_page(
+                jql=jql,
```
sooperset, line 280: Unbounded loop risk. Add `max_total_results` cap.

sooperset:
Thank you for the contribution, @solganik! This has been reimplemented in #1079 with updated tests and project conventions. Your work is credited via Co-authored-by trailer.

PR #874, opened 2026-01-26
fix(jira): use correct API version for remote issue links on Server/DC

major:
## Summary

- Fix remote issue link creation to use API v2 for Server/Data Center and v3 for Cloud
- Add `make_issue_data` test fixture to reduce boilerplate in test files
- Add test coverage for Server/DC remote link creation

## Details

The `create_remote_issue_link` method was hardcoded to use the v3 API endpoint, which only works on Jira Cloud. This caused failures when creating remote issue links on Jira Server/Data Center deployments.

This fix adds conditional logic to select the appropriate API version based on the deployment type, following the same pattern used elsewhere in the codebase (e.g., `jira/issues.py`, `jira/search.py`).

### DIFF
--- a/src/mcp_atlassian/jira/links.py
+++ b/src/mcp_atlassian/jira/links.py
@@ -177,7 +177,11 @@ class LinksMixin(JiraClient):
 
         try:
             # Create the remote issue link using the Jira API
-            endpoint = f"rest/api/3/issue/{issue_key}/remotelink"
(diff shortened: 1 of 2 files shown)


major:
Hmm, still hoping to get this in!

sooperset:
Thanks for the contribution! Good catch on the API version for remote issue links. Merged to main.

Issue #852, opened 2026-01-14
jira_get_issue does not return comments


dsurchisactive:
## Description

When using `jira_get_issue` with `comment_limit` set to a positive value, the response does not include any comments, even when comments exist on the issue.

## Steps to Reproduce

1. Call `jira_get_issue` with an issue key that has comments
2. Set `comment_limit` to 20 or higher
3. Optionally set `fields` to `*all` and `expand` to `renderedFields`

## Expected Behavior

The response should include a `comments` field with the issue's comments.

## Actual Behavior

The response contains issue details (summary, description, status, assignee, etc.) but no comments field is returned, even though comments exist on the issue (verified via Jira UI).

## Environment

- Jira Cloud (atlassian.net)
- mcp-atlassian installed via `uvx mcp-atlassian`

guyuga:
having the same issue

guyuga:
found a workaround, try calling with the following: 
{
  "issue_key": "PROJ-123",
  "comment_limit": 20,
  "fields": "summary,status,description,comment"
}

Troubladore:
I believe this may be addressed by PR #1125, which adds an `include` parameter to `jira_get_issue` — passing `include="comments"` (or `include="all"`) should return comments inline.

If you'd like to try the fix, it's available on the [`community/main`](https://github.com/Troubladore/mcp-atlassian/tree/community/main) branch, which integrates reviewed community PRs that are awaiting merge:
```
pip install git+https://github.com/Troubladore/mcp-atlassian@community/main
```
If you get a chance to test it, I'd appreciate knowing whether it resolves your case.

PR #876, opened 2026-01-26
fix(jira): auto-include comment field when comment_limit > 0

major:
## Description

When using `jira_get_issue` with `comment_limit` set to a positive value, comments were not returned unless the `comment` field was explicitly included in the `fields` parameter.

**Root causes:**
1. `comment` is not in `DEFAULT_READ_JIRA_FIELDS`, so the API request didn't include the comment field
2. The existing comparison to detect default fields used string comparison on a set, which fails due to Python hash randomization across processes (#877)

Fixes: #852
Fixes: #877

## Changes

- Add logic to automatically include `comment` in the fields parameter when `comment_limit` is positive or `None` (unlimited)
- Fix set comparison to use set equality instead of string equality (prevents hash randomization issues)
- Add unit tests to verify the fix and ensure `comment` is excluded when `comment_limit=0`

## Testing

- [x] Manual checks performed: Verified that `comment` field is included in API request when `comment_limit > 0`

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -90,11 +90,11 @@ class IssuesMixin(
             elif isinstance(fields_param, list | tuple | set):
                 fields_param = ",".join(fields_param)
 
-            # Ensure necessary fields are included based on special parameters
-            if (
-                fields_param == ",".join(DEFAULT_READ_JIRA_FIELDS)
-                or fields_param == "*all"
-            ):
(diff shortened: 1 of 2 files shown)


major:
Hmm, still hoping to get this in!

sooperset:
Thanks for the contribution! Auto-including the comment field when `comment_limit` is positive is a nice quality-of-life improvement. Merged to main (rebased with conflict resolution).

major:
@sooperset No problem! This MCP server is really handy!

PR #902, opened 2026-02-09
fix(jira): use ASCII transliteration for user lookup

brettfire:
## Summary
Use `unidecode` for ASCII transliteration in `_lookup_user_directly()` to handle display names with diacritics like Polish "ł" or German "ü".

## Problem

The `_lookup_user_directly` function uses string comparison to match users returned by the Jira API. When a user's display name contains Unicode characters (e.g., "Paweł Kowalczyk"), searching with ASCII approximations (e.g., "Pawel Kowalczyk") fails despite the API finding the correct user.

### Root Cause

The previous implementation used `casefold()` which only handles case normalization, not character transliteration:

```python
"Paweł".casefold()  # → "paweł" (ł preserved)
"Pawel".casefold()  # → "pawel"
# These don't match!
```

### Reproduction

```python
# API correctly returns the user with fuzzy matching
curl "https://INSTANCE.atlassian.net/rest/api/3/user/search?query=Pawel%20Kowalczyk"
# Returns: [{"displayName": "Paweł Kowalczyk", "accountId": "712020:..."}]

# But MCP tool fails because comparison fails
jira_get_user_profile(user_identifier="Pawel Kowalczyk")
# {"success": false, "error": "Could not determine how to look up user..."}

# Unicode name works
jira_get_user_profile(user_identifier="Paweł Kowalczyk")
# {"success": true, ...}
```

## Solution

Use `unidecode` library to transliterate Unicode to ASCII before comparison:

```python
from unidecode import unidecode

def normalize_text(text: str) -> str:
    if not text:
        return ""
    return unidecode(text).casefold()
```

This converts:
- Polish: ł → l, ó → o, ą → a
- German: ü → u, ö → o, ß → ss
- And many other character sets

## Changes

- Add `unidecode>=1.3.0` to dependencies
- Update `normalize_text()` in `jira/users.py` to use unidecode
- Existing tests pass, ASCII approximations now work

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ dependencies = [
     "keyring>=25.6.0",
     "cachetools>=5.0.0",
     "types-cachetools>=5.5.0.20240820",
+    "unidecode>=1.3.0",
 ]
 [[project.authors]]
 name = "sooperset"
--- a/src/mcp_atlassian/jira/users.py
+++ b/src/mcp_atlassian/jira/users.py
@@ -2,7 +2,10 @@
 
 import logging
 import re
+import unicodedata
 from typing import TYPE_CHECKING, TypeVar
+
+from unidecode import unidecode
 
 import requests
 from requests.exceptions import HTTPError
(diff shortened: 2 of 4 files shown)


sooperset:
Thanks for the contribution! ASCII transliteration via unidecode for Unicode user lookup is a solid approach — properly handling names like Polish "ł" matching "l". Merged to main (rebased with type annotation fix).

PR #903, opened 2026-02-09
fix(jira): extract ORDER BY clause before adding project filter

brettfire:
## Summary

  Fix invalid JQL generation when combining `projects_filter` with queries containing `ORDER BY` clause.

  ## Problem

  When a JQL query contains ORDER BY and `projects_filter` is applied, the library generates invalid JQL:

  ```python
  # Input
  jql = 'assignee = "User" ORDER BY updated DESC'
  projects_filter = "AI"

  # Generated (BROKEN)
  '(assignee = "User" ORDER BY updated DESC) AND project = "AI"'
  # Error: "Expecting ')' but got 'ORDER'"

  # Should be
  '(assignee = "User") AND project = "AI" ORDER BY updated DESC'
```

## Reproduction
```
  # Broken JQL - API rejects it
  curl -X POST "https://INSTANCE.atlassian.net/rest/api/3/search/jql" \
    -d '{"jql": "(assignee = \"User\" ORDER BY updated DESC) AND project = \"PROJ\"}'
  # Error: "Expecting ')' but got 'ORDER'"

  # Correct JQL - works
  curl -X POST "https://INSTANCE.atlassian.net/rest/api/3/search/jql" \
    -d '{"jql": "(assignee = \"User\") AND project = \"PROJ\" ORDER BY updated DESC"}'
  # Returns results
```

##  Solution

  Extract ORDER BY clause before wrapping JQL with project filter, then append at end.

## Changes

  - Add import re for regex matching
  - Update search_issues() in jira/search.py to extract and re-append ORDER BY
  - Add tests for trailing ORDER BY with project filter

## Test Plan

  - Test JQL with trailing ORDER BY + projects_filter generates valid JQL
  - Test case-insensitive ORDER BY matching
  - Test backward compatibility (existing behavior unchanged)
  - All 53 related tests pass

### DIFF
--- a/src/mcp_atlassian/jira/search.py
+++ b/src/mcp_atlassian/jira/search.py
@@ -1,6 +1,7 @@
 """Module for Jira search operations."""
 
 import logging
+import re
 from typing import Any
 
 import requests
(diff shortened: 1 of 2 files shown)


sooperset:
Thanks for the contribution! Good catch on the ORDER BY + project filter interaction — extracting the clause before wrapping is the correct fix. Merged to main.

PR #363, opened 2025-05-02
Feature/bitbucket integration

timelfrink:
## Description                                                                                                                                                                                                                              

This PR adds basic Bitbucket Server integration to the MCP Atlassian toolkit. The integration enables users to browse repositories, work with pull requests, search code, inspect commits, and check build statuses.                                                                                                         

## Changes                                                                                                                                                                                                                                                                                                                   

This PR adds the following Bitbucket Server tools:                                                                                                                                                                                                                                                                           

- `bitbucket_get_pull_request`: Get details of a specific pull request                                                                                                                                                                                                                                                       
- `bitbucket_add_comment`: Add a comment to a pull request                                                                                                                                                                                                                                                                   
- `bitbucket_get_diff`: Get diff for a pull request showing code changes                                                                                                                                                                                                                                                     
- `bitbucket_get_reviews`: Get reviews for a pull request                                                                                                                                                                                                                                                                    
- `bitbucket_get_activities`: View pull request activities history                                                                                                                                                                                                                                                           
- `bitbucket_get_branches`: List all branches in a repository with filtering options                                                                                                                                                                                                                                         
- `bitbucket_get_branch_commits`: Retrieve commits from a specific branch                                                                                                                                                                                                                                                    
- `bitbucket_get_commit`: Get detailed information about a specific commit                                                                                                                                                                                                                                                   
- `bitbucket_get_commit_changes`: View file changes in a specific commit                                                                                                                                                                                                                                                     
- `bitbucket_get_build_status`: Check CI build status for commits                                                                                                                                                                                                                                                            
- `bitbucket_search_code`: Search code content in repositories                                                                                                                                                                                                                                                               
- `bitbucket_search_repositories`: Search for repositories                                                                                                                                                                                                                                                                   
- `bitbucket_get_file_content`: Get the content of a file from a repository                                                                                                                                                                                                                                                  

Additional changes:                                                                                                                                                                                                                                                                                                          
- Added configuration options for Bitbucket Server in .env.example                                                                                                                                                                                                                                                           
- Updated README.md with complete documentation for all Bitbucket tools                                                                                                                                                                                                                                                      
- Added workflow examples for code navigation and pull request reviews                                                                                                                                                                                                                                                       

## Testing                                                                                                                                                                                                                                                                                                                   

- [x] Unit tests added for all new functionality                                                                                                                                                                                                                                                                             
- [x] Integration tests for API interactions                                                                                                                                                                                                                                                                                 
[...]

PR #917, opened 2026-02-16
Add Zephyr integration

dloiacono:
feat: Add Zephyr Squad (Jira plugin) support

Add comprehensive Zephyr Squad integration for on-premise Jira installations
with the Zephyr plugin. This complements the existing Zephyr Scale support
and uses Jira's ZAPI endpoints (/rest/zapi/latest/).

## Description

This PR adds full support for Zephyr Squad, the Jira test management plugin commonly used in on-premise deployments. Zephyr Squad differs from Zephyr Scale in that it's accessed through Jira's REST API rather than as a standalone service.

## Changes

- Added `ZephyrSquadMixin` class with 7 test management methods
- Integrated mixin into `JiraFetcher` for seamless Jira client usage
- Created 6 new MCP tools: `zephyr_squad_get_cycles`, `zephyr_squad_create_cycle`, `zephyr_squad_get_executions`, `zephyr_squad_create_execution`, `zephyr_squad_execute_test`, `zephyr_squad_get_test_steps`
- Updated README to distinguish Zephyr Scale vs Squad with compatibility matrix

## Testing

- [x] Manual checks performed: Verified build succeeds, linting passes, tested with MCP inspector

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -116,8 +116,11 @@ CONFLUENCE_API_TOKEN=your_confluence_api_token_here
 # =============================================
 # Only applicable if your JIRA_URL/CONFLUENCE_URL points to a Server/DC instance (not *.atlassian.net).
 # Default is true. Set to false if using self-signed certificates (not recommended for production environments).
+# Note: Enterprise CA certificates in the OS trust store are trusted automatically via truststore.
 #JIRA_SSL_VERIFY=true
 #CONFLUENCE_SSL_VERIFY=true
+# Set to false to disable OS trust store integration and use the bundled certifi CA bundle instead.
+#MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE=true
 #ZEPHYR_SSL_VERIFY=true
 
 # --- CLIENT CERTIFICATE AUTHENTICATION (mTLS) ---
--- a/docs.json
+++ b/docs.json
@@ -2,6 +2,9 @@
   "$schema": "https://mintlify.com/docs.json",
   "theme": "mint",
   "name": "MCP Atlassian",
+  "seo": {
+    "indexHiddenPages": false
+  },
   "favicon": "/docs/favicon.svg",
   "colors": {
     "primary": "#0052CC",
--- a/docs/authentication.mdx
+++ b/docs/authentication.mdx
@@ -58,7 +58,8 @@ CONFLUENCE_PERSONAL_TOKEN=your_personal_access_token
 ```
 
 <Note>
-For self-signed certificates, set `JIRA_SSL_VERIFY=false` and/or `CONFLUENCE_SSL_VERIFY=false`.
+Enterprise CA certificates in the OS trust store (Windows Certificate Store, macOS Keychain, Linux system CAs) are trusted automatically.
+For self-signed certificates not in the OS trust store, set `JIRA_SSL_VERIFY=false` and/or `CONFLUENCE_SSL_VERIFY=false`.
 </Note>
 
 ## OAuth 2.0 (Cloud) - Advanced
--- a/docs/configuration.mdx
+++ b/docs/configuration.mdx
@@ -127,6 +127,7 @@ This guide covers IDE integration, environment variables, and advanced configura
 | `CONFLUENCE_API_TOKEN` | Confluence API token (Cloud) |
 | `CONFLUENCE_PERSONAL_TOKEN` | Confluence Personal Access Token (Server/DC) |
 | `CONFLUENCE_SSL_VERIFY` | SSL verification (`true`/`false`) |
+| `MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE` | Use OS native trust store (`true`/`false`, default: `true`) |
 
 ### Filtering Options
 
--- a/docs/troubleshooting.mdx
+++ b/docs/troubleshooting.mdx
@@ -18,11 +18,25 @@ description: "Fix auth errors (401/403), field issues, rate limiting, SSL proble
   </Accordion>
 
   <Accordion title="SSL Certificate Issues">
-    For Server/Data Center with self-signed certificates:
+    mcp-atlassian uses the OS native trust store by default (via
+    [truststore](https://truststore.readthedocs.io/)), so certificates
+    signed by enterprise CAs installed in Windows Certificate Store,
+    macOS Keychain, or the Linux system CA bundle are trusted
+    automatically.
+
+    If you still see SSL errors with **self-signed** certificates that
+    are **not** in the OS trust store, disable verification:
 
     ```bash
     JIRA_SSL_VERIFY=false
     CONFLUENCE_SSL_VERIFY=false
+    ```
+
+    To disable the OS trust store integration (fall back to the bundled
+    certifi CA bundle):
+
+    ```bash
+    MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE=false
     ```
   </Accordion>
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     "cachetools>=5.0.0",
     "types-cachetools>=5.5.0.20240820",
     "unidecode>=1.3.0",
+    "truststore>=0.10.0",
     "tzdata>=2024.1; platform_system=='Windows'",
 ]
 [[project.authors]]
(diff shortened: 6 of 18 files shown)


KaiWissel:
Cool feature. Looking forward to it.

dloiacono:
@sooperset can you review it so we can merge it ?

thekub3r:
Nice addition! Clear Zephyr Squad support, well-structured tools, and good alignment with the existing Jira client—this really strengthens the test management story for on‑prem setups.

src/mcp_atlassian/zephyr/testcases.py
```diff
@@ -0,0 +1,184 @@
...
+            "projectKey": project_key,
+            "name": name,
+        }
+
+        if objective:
+            payload["objective"] = objective
+        if precondition:
+            payload["precondition"] = precondition
+        if priority:
+            payload["priority"] = priority
```
sooperset: The Zephyr Scale v2 API expects `priorityName` and `statusName` for POST /testcases, not `priority` and `status`. Per the [official docs](https://support.smartbear.com/zephyr-scale-cloud/api-docs/), the correct field names are:

```python
if priority:
    payload["priorityName"] = priority
if status:
    payload["statusName"] = status
```

As-is, the priority and status values are silently ignored and defaults are used.

The same issue exists in `update_test_case` (line 137-139), though note that PUT uses object format (`{"id": N}`) for these fields — see the separate comment about PUT semantics.

src/mcp_atlassian/zephyr/testcycles.py
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        if description:
+            payload["description"] = description
+        if planned_start_date:
+            payload["plannedStartDate"] = planned_start_date
+        if planned_end_date:
+            payload["plannedEndDate"] = planned_end_date
+        if status:
+            payload["status"] = status
```
sooperset: Same field name mismatch as testcases — the v2 API expects `statusName`, not `status`:

```python
if status:
    payload["statusName"] = status
```

Also applies to `update_test_cycle` at line 133.

src/mcp_atlassian/zephyr/client.py
```diff
@@ -0,0 +1,175 @@
...
+            method=method, url=url, params=params, json=json, data=data
+        )
+
+        try:
+            response.raise_for_status()
+        except requests.HTTPError as e:
+            logger.error(f"HTTP error: {e}, Response: {response.text}")
+            raise
+
+        if response.status_code == 204:  # No content
```
sooperset: Several Zephyr Scale endpoints return success with an empty body — for example, `POST /testexecutions` returns `201 Created` with no body, and `PUT /testcases/{key}` returns `200 OK` with no body (both confirmed in the API docs).

Currently only `204` is handled as empty. For other 2xx codes with empty body, `response.json()` will raise `JSONDecodeError` and the successful API call will appear as an error.

Consider:

```python
if response.status_code == 204 or not response.text.strip():
    return {}
```

src/mcp_atlassian/zephyr/testcases.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        payload: dict[str, Any] = {}
+
+        if name:
+            payload["name"] = name
+        if objective:
+            payload["objective"] = objective
+        if precondition:
+            payload["precondition"] = precondition
+        if priority:
```
sooperset: The Zephyr Scale v2 API docs explicitly state for `PUT /testcases/{key}`:

> "Please take into account that for each non-specified field the value will be cleared."

Sending a sparse payload (only the changed fields) will wipe all other fields on the test case. The API requires the full object representation on PUT.

Additionally, PUT expects `priority` and `status` as objects (`{"id": N, "self": "..."}`) not as name strings — the format differs from POST which uses `priorityName`/`statusName`.

Consider either:
1. GET the current test case first, merge changes, then PUT the complete object
2. Or document this as a known limitation

src/mcp_atlassian/zephyr/config.py
```diff
@@ -0,0 +1,212 @@
...
+        username = os.getenv("ZEPHYR_USERNAME")
+        password = os.getenv("ZEPHYR_PASSWORD")
+        personal_token = os.getenv("ZEPHYR_PERSONAL_TOKEN")
+
+        # Check for OAuth configuration
+        oauth_config = get_oauth_config_from_env()
+        auth_type = None
+
+        # Use the shared utility function directly
+        is_cloud = is_atlassian_cloud_url(url)
```
sooperset, line 95: `is_atlassian_cloud_url` checks for `.atlassian.net` domains, but Zephyr Scale Cloud uses `api.zephyrscale.smartbear.com`. This means the documented Cloud setup (`ZEPHYR_URL=https://api.zephyrscale.smartbear.com/v2` + `ZEPHYR_API_TOKEN`) is misclassified as Server/DC, causing valid bearer-token credentials to be rejected.

The same issue exists in `utils/environment.py:get_available_services()` which uses the same check — so even if the config issue is worked around, Zephyr tools would be filtered out at tool-list time.

Consider adding Zephyr Scale Cloud URL detection (e.g., checking for `zephyrscale.smartbear.com` or `api.atlassian.com/ex/zephyr`) alongside the existing Atlassian Cloud check.

src/mcp_atlassian/zephyr/testexecutions.py
```diff
@@ -0,0 +1,180 @@
...
+        if test_case_key:
+            params["testCase"] = test_case_key
+
+        return self.get("testexecutions", params=params)
+
+    def create_test_execution(
+        self,
+        project_key: str,
+        test_case_key: str,
+        test_cycle_key: str | None = None,
```
sooperset: (minor) Per the Zephyr Scale v2 API docs, `testCycleKey` is a **required** field for `POST /testexecutions`. Making it optional here could lead to 400 errors when callers omit it.

src/mcp_atlassian/zephyr/testcases.py
dloiacono: done

src/mcp_atlassian/zephyr/testcycles.py
dloiacono: done

src/mcp_atlassian/zephyr/client.py
dloiacono: done

src/mcp_atlassian/zephyr/testcases.py
dloiacono: done

src/mcp_atlassian/zephyr/config.py
dloiacono: done

src/mcp_atlassian/zephyr/testexecutions.py
dloiacono: done

sooperset:
Hi @dloiacono! Just checking in on this PR.

I can see that @sooperset left a detailed review on Feb 25 covering several API spec issues (field name mismatches like `priority` vs `priorityName`, PUT semantics clearing unspecified fields, empty response body handling, and Zephyr Scale Cloud URL detection). You replied "done" to all 6 inline comments and pushed a fix commit (`fix(zephyr): make test_cycle_key required and fix v2 API field mappings`) that appears to address those items.

However, CI is currently passing on the latest commit — that is good — but could you confirm:

1. Have all the issues from the Feb 25 review been fully addressed? In particular, the GET-merge-PUT pattern for update operations to prevent field clearing — is that implemented?
2. The latest commits include merge conflict resolutions from `main`. Have you verified the merge resolutions are correct, especially in `dependencies.py`, `main.py`, and `urls.py`?

Let us know if you need any guidance. Thanks for your patience on this one!

dloiacono:
Hi @sooperset I provided all the fix for your review.
all merge conflics have been solved as well.

sooperset:
Thanks for the great contribution, @dloiacono! Zephyr Scale and Squad support is a valuable addition for test management users. Appreciate your patience working through the review feedback.

sooperset:
Hey @dloiacono, after reviewing this more carefully, I realized Zephyr Scale and Zephyr Squad are SmartBear products with their own separate API, auth, and config — not Atlassian products. This falls outside the scope of mcp-atlassian, and I should have caught this before merging. Sorry about that.

We previously declined Bitbucket integration for the same reason (see #363) — keeping the project focused on core Atlassian products (Jira, Confluence) so we can maintain quality and stability.

I'll be reverting this merge. A dedicated package (e.g. `mcp-zephyr`) would be a better home for this integration. Really appreciate all the work you put into this.

PR #919, opened 2026-02-16
fix: return attachments as embedded resources instead of writing to filesystem

### LINKED ISSUE #920: jira_download_attachments is broken when server runs remotely (Docker)
## Description

The `jira_download_attachments` tool writes files to the server's local filesystem via a `target_dir` parameter. When the MCP server runs remotely (e.g. in a Docker container), the downloaded files are inaccessible to clients since they exist inside the container's ephemeral filesystem.

## Steps to Reproduce

1. Deploy mcp-atlassian via Docker using the official image (`ghcr.io/sooperset/mcp-atlassian:latest`)
2. Call `jira_download_attachments` with an issue key and a `target_dir` (e.g. `/tmp/attachments`)
3. Observe the response — it reports "success" with file paths like `/tmp/attachments/file.png`
4. Check the local machine — the directory is empty because the files were written inside the now-dead container

## Expected Behavior

Attachment contents should be returned over the MCP protocol so they are accessible to any client, regardless of where the server is deployed.

## Actual Behavior

The tool reports success but the files are written to the server's local filesystem. When running in Docker with `--rm`, the container is destroyed after the request and the files are lost. The client receives only a JSON message with unusable file paths pointing to a dead container.

## Environment

- mcp-atlassian Docker image: `ghcr.io/sooperset/mcp-atlassian:latest`
- Transport: stdio via Docker
- MCP client: VS Code with GitHub Copilot

## Suggested Fix

Return attachments as base64-encoded `EmbeddedResource` objects (using `BlobResourceContents` from `mcp.types`) instead of writing to the filesystem. This is the standard MCP pattern for returning binary data. See PR #919.

WZHBEDZ:
## Summary

The `jira_download_attachments` tool writes files to the server's local filesystem via a `target_dir` parameter. When the MCP server runs remotely (e.g. in a Docker container), the downloaded files are inaccessible to clients since they exist inside the container's ephemeral filesystem.

This PR returns attachment contents inline as base64-encoded `EmbeddedResource` objects over the MCP protocol, eliminating the need for filesystem access on the server.

## Problem

When `jira_download_attachments` is called on a Docker-deployed MCP server:
1. The tool reports **success** and lists file paths like `/tmp/attachments/file.png`
2. Those paths exist **inside the container**, which is already gone (`--rm`)
3. The client receives a JSON success message with unusable file paths

## Solution

- Add `fetch_attachment_content()` method — downloads attachment bytes into memory
- Add `get_issue_attachment_contents()` method — fetches all attachments for an issue as raw bytes
- Rewrite `download_attachments` server tool to return `list[TextContent | EmbeddedResource]` instead of writing to disk
- Remove `target_dir` parameter (no longer needed)
- Each attachment is returned as a `BlobResourceContents` with base64-encoded data

The existing filesystem-based methods (`download_attachment`, `download_issue_attachments`) are preserved for backward compatibility.

## Testing

- 9 new unit tests added for the new methods
- All 33 attachment tests pass
- Pre-commit checks (ruff, ruff-format, mypy) pass
- Manually tested via Docker: 2/2 attachments from a real Jira issue returned correctly as embedded images

Fixes #920

### DIFF
--- a/src/mcp_atlassian/jira/attachments.py
+++ b/src/mcp_atlassian/jira/attachments.py
@@ -1,6 +1,7 @@
 """Attachment operations for Jira API."""
 
 import logging
+import mimetypes
 import os
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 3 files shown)


sooperset:
Thanks for the contribution! Returning attachments as EmbeddedResource (base64) is exactly the right pattern for MCP — no filesystem dependency, works in Docker. Merged to main (rebased with conflict resolution).

PR #925, opened 2026-02-17
fix(jira): assign issue post-creation to prevent silent assignee drop

### LINKED ISSUE #924: Assignee silently ignored when creating issues via jira_create_issue
### Bug Description

When passing an `assignee` parameter to `jira_create_issue`, the issue is created successfully but the assignee is not set. No error is returned — the issue simply appears as "Unassigned." Setting the assignee via a separate `jira_update_issue` call works correctly.

### Steps to Reproduce

1. Call `jira_create_issue` with a valid `assignee` (email, display name, or account ID)
2. Observe the created issue has no assignee

### Expected Behavior

The created issue should be assigned to the specified user.

### Actual Behavior

The issue is created as "Unassigned." No error or warning is surfaced to the caller.

### Root Cause

`create_issue()` adds the assignee to the `fields` dict and passes it to `self.jira.create_issue(fields=fields)`, but certain Jira Server/DC configurations don't honor the assignee field during issue creation. The Jira REST API's create endpoint doesn't reliably process assignee in the fields payload across all instance types.

### Atlassian Instance Type

- [x] Jira Server / Data Center
- [ ] Jira Cloud (untested, potentially affected)

major:
## Description

Fixes the assignee being silently ignored when creating issues via `jira_create_issue`.

The `create_issue()` method adds the assignee to the `fields` dict passed to the Jira REST API's create endpoint, but certain Jira Server/DC configurations don't honor this field during creation. The issue gets created as "Unassigned" with no error.

This adds a post-creation `jira.assign_issue()` call as a safety net — a dedicated REST API endpoint (`PUT /rest/api/2/issue/{key}/assignee`) that reliably works across all Jira instance types. The assignee is kept in the create fields (belt & suspenders) so there's zero regression risk for Cloud users where it already works. This follows the same two-step pattern already used for epic field updates.

Fixes #924

## Changes

- Add post-creation `assign_issue()` call as a safety net for configs where create-time assignee is silently dropped
- Keep assignee in the create `fields` dict (no regression for Cloud or configs where it already works)
- Update existing tests to verify both the create-time field and the post-creation assignment

## Testing

- [x] Unit tests updated and passing (44/44)
- [ ] Integration tests
- [x] Manual testing (confirmed on Jira Server/DC)

## Checklist

- [x] Code follows the project's style guidelines
- [x] Tests have been added/updated
- [x] All tests pass
- [x] Documentation updated (if applicable)

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -599,14 +599,18 @@ class IssuesMixin(
             if description:
                 fields["description"] = self._markdown_to_jira(description)
 
-            # Add assignee if provided
+            # Resolve and set assignee in the create fields, and also store
+            # the identifier for a post-creation assign_issue() call.
+            # Some Jira Server/DC configurations silently ignore the assignee
+            # field during creation, so the post-creation call acts as a safety
+            # net (similar to the epic two-step pattern).
+            assignee_identifier = None
             if assignee:
                 try:
-                    # _get_account_id now returns the correct identifier (accountId for cloud, name for server)
                     assignee_identifier = self._get_account_id(assignee)
                     self._add_assignee_to_fields(fields, assignee_identifier)
                 except ValueError as e:
-                    logger.warning(f"Could not assign issue: {str(e)}")
+                    logger.warning(f"Could not resolve assignee: {str(e)}")
 
             # Add components if provided
             if components:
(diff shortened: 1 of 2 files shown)


sooperset:
Merged via squash.

- Prepared head SHA: d4332d3a58cccd08cb9bb4258e26d00ad1147193
- Merge commit: e5bcdb65d32f92498ecd27d94321da93933ab8c3

Thanks @major!

sooperset:
Thank you @major for this practical Server/DC fix. We merged it after rebasing to current main.

PR #910, opened 2026-02-11
Feature (Jira) /development info api

imknysh:
## Description
Add support for retrieving development panel information (pull requests, branches, commits) linked to Jira issues. This uses the /rest/dev-status/1.0/issue/detail endpoint to fetch data from connected source control systems like Bitbucket Server, Bitbucket Cloud, GitHub, and GitLab

This feature enables users to query PRs and development activity associated with Jira issues, which is useful for tracking code changes, reviewing PR status, and understanding the development history of issues.

Fixes: N/A (new feature)

## Changes
Add DevelopmentMixin class in _development.py_  with methods:

- _get_issue_development_info()_ - retrieve dev info for a single issue
- _get_issues_development_info()_ - batch retrieve for multiple issues

Integrate DevelopmentMixin into _JiraFetcher_ class
Add two new MCP tools in jira.py:

- jira_get_issue_development_info - single issue
- jira_get_issues_development_info - batch for multiple issues

Update documentation in README.md and tools-reference.mdx

## Testing

- [x] Unit tests added/updated
- [ ] Integration tests passed
- [x] Manual checks performed: `[Tested against Jira Server (v8.x) with Bitbucket Server integration. Successfully retrieved PRs for multiple issues, verified correct parsing of PR metadata (name, status, author, source/destination branches, reviewers).]`

## Checklist

- [x] Code follows project style guidelines (linting passes).
- [x] Tests added/updated for changes.
- [x] All tests pass locally.
- [x] Documentation updated (if needed).

PR #928, opened 2026-02-18
fix(jira): add 404/403 handling for dev info API

sooperset:
## Summary

- Handles 404 from the dev-status endpoint gracefully: returns a descriptive message (`"Development info is not available — the Jira dev-status plugin may not be installed on this instance."`) instead of raising an exception. The dev-status plugin is not a standard Jira feature and is absent on many Server/DC instances.
- Handles 403 similarly: returns `"Access denied to development info — check Jira permissions or dev-status plugin configuration."` without raising.
- Updates the inline comment on `_session.get()` to document why the low-level session call is required (no higher-level wrapper exists for the `/rest/dev-status/1.0/` plugin endpoint).
- Adds a `TODO` for future result caching (up to 12 API calls per issue in auto-discovery mode).
- In auto-discovery mode, stops looping after the first 404/403 instead of exhausting all 12 combinations.

## Tests

Three new tests in `tests/unit/jira/test_development.py`:
- `test_get_issue_development_info_plugin_not_found` — 404 → error message, no exception
- `test_get_issue_development_info_access_denied` — 403 → error message, no exception
- `test_get_issue_development_info_auto_discovery_404` — auto-discovery exits after first 404

Closes follow-up from PR #910

### DIFF
--- a/src/mcp_atlassian/jira/development.py
+++ b/src/mcp_atlassian/jira/development.py
@@ -84,6 +84,12 @@ class DevelopmentMixin(JiraClient):
                         result = self._fetch_dev_info_for_app_type(
                             issue_key, issue_id, app_type, dt
                         )
+                        if "error" in result:
+                            # Plugin unavailable or access denied — capture the first
+                            # error and stop; all subsequent calls will fail the same way
+                            if not merged_result.get("error"):
+                                merged_result["error"] = result["error"]
+                            break
                         # Merge results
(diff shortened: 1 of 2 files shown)


sooperset:
Merged via squash.

- Prepared head SHA: 67d0e004ebcef14dc3163a38603a2bb214c72ffe
- Merge commit: 22540aa9c76853a9650e20fa614d6a2b98de2d17

Thanks @sooperset!

Issue #909, opened 2026-02-10
[Bug]: Orphaned Python/Docker processes after closing MCP clients/agents


rodrigoGA:
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/sooperset/mcp-atlassian/issues) to make sure this bug has not already been reported.
- [x] I have checked the [README](https://github.com/sooperset/mcp-atlassian/blob/main/README.md) for relevant information.

### Bug Description

When using `mcp-atlassian` via MCP stdio (launched by the client using `uvx` or `docker run -i --rm`), closing an agent/chat (or restarting MCP servers inside the client) leaves behind running Python processes and/or Docker containers. Repeating this creates multiple orphaned processes/containers.

### Steps to Reproduce

Configure MCP server in a client (Cursor / Claude Desktop / other) using stdio

### Expected Behavior

When the client closes the MCP connection or the agent is closed, the stdio server process/container should terminate. With docker run --rm, containers should be removed once the process exits.

### Actual Behavior

```shell
Old processes/containers remain running after agent close/disconnect and accumulate over time.
```

### mcp-atlassian Version

0.13.0

### Installation Method

From PyPI (pip install mcp-atlassian / uv add mcp-atlassian)

### Operating System

Windows

### Python Version

3.12.5

### Atlassian Instance Type

Jira Server / Data Center

### Client Application

codex

PR #931, opened 2026-02-18
fix(server): terminate stdio server on stdin EOF (#909)

sooperset:
## Summary
- Fixes orphaned `mcp-atlassian` stdio processes by adding an explicit stdin EOF guard around stdio server execution.
- When stdin closes, the guard now cancels the stdio server task and exits cleanly instead of leaving the process running.
- Keeps SSE and streamable-http behavior unchanged (no stdin monitoring), and updates tests to cover the stdio EOF shutdown path.

## Root Cause
The stdio path relied on `main_mcp.run_async()` to handle stdin closure internally, but EOF was not triggering server termination, leaving orphaned processes.

## Validation
- `uv run pytest -q`

### DIFF
--- a/src/mcp_atlassian/__init__.py
+++ b/src/mcp_atlassian/__init__.py
@@ -40,6 +40,41 @@ logging_stream = sys.stdout if is_env_truthy("MCP_LOGGING_STDOUT") else sys.stde
 logger = setup_logging(logging_level, logging_stream)
 
 
+async def _watch_stdin_eof() -> None:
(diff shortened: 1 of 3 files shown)


sooperset:
Merged via squash.

- Prepared head SHA: 7c1070c2d6574dc38e32b99df2cb9467c11af264
- Merge commit: f73d8d482c4fcf9e1169b63b558ed49894357c0e

Thanks @sooperset!

Issue #507, opened 2025-06-16
[Bug]:  Streamable-HTTP Transport ISSUE


WZHBEDZ:
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/sooperset/mcp-atlassian/issues) to make sure this bug has not already been reported.
- [x] I have checked the [README](https://github.com/sooperset/mcp-atlassian/blob/main/README.md) for relevant information.

### Bug Description

I setuped Streamable-HTTP Transport Example but it is not worjing well as SSE endpoint configuration

### Steps to Reproduce

# OR for streamable-http transport
docker run --rm -p 9000:9000 \
  --env-file /path/to/your/.env \
  ghcr.io/sooperset/mcp-atlassian:latest \
  --transport streamable-http --port 9000 -vv

{
  "mcpServers": {
    "mcp-atlassian-service": {
      "url": "http://localhost:9000/mcp"
    }
  }
}

### Expected Behavior

To work like SSE setup

### Actual Behavior

```shell
IT is asking for authentication headher giving 307 status code
```

### mcp-atlassian Version

latest

### Installation Method

Docker

### Operating System

macOS

### Python Version

it is docker setup

### Atlassian Instance Type

Jira Server / Data Center

### Client Application

github copilot

sooperset:
Hi @WZHBEDZ,

Thanks for reporting this! I've investigated this issue and found that the 307 redirect is not coming from mcp-atlassian itself, but likely from the underlying FastMCP framework or Docker networking layer.

Would you try adding authentication headers to your config and share the result?
   ```json
   {
     "mcpServers": {
       "mcp-atlassian-service": {
         "url": "http://localhost:9000/mcp",
         "headers": {
           "Authorization": "Token YOUR_PAT_TOKEN"
         }
       }
     }
   }
   ```

   ```bash
   curl -v -X POST \
     -H "Content-Type: application/json" \
     -d '{"jsonrpc":"2.0","method":"tools/list","params":{},"id":1}' \
     http://localhost:9000/mcp
   ```

WZHBEDZ:
Not working still 307, not sure this is docker issue...

<img width="984" alt="Image" src="https://github.com/user-attachments/assets/68f7dae4-ee7e-4970-b59b-0f30a98d3f2d" />

rupinder2:
[...]

PR #933, opened 2026-02-18
fix(server): fix streamable-HTTP transport configuration (#507)

sooperset:
## Summary
- Fixes the streamable-HTTP `/mcp` 307 redirect issue by normalizing the effective MCP path (trailing slash removed) before FastMCP route registration.
- Updates `UserTokenMiddleware` to check auth-processing eligibility against the active runtime streamable-HTTP path instead of deprecated static settings, which keeps path checks correct in Docker/custom-path HTTP mode.
- Adds regression tests for path normalization and runtime-path auth matching so streamable-HTTP works while SSE behavior remains unchanged.

## Root Cause
The streamable-HTTP route path could be registered with a trailing slash in some HTTP/Docker setups, causing Starlette to issue `307 Temporary Redirect` from `/mcp` to `/mcp/`. Some MCP clients do not handle this redirect reliably for transport negotiation. In parallel, middleware path checks used static settings instead of the active runtime path.

## Validation
- `uv run pytest -q`

### DIFF
--- a/src/mcp_atlassian/servers/main.py
+++ b/src/mcp_atlassian/servers/main.py
@@ -8,9 +8,11 @@ from typing import Any, Literal, Optional
 
 from cachetools import TTLCache
 from fastmcp import FastMCP
+from fastmcp import settings as fastmcp_settings
+from fastmcp.server.event_store import EventStore
+from fastmcp.server.http import StarletteWithLifespan
 from fastmcp.tools import Tool as FastMCPTool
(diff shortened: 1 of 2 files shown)


sooperset:
Merged via squash.

- Prepared head SHA: a79fc9cf854881e30922fc6af77ff9e29ad2ead2
- Merge commit: fb3c79eb3f202565b8400a7ee6f180506cfe6023

Thanks @sooperset!

PR #938, opened 2026-02-18
Add optional compatibility argument to jira_get_link_types

jwt625:
## Description

This PR makes `jira_get_link_types` resilient to inference gateways that silently drop empty-argument tool calls.

In some OpenAI-compatible/LiteLLM paths, no-arg tools can fail to execute reliably when the model emits `{}` arguments. A common workaround is to send a non-empty compatibility argument. Before this change, `jira_get_link_types` rejected that with a Pydantic `unexpected_keyword_argument` error.  
This PR adds an optional ignored compatibility argument so both empty and non-empty compatibility invocations succeed.

Fixes: added dummy input argument in `jira_get_link_types`.

## Changes

- Updated `jira_get_link_types` in `src/mcp_atlassian/servers/jira.py` to accept optional `dummy: str | None = None`.
- Documented `dummy` as a compatibility-only argument that is ignored by server behavior.
- Preserved existing behavior and output for standard calls with empty arguments.

## Testing

- N/A: Unit tests added/updated
- [x] Manual checks performed: `Validated MCP tools/call for jira_get_link_types with arguments {} and {"dummy":"x"}; both returned normal link-types payloads after restart.`

## Checklist

- N/A: Tests added/updated for changes.
- [x] All tests pass locally.
- N/A: Documentation updated (if needed).

### DIFF
--- a/src/mcp_atlassian/servers/jira.py
+++ b/src/mcp_atlassian/servers/jira.py
@@ -727,11 +727,24 @@ async def get_sprint_issues(
     tags={"jira", "read"},
     annotations={"title": "Get Link Types", "readOnlyHint": True},
 )
-async def get_link_types(ctx: Context) -> str:
+async def get_link_types(
+    ctx: Context,
+    dummy: Annotated[
+        str | None,
+        Field(
+            default=None,
+            description=(
+                "(Optional) Compatibility argument. Ignored by the server. "
+                "Use only when an inference gateway requires non-empty tool arguments."
+            ),
+        ),
+    ] = None,
+) -> str:
     """Get all available issue link types.
 
     Args:
         ctx: The FastMCP context.
+        dummy: Optional compatibility argument. Ignored.
 
     Returns:
         JSON string representing a list of issue link type objects.


src/mcp_atlassian/servers/jira.py
```diff
@@ -727,11 +727,24 @@ async def get_sprint_issues(
     tags={"jira", "read"},
     annotations={"title": "Get Link Types", "readOnlyHint": True},
 )
-async def get_link_types(ctx: Context) -> str:
+async def get_link_types(
+    ctx: Context,
+    dummy: Annotated[
```
sooperset, line 732: Adding a `dummy` parameter creates precedent concerns: if more no-arg tools appear, each would need the same workaround.

Alternatives to consider:
1. Fix the broken gateway upstream.
2. A server-level middleware that silently discards unknown/extra arguments globally.
3. If keeping per-tool: rename `dummy` to something less conspicuous (this shows up in the MCP tool schema that LLM clients see).

Also missing: unit tests covering the new parameter.

Could you share which gateway/setup triggers this?

sooperset (commented):
Thanks for the contribution! The problem is real — some gateways do choke on zero-argument tool calls. A few concerns with the per-tool approach — see inline comment.

sooperset:
Thank you for identifying this issue, @jwt625! The zero-argument tool problem with `jira_get_link_types` has been resolved in PR #959 with a different approach.

Instead of adding a `dummy` compatibility parameter, we added a semantically meaningful `name_filter` parameter that lets users filter link types by name substring (case-insensitive). This solves the zero-arg problem AND adds useful functionality.

PR #959 also includes a broader schema compatibility layer:
- Server-level `anyOf` flattening for Vertex AI / Google ADK compatibility
- 9 complex union parameters narrowed to string-based types
- 314 CI tests ensuring no tool has zero arguments (among other constraints)

Closing this PR since the fix has been merged. Thanks again for the contribution!

PR #686, opened 2025-09-08
feat(jira): add custom field options MCP tools

pibylick:
Implements feature request from #673

This PR adds three new MCP tools for retrieving Jira custom field options:

- **jira_get_customfield_options**: Get global options for a custom field
- **jira_get_customfield_contexts**: Get contexts for a custom field  
- **jira_get_customfield_context_options**: Get context-specific options

## Features
- Cross-platform support (Cloud and Server/DC with appropriate API versions)
- Comprehensive input validation and error handling
- Pagination support with configurable limits
- Detailed logging and debugging capabilities
- Full authentication method support (API tokens, PAT, OAuth)

## Implementation
- New Pydantic models for field options, contexts, and responses
- API client methods in FieldsMixin with Cloud/Server detection
- MCP tool implementations with comprehensive parameter documentation
- Unit tests covering models, validation, and API version selection
- Updated documentation with new tools in README

## Technical fixes
- Use from_api_response() methods in API client for proper response parsing
- Handle both Cloud ('values') and DC/Server ('options') API response formats
- Robust pagination handling with fallbacks for missing DC/Server fields
- Enhanced error resilience for malformed responses

This enables AI assistants to discover available custom field options before creating/updating Jira issues, improving automation accuracy.

Fixes #673

PR #945, opened 2026-02-20
feat(jira): optimize customfield options handling on top of latest main

pibylick:
## Description
This PR supersedes and is related to the closed PR #686.

It rebases the customfield options work on top of the latest `main` and adds performance-focused improvements for large option sets.

## What changed
- aligned branch changes with current `main`
- resolved method collision risk in customfield handlers
- added compact JSON responses for customfield tools
- added `values_only` mode for lighter payloads
- added server-side filtering (`contains`, `return_limit`) to reduce oversized tool outputs
- updated and extended tests for payload/format handling

## Why
Some Jira custom fields can have thousands of options, which can exceed MCP tool output limits. These changes reduce payload size and improve practical usability with LLM workflows.

Related to #686

## Real-world case that motivated this optimization
In our Jira instance, a custom field has around 6500 option values.
Returning all options in one tool response can exceed MCP context/output limits.

To address this, this PR adds filtering support so clients/LLMs can request targeted subsets instead of downloading the entire option list in one response.


src/mcp_atlassian/jira/fields.py
```diff
@@ -697,3 +703,228 @@ def similarity(keyword: str, field: dict) -> int:
         except Exception as e:
             logger.error(f"Error searching fields: {str(e)}")
             return []
+
+    def get_customfield_contexts(
```
sooperset, line 707: This duplicates `FieldOptionsMixin.get_field_contexts()` in `jira/field_options.py` (from PR #944, already on main). The existing implementation includes Cloud/Server dispatch and pagination.

Field option methods belong in `FieldOptionsMixin` (the dedicated mixin), not `FieldsMixin`.

src/mcp_atlassian/models/jira/field_option.py
```diff
@@ -103,3 +103,300 @@ def to_simplified_dict(self) -> dict[str, Any]:
                 child.to_simplified_dict() for child in self.child_options
             ]
         return result
+
+
+class JiraFieldOption(BaseModel):
```
sooperset, line 108: This duplicates `FieldOption` and `FieldContext` models already on main (from PR #944). The existing models extend `ApiModel` per project convention — these use `BaseModel` which breaks that convention.

pibylick:
Closing this in favor of #974. We moved the valuable parts discussed here (contains filtering, return_limit, values_only, and compact JSON output) into #974, implemented on top of the already-merged #944  path to avoid duplicating tool/model flows.

Issue #946, opened 2026-02-20
[Bug]: Issue When use Download Attachments tool


Ibrahimogod:
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/kdtix-open/mcp-atlassian/issues) to make sure this bug has not already been reported.
- [x] I have checked the [README](https://github.com/kdtix-open/mcp-atlassian/blob/main/README.md) for relevant information.

### Bug Description

When I call `Download Attachments` it fails with logs

```
2026-02-20 16:25:22.621 [warning] [server stderr] [02/20/26 14:25:22] Error validating tool 'download_attachments': 1 validation
2026-02-20 16:25:22.621 [warning] [server stderr]                     error for call[download_attachments]
2026-02-20 16:25:22.621 [warning] [server stderr]                     target_dir
2026-02-20 16:25:22.621 [warning] [server stderr]                       Unexpected keyword argument
2026-02-20 16:25:22.621 [warning] [server stderr]                     [type=unexpected_keyword_argument,
2026-02-20 16:25:22.621 [warning] [server stderr]                     input_value='d:\\Users\\ibadr\\Projects\\FocalAgent',
2026-02-20 16:25:22.621 [warning] [server stderr]                     input_type=str]
2026-02-20 16:25:22.622 [warning] [server stderr]                         For further information visit
2026-02-20 16:25:22.622 [warning] [server stderr]                     https://errors.pydantic.dev/2.12/v/unexpected_keyword_argume
2026-02-20 16:25:22.622 [warning] [server stderr]                     nt
2026-02-20 16:25:22.622 [warning] [server stderr]                     ╭─────────── Traceback (most recent call last) ────────────╮
2026-02-20 16:25:22.622 [warning] [server stderr]                     │ /app/.venv/lib/python3.10/site-packages/fastmcp/tools/to │
2026-02-20 16:25:22.622 [warning] [server stderr]                     │ ol_manager.py:160 in call_tool                           │
2026-02-20 16:25:22.622 [warning] [server stderr]                     │                                                          │
2026-02-20 16:25:22.622 [warning] [server stderr]                     │ /app/.venv/lib/python3.10/site-packages/fastmcp/tools/to │
2026-02-20 16:25:22.622 [warning] [server stderr]                     │ ol.py:381 in run                                         │
2026-02-20 16:25:22.622 [warning] [server stderr]                     │                                                          │
[...]

PR #947, opened 2026-02-20
fix(jira): add optional target_dir to download_attachments tool

Ibrahimogod:
## Description

The `jira_download_attachments` MCP tool only accepted `issue_key` as input. Passing `target_dir` caused a Pydantic validation error:

```
1 validation error for call[download_attachments]
target_dir
  Unexpected keyword argument [type=unexpected_keyword_argument, ...]
```

This PR adds an optional `target_dir` parameter that lets callers request attachments be saved directly to a local directory on disk. The existing filesystem-free (base64 embedded resource) behaviour is fully preserved when `target_dir` is omitted.

## Changes

- Add optional `target_dir: str | None = None` parameter to `download_attachments` in `src/mcp_atlassian/servers/jira.py`
- When `target_dir` is provided, delegate to `jira.download_issue_attachments()` and return a JSON `TextContent` summary of saved files
- When `target_dir` is omitted, preserve existing behaviour: fetch into memory via `jira.get_issue_attachment_contents()` and return `EmbeddedResource` items
- Update docstring to document both code paths

## Testing

- [x] Manual checks performed: called tool with `target_dir` argument — no longer raises validation error; files are saved to the specified directory and a JSON summary is returned

### New unit tests (`tests/unit/servers/test_jira_server.py`)

| Test | Verifies |
|---|---|
| `test_download_attachments_no_target_dir_returns_embedded_resources` | Without `target_dir`: `get_issue_attachment_contents` is called; response contains a text summary + one `EmbeddedResource` per attachment |
| `test_download_attachments_with_target_dir_saves_to_disk` | With `target_dir`: `download_issue_attachments` is called with the correct path; `get_issue_attachment_contents` is NOT called; response contains a JSON text summary |

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/servers/jira.py
+++ b/src/mcp_atlassian/servers/jira.py
@@ -522,22 +522,48 @@ async def download_attachments(
             pattern=ISSUE_KEY_PATTERN,
         ),
     ],
+    target_dir: Annotated[
+        str | None,
+        Field(
+            description=(
+                "Optional local directory path to save attachment files to disk. "
+                "When provided, files are written to this directory and a summary "
+                "is returned. When omitted, attachment contents are returned as "
+                "base64-encoded embedded resources."
+            ),
+        ),
+    ] = None,
 ) -> list[TextContent | EmbeddedResource]:
     """Download attachments from a Jira issue.
 
-    Returns attachment contents as base64-encoded embedded resources so that
-    they are available over the MCP protocol without requiring filesystem
-    access on the server.
+    When target_dir is provided, attachments are saved to that directory on
+    disk and a JSON summary of the operation is returned.  When target_dir is
+    omitted, attachment contents are returned as base64-encoded embedded
+    resources so that they are available over the MCP protocol without
+    requiring filesystem access on the server.
 
     Args:
         ctx: The FastMCP context.
         issue_key: Jira issue key.
+        target_dir: Optional local directory to save attachments to disk.
 
     Returns:
-        A list containing a text summary and one EmbeddedResource per
-        successfully downloaded attachment.
+        A list containing a text summary and, when target_dir is not provided,
+        one EmbeddedResource per successfully downloaded attachment.
     """
     jira = await get_jira_fetcher(ctx)
+
+    if target_dir is not None:
+        result = jira.download_issue_attachments(
+            issue_key=issue_key, target_dir=target_dir
+        )
+        return [
+            TextContent(
+                type="text",
+                text=json.dumps(result, indent=2, ensure_ascii=False),
+            )
+        ]
+
     result = jira.get_issue_attachment_contents(issue_key=issue_key)
 
     contents: list[TextContent | EmbeddedResource] = []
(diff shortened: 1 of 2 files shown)


src/mcp_atlassian/servers/jira.py
```diff
@@ -522,22 +522,48 @@ async def download_attachments(
             pattern=ISSUE_KEY_PATTERN,
         ),
     ],
+    target_dir: Annotated[
```
sooperset, line 525: This violates the project's attachment contract: MCP tool-layer functions must return `EmbeddedResource` with base64 content — no filesystem writes from the tool layer. The rationale is Docker compatibility and stateless operation.

The mixin layer (`jira/attachments.py`) already supports filesystem downloads via `download_issue_attachments()` for direct library consumers, which is the right place for this.

src/mcp_atlassian/servers/jira.py
```diff
@@ -522,22 +522,48 @@ async def download_attachments(
...
 
     Returns:
-        A list containing a text summary and one EmbeddedResource per
-        successfully downloaded attachment.
+        A list containing a text summary and, when target_dir is not provided,
+        one EmbeddedResource per successfully downloaded attachment.
     """
     jira = await get_jira_fetcher(ctx)
+
+    if target_dir is not None:
```
sooperset, line 556: **Path traversal risk:** `target_dir` is passed directly from the MCP tool layer without sanitization. An LLM client could pass `target_dir="/etc"` or `target_dir="../../sensitive"`.

Also, looking at issue #946 — the root cause appears to be the LLM hallucinating a `target_dir` parameter that was never part of the tool schema. The server correctly rejected the unexpected keyword. This is expected behavior, not a bug.

sooperset (changes requested):
Thanks for the quick fix! Unfortunately this conflicts with the project's attachment contract — MCP tool-layer functions must return EmbeddedResource only, no filesystem writes. See inline comments for details.

src/mcp_atlassian/servers/jira.py
sooperset: `target_dir` flows directly to `Path(target_dir).mkdir(parents=True, exist_ok=True)` with no path containment checks. Any path is accepted — in Docker with bind mounts this could write to host filesystem locations. The tool is also tagged `{"jira", "read"}` with `readOnlyHint: True`, so it bypasses `READ_ONLY_MODE` while now performing writes.

If disk-writing from MCP is genuinely needed, it might work better as a separate `"write"`-tagged tool with path sanitization (allowlisted directories, `realpath()` checks).

sooperset:
Closing — this approach violates the embedded-first contract that MCP tools follow. Attachments are returned as base64-encoded `EmbeddedResource` objects over the protocol, with no filesystem dependency on the server. Adding `target_dir` would couple the server to a local filesystem, which breaks containerized and remote deployments. The existing tool (`jira_download_attachments`) already provides full attachment content inline.

PR #949, opened 2026-02-20
fix: remove credential/token logging and harden security

JakubAnderwald:
## Description

Remove all credential/token logging patterns across the codebase and harden security in SSL handling, file downloads, and JQL construction. Tokens, secrets, and credential material (full or partial) should never be written to logs at any level.

## Changes

- **oauth.py**: Remove token exchange payload logging (contained `client_secret`), remove response body logging (contained `access_token`/`refresh_token`), replace partial token logging with safe confirmation messages, remove unused `pprint` import
- **oauth_setup.py**: Redact `client_secret` in environment variable output, replace partial access/refresh token logging with non-sensitive messages
- **dependencies.py**: Replace partial user token logging (`token ...{last 8 chars}`) with `token ...<redacted>` in both Jira and Confluence fetcher creation
- **ssl.py**: Remove `SSL_OP_LEGACY_SERVER_CONNECT` and `OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION` flags that unnecessarily weaken the TLS handshake; update docstring
- **attachments.py**: Add path traversal guard to `download_attachment()` ensuring resolved path stays within current working directory
- **search.py**: Sanitize JQL project filter interpolation by escaping double-quote characters in project names before interpolation

## Testing

- [x] Manual checks performed: comprehensive grep sweep confirming zero credential logging remains (`token[:`, `access_token[`, `refresh_token[`, `client_secret` in log statements, `response.text` in debug logs, `user_token[`, `pformat.*payload` — all clean)

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/attachments.py
+++ b/src/mcp_atlassian/jira/attachments.py
@@ -38,13 +38,11 @@ class AttachmentsMixin(JiraClient, AttachmentsOperationsProto):
                 target_path = os.path.abspath(target_path)
 
             # Guard against path traversal
-            base_dir = Path(os.getcwd()).resolve()
-            resolved = Path(target_path).resolve()
-            if not resolved.is_relative_to(base_dir):
+            base_dir = os.getcwd()
+            if not Path(target_path).is_relative_to(base_dir):
                 raise ValueError(
-                    f"Path traversal detected: {resolved} is outside {base_dir}"
+                    f"Path traversal detected: {target_path} is outside {base_dir}"
                 )
-            target_path = str(resolved)
 
             logger.info(f"Downloading attachment from {url} to {target_path}")
 
@@ -217,13 +215,11 @@ class AttachmentsMixin(JiraClient, AttachmentsOperationsProto):
             target_dir = os.path.abspath(target_dir)
 
         # Guard against path traversal
-        base_dir = Path(os.getcwd()).resolve()
-        resolved_dir = Path(target_dir).resolve()
-        if not resolved_dir.is_relative_to(base_dir):
+        base_dir = os.getcwd()
+        if not Path(target_dir).is_relative_to(base_dir):
             raise ValueError(
-                f"Path traversal detected: {resolved_dir} is outside {base_dir}"
+                f"Path traversal detected: {target_dir} is outside {base_dir}"
             )
-        target_dir = str(resolved_dir)
 
         logger.info(
             f"Downloading attachments for {issue_key} to directory: {target_dir}"
--- a/src/mcp_atlassian/jira/search.py
+++ b/src/mcp_atlassian/jira/search.py
@@ -60,7 +60,9 @@ class SearchMixin(JiraClient, IssueOperationsProto):
                 # Build the project filter query part
                 # Sanitize project names to prevent JQL injection
                 # Escape backslashes before double-quotes to prevent bypass
-                projects = [p.replace('\\', '\\\\').replace('"', '\\"') for p in projects]
+                projects = [
+                    p.replace("\\", "\\\\").replace('"', '\\"') for p in projects
+                ]
 
                 if len(projects) == 1:
                     project_query = f'project = "{projects[0]}"'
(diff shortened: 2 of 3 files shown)


src/mcp_atlassian/jira/search.py
```diff
@@ -58,6 +58,9 @@ def search_issues(
                 projects = [p.strip() for p in filter_to_use.split(",")]
 
                 # Build the project filter query part
+                # Sanitize project names to prevent JQL injection
+                projects = [p.replace('"', '\\"') for p in projects]
```
sooperset: **Blocker: JQL injection bypassable via backslash.**

Escaping `"` without first escaping `\` allows bypass. A project name ending with `\` produces `\"` which in JQL is an escaped quote — the closing quote is consumed and injection escapes the string.

Need to escape `\` before `"`:

```python
projects = [p.replace('\\', '\\\\').replace('"', '\\"') for p in projects]
```

src/mcp_atlassian/jira/attachments.py
```diff
@@ -37,6 +37,13 @@ def download_attachment(self, url: str, target_path: str) -> bool:
             if not os.path.isabs(target_path):
                 target_path = os.path.abspath(target_path)
 
+            # Guard against path traversal
+            base_dir = os.path.abspath(os.getcwd())
+            if not target_path.startswith(base_dir):
```
sooperset: String `startswith()` on resolved paths is a known footgun — `/tmp/attack` starts with `/tmp/att` if cwd happens to be `/tmp/att`.

Consider using `Path.resolve().is_relative_to()` instead:

```python
base_dir = Path(os.getcwd()).resolve()
resolved = Path(target_path).resolve()
if not resolved.is_relative_to(base_dir):
    raise ValueError(...)
```

Also, `download_issue_attachments` has no equivalent guard on its `target_dir`.

src/mcp_atlassian/utils/oauth_setup.py
```diff
@@ -293,7 +289,7 @@ def run_oauth_flow(args: OAuthSetupArgs) -> bool:
             )
             logger.info("------------------------------------------------------------")
             logger.info(f"ATLASSIAN_OAUTH_CLIENT_ID={oauth_config.client_id}")
-            logger.info(f"ATLASSIAN_OAUTH_CLIENT_SECRET={oauth_config.client_secret}")
+            logger.info("ATLASSIAN_OAUTH_CLIENT_SECRET=<redacted>")
```
sooperset, line 292: This env-var line was correctly redacted, but `client_secret` is still logged in the VS Code config JSON blob further down in this file (around line 345 on main) via `logger.info(vscode_json)`. That needs redaction too.

sooperset (changes requested):
Thanks for the security hardening work! The credential logging removal and TLS cleanup are well-targeted. Three items need attention — see inline comments.

src/mcp_atlassian/jira/search.py
JakubAnderwald: Fixed — now escaping backslashes before double-quotes:
```python
projects = [p.replace("\\", "\\\\").replace("\"", "\\\"") for p in projects]
```
This prevents the `\` + `\"` → `\\"` bypass.

src/mcp_atlassian/jira/attachments.py
JakubAnderwald: Both addressed:

1. Replaced `str.startswith()` with `Path.is_relative_to()` in `download_attachment()` — avoids the false-positive prefix match issue.
2. Added the same `is_relative_to()` guard to `download_issue_attachments()` on `target_dir`, upfront before `mkdir(parents=True)`, so callers get a clean `ValueError` instead of confusing per-download failures.

src/mcp_atlassian/utils/oauth_setup.py
JakubAnderwald: Good catch — replaced `oauth_config.client_secret` with `"<redacted - set via environment variable>"` in the VS Code config dict. The Docker args already pass it via `-e ATLASSIAN_OAUTH_CLIENT_SECRET`, so the logged JSON template does not need the actual value.

JakubAnderwald:
Is the history of this PR basically one coding agent talking to another agent? :)

sooperset:
Haha, you caught us\! 😄 We've been getting so many great community PRs and issues that we started using AI agents to help keep up with reviews. The review quality is still validated by humans though — your security fixes were genuinely solid work. Thanks again for the contribution\!

PR #973, opened 2026-02-23
fix(jira): accept JSON string for update_issue fields parameter

hteichmann-strato:
## Summary

- `d57b7fd` converted all dict-typed tool parameters to `str` for AI platform schema compatibility, but the `fields` parameter in `update_issue()` was missed
- LLMs passing `fields` as a JSON string receive a Pydantic validation error before the function body is reached: `Input should be a valid dictionary [type=dict_type, input_value='{"description": "..."}', input_type=str]`

## Fix

- Change `fields` type annotation from `dict[str, Any]` to `dict[str, Any] | str`
- Route through the existing `_parse_additional_fields()` helper, consistent with how `additional_fields` is already handled in the same function

## Test plan

- [ ] Existing `update_issue` unit tests pass (`pytest tests/unit/servers/test_jira_server.py -k update`)
- [ ] Manually verified: passing `fields` as a JSON string no longer raises a validation error

🤖 Co-authored with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/mcp_atlassian/servers/jira.py
+++ b/src/mcp_atlassian/servers/jira.py
@@ -1166,7 +1166,7 @@ async def update_issue(
         ),
     ],
     fields: Annotated[
-        dict[str, Any],
+        str,
         Field(
             description=(
                 "Dictionary of fields to update. For 'assignee', provide a string identifier (email, name, or accountId). "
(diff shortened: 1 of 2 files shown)


src/mcp_atlassian/servers/jira.py
```diff
@@ -1166,7 +1166,7 @@ async def update_issue(
         ),
     ],
     fields: Annotated[
-        dict[str, Any],
+        str,
         Field(
             description=(
                 "Dictionary of fields to update. For 'assignee', provide a string identifier (email, name, or accountId). "
```
sooperset, line 1172: (minor) The description still says "Dictionary of fields to update" with Python single-quote syntax, but the type is now `str`. An LLM following this could send `{'key': 'val'}` which `json.loads` rejects. I'll update this to match the `additional_fields` pattern in a follow-up.

PR #974, opened 2026-02-23
feat(jira): extract valuable field-options optimizations from #945 on top of #944

pibylick:
## Summary
This PR takes the valuable parts called out in #945 and adds them on top of the already-merged #944 `jira_get_field_options` path, without reintroducing duplicate tool/model implementations.

## What was added above #944 functionality
- `contains` parameter for case-insensitive server-side filtering
- `return_limit` parameter to cap returned options after filtering
- `values_only` mode to return only option values for smaller payloads
- Compact JSON serialization for lighter MCP responses

## Important compatibility note
- Default response shape remains backward-compatible with #944 (list of option objects).
- New behavior is opt-in via the added parameters.

## Tests
- Added server tests for:
  - default response shape
  - `contains` filtering
  - `return_limit` limiting
  - `values_only` output
- Ran field-options unit tests and pre-commit checks on changed files.

### DIFF
--- a/src/mcp_atlassian/servers/jira.py
+++ b/src/mcp_atlassian/servers/jira.py
@@ -73,6 +73,16 @@ def _to_values_only_payload(
     return values
 
 
+def _matches_contains(option: dict[str, Any], needle: str) -> bool:
+    """Return True if option or any of its child options match the needle."""
+    if needle in str(option.get("value", "")).casefold():
+        return True
+    for child in option.get("child_options", []):
+        if isinstance(child, dict) and needle in str(child.get("value", "")).casefold():
+            return True
+    return False
+
+
 def _apply_option_filters(
     options: list[dict[str, Any]],
     contains: str | None = None,
@@ -90,14 +100,13 @@ def _apply_option_filters(
     """
     filtered_options = options
 
-    # Apply substring filter
+    # Apply substring filter (matches parent or child values)
     if contains:
         needle = contains.casefold()
         filtered_options = [
             option
             for option in filtered_options
-            if isinstance(option, dict)
-            and needle in str(option.get("value", "")).casefold()
+            if isinstance(option, dict) and _matches_contains(option, needle)
         ]
 
     # Apply result limit
--- a/tests/unit/servers/test_jira_server.py
+++ b/tests/unit/servers/test_jira_server.py
@@ -596,6 +596,50 @@ async def test_get_field_options_contains_filter(jira_client):
 
 
 @pytest.mark.anyio
+async def test_get_field_options_contains_filter_cascading(
+    jira_client, mock_jira_fetcher
+):
+    """Ensure contains matches child values on cascading select fields."""
+
+    def mock_cascading_options(
+        field_id, context_id=None, project_key=None, issue_type=None
+    ):
+        raw = [
+            {
+                "id": "1",
+                "value": "US",
+                "child_options": [
+                    {"id": "11", "value": "California"},
+                    {"id": "12", "value": "Texas"},
+                ],
+            },
+            {
+                "id": "2",
+                "value": "Canada",
+                "child_options": [{"id": "21", "value": "Ontario"}],
+            },
+        ]
+        options = []
+        for payload in raw:
+            option = MagicMock()
+            option.to_simplified_dict.return_value = payload
+            options.append(option)
+        return options
+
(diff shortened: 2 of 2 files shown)


src/mcp_atlassian/servers/jira.py
```diff
@@ -440,7 +537,16 @@ async def get_field_options(
...
+
+    # Apply server-side filtering
+    filtered_result = _apply_option_filters(
+        result, contains=contains, return_limit=return_limit
+    )
+
+    # Keep backward-compatible output shape by default
+    if values_only:
+        return _json_dumps_compact(_to_values_only_payload(filtered_result))
+    return _json_dumps_compact(filtered_result)
```
sooperset: This changes the default output from `json.dumps(result, indent=2)` to compact JSON, even when no new parameters are used. Every other tool in the file uses `indent=2`. Since the PR description says "Default response shape remains backward-compatible with #944", was this change intentional? Consider keeping `indent=2` for the non-`values_only` path.

src/mcp_atlassian/servers/jira.py
```diff
@@ -31,6 +31,70 @@
...
+    """Convert option objects to a compact values-only list.
+
+    Args:
+        options: List of option dictionaries with 'value' keys.
+
+    Returns:
+        List of option values only.
+    """
+    values: list[str] = []
+    for option in options:
```
sooperset, line 62: (question) For cascading select fields, `to_simplified_dict()` emits `child_options`. This function only reads the top-level `value`, so `values_only` mode would return `["US"]` and silently drop child values like `"California"`. Same applies to the `contains` filter. Worth either handling recursion or documenting this limitation in the parameter descriptions.

src/mcp_atlassian/servers/jira.py
```diff
@@ -31,6 +31,82 @@
...
+    filtered_options = options
+
+    # Apply substring filter
+    if contains:
+        needle = contains.casefold()
+        filtered_options = [
+            option
+            for option in filtered_options
+            if isinstance(option, dict)
+            and needle in str(option.get("value", "")).casefold()
```
sooperset: Follow-up on the earlier cascading select thread: `values_only` recursion was fixed, but `contains` still only checks the parent `option.get("value")`. A search like `contains="California"` on a cascading field with parent `"US"` returns nothing.

Consider matching against child values too:
```python
def _matches_contains(option: dict[str, Any], needle: str) -> bool:
    if needle in str(option.get("value", "")).casefold():
        return True
    for child in option.get("child_options", []):
        if isinstance(child, dict) and needle in str(child.get("value", "")).casefold():
            return True
    return False
```

Alternatively, documenting that `contains` only filters top-level values would also be fine.

tests/unit/servers/test_jira_server.py
```diff
@@ -524,6 +553,143 @@ async def test_search(jira_client, mock_jira_fetcher):
...
+
+    content = json.loads(response.content[0].text)
+    assert isinstance(content, list)
+    assert content == ["Frontend", "Product Ops"]
+
+
+@pytest.mark.anyio
+async def test_get_field_options_values_only_cascading(jira_client, mock_jira_fetcher):
+    """Ensure values_only preserves parent-child structure for cascading selects."""
+    # Mock cascading select options
```
sooperset, line 682: (minor) `ruff-format` fails here — needs a blank line between the docstring and the `# Mock cascading select options` comment. Pre-commit auto-fixes it.

pibylick:
in my opinion ready to be merged

src/mcp_atlassian/servers/jira.py
```diff
@@ -410,6 +495,37 @@ async def get_field_options(
...
+            description=(
+                "Optional cap on number of returned options after filtering (1-10000)."
+            ),
+            default=None,
+            ge=1,
+            le=10000,
+        ),
+    ] = None,
+    values_only: Annotated[
+        bool | None,
```
sooperset, line 519: (minor) `values_only` is `bool | None` with `default=None`, but the code treats `None` and `False` identically (`if values_only:`). Every other boolean param in this file uses `bool` with a concrete default — `bool = False` would be simpler here.

sooperset:
Thank you for the contribution, @pibylick! This has been reimplemented in #1074 with updated tests and project conventions. Your work is credited via Co-authored-by trailer.

PR #976, opened 2026-02-23
feat(auth): use OS native trust store for SSL verification via truststore

### LINKED ISSUE #975: [Feature]: Use OS native trust store for SSL verification (enterprise CA support)
### Problem

Enterprise environments commonly use internal Certificate Authorities whose root certificates are installed in the operating system's trust store (e.g., Windows Certificate Store, macOS Keychain). When connecting to on-premise Jira Server/Data Center or Confluence instances that use certificates signed by these enterprise CAs, mcp-atlassian fails with:

```
SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1017)'))
```

This forces users to disable SSL verification entirely with `--no-jira-ssl-verify` (insecure)

### Root Cause

Python's `requests` library does **not** use the operating system's trust store. Instead, it bundles its own CA certificates via the `certifi` package. This means enterprise CA certificates installed system-wide are completely invisible to mcp-atlassian, regardless of the OS configuration.

Additionally, `REQUESTS_CA_BUNDLE` environment variable support has two problems:
1. The `atlassian-python-api` library passes `trust_env=False` to the `requests.Session` for PAT and OAuth authentication methods, which disables reading `REQUESTS_CA_BUNDLE` entirely
2. The `ssl_verify` configuration only accepts boolean values, not file paths — so even if `trust_env` were enabled, there's no way to pass a CA bundle path through the config

### Proposed Solution

Use the [`truststore`](https://truststore.readthedocs.io/) library to make all SSL connections use the OS native trust store. This is a single call at startup:

```python
import truststore
truststore.inject_into_ssl()
```

Key requirements:
- Must be called **before** `requests.adapters` is imported (that module creates a preloaded `SSLContext` at import time using certifi)
- Should be enabled by default (opt-out via `MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE=false`)
- `truststore` supports Windows (CryptoAPI), macOS (Security framework), and Linux (OpenSSL system certs)

### Impact

This affects all users connecting to on-premise Atlassian instances behind enterprise CAs — a very common enterprise deployment scenario. Related issues that are symptoms of this root cause:
- #148 (SSL issue when running via uvx)
- #135 (SSL verify not fully supported)
- #327 (Docker cannot bypass SSL verification)

These were all worked around by disabling verification, but the proper fix is to trust the certificates the OS already trusts.

### Additional Context

- `truststore` requires Python 3.10+ (already the minimum for mcp-atlassian)
- The [`truststore` library](https://pypi.org/project/truststore/) is maintained by a CPython core developer (Seth Larson) and is already integrated into pip
- This also benefits Docker deployments where custom CA certificates are added to the container's system trust store

tsondergaard:
## Description

This enables enterprise environments with internal CAs to connect to on-premise Jira/Confluence instances without disabling SSL verification.

Opt out by setting MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE=false.

Fixes #975

## Changes

Inject truststore at startup so all SSL connections verify certificates against the OS trust store (Windows Certificate Store, macOS Keychain, Linux system CAs) instead of the bundled certifi CA bundle.

## Testing

- [x] Manual checks performed: `Tested with uvx from Visual Studio Code against JIRA instance using locally trusted CA certificate`

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -99,8 +99,11 @@ CONFLUENCE_API_TOKEN=your_confluence_api_token_here
 # =============================================
 # Only applicable if your JIRA_URL/CONFLUENCE_URL points to a Server/DC instance (not *.atlassian.net).
 # Default is true. Set to false if using self-signed certificates (not recommended for production environments).
+# Note: Enterprise CA certificates in the OS trust store are trusted automatically via truststore.
 #JIRA_SSL_VERIFY=true
 #CONFLUENCE_SSL_VERIFY=true
+# Set to false to disable OS trust store integration and use the bundled certifi CA bundle instead.
+#MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE=true
 
 # --- CLIENT CERTIFICATE AUTHENTICATION (mTLS) ---
 # For mutual TLS authentication with Server/DC instances.
--- a/docs/authentication.mdx
+++ b/docs/authentication.mdx
@@ -60,7 +60,8 @@ CONFLUENCE_PERSONAL_TOKEN=your_personal_access_token
 ```
 
 <Note>
-For self-signed certificates, set `JIRA_SSL_VERIFY=false` and/or `CONFLUENCE_SSL_VERIFY=false`.
+Enterprise CA certificates in the OS trust store (Windows Certificate Store, macOS Keychain, Linux system CAs) are trusted automatically.
+For self-signed certificates not in the OS trust store, set `JIRA_SSL_VERIFY=false` and/or `CONFLUENCE_SSL_VERIFY=false`.
 </Note>
 
 ## OAuth 2.0 (Cloud) - Advanced
--- a/docs/configuration.mdx
+++ b/docs/configuration.mdx
@@ -129,6 +129,7 @@ This guide covers IDE integration, environment variables, and advanced configura
 | `CONFLUENCE_API_TOKEN` | Confluence API token (Cloud) |
 | `CONFLUENCE_PERSONAL_TOKEN` | Confluence Personal Access Token (Server/DC) |
 | `CONFLUENCE_SSL_VERIFY` | SSL verification (`true`/`false`) |
+| `MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE` | Use OS native trust store (`true`/`false`, default: `true`) |
 
 ### Filtering Options
 
--- a/docs/troubleshooting.mdx
+++ b/docs/troubleshooting.mdx
@@ -20,11 +20,25 @@ description: "Common issues and debugging tips for MCP Atlassian"
   </Accordion>
 
   <Accordion title="SSL Certificate Issues">
-    For Server/Data Center with self-signed certificates:
+    mcp-atlassian uses the OS native trust store by default (via
+    [truststore](https://truststore.readthedocs.io/)), so certificates
+    signed by enterprise CAs installed in Windows Certificate Store,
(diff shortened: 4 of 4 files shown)


src/mcp_atlassian/__init__.py
```diff
@@ -5,6 +5,20 @@
 import threading
 from importlib.metadata import PackageNotFoundError, version
 
+# Inject truststore BEFORE any requests/urllib3 imports to ensure the
+# OS-native trust store (e.g. Windows Certificate Store) is used for
+# SSL verification instead of the bundled certifi CA certificates.
+# This must happen before 'requests.adapters' is imported because that
+# module creates a preloaded SSLContext at import time.
+if os.getenv("MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE", "true").lower() not in (
```
sooperset: `os.getenv()` runs at import time, but `load_dotenv()` isn't called until `main()`. Setting `MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE=false` in `.env` — the primary config path — has no effect. Only shell-exported vars work.

Consider a minimal `load_dotenv()` before this check, or reading just this variable from `.env` directly.

src/mcp_atlassian/__init__.py
```diff
@@ -5,6 +5,20 @@
...
+    "false",
+    "0",
+    "no",
+):
+    import truststore
+
+    truststore.inject_into_ssl()
```
sooperset: If `inject_into_ssl()` fails (unusual platform, sandboxed env), the app crashes at import time with no fallback. Consider wrapping in try/except:

```python
try:
    import truststore
    truststore.inject_into_ssl()
except Exception:
    import logging
    logging.getLogger("mcp-atlassian").warning(
        "Failed to inject OS trust store; falling back to bundled certificates",
        exc_info=True,
    )
```

tsondergaard: I had that, but removed it with the assumption that these cases would be addressed with `MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE=0`. I will add it back.

bbartels:
We would love to deploy the mcp interally, this is blocking us currently! Any chance you'll get around to addressing the PR feedback? @tsondergaard Happy to do so myself as well if you are busy

src/mcp_atlassian/__init__.py
tsondergaard: Done.

src/mcp_atlassian/__init__.py
tsondergaard: Done

tsondergaard:
Hi @bbartels,

> We would love to deploy the mcp interally, this is blocking us currently! Any chance you'll get around to addressing the PR feedback? @tsondergaard Happy to do so myself as well if you are busy

I've addressed the feedback. I don't use a dotenv file myself in the setup I use, so if you could test that it would be great. Your review comments are also welcome.

bbartels (approved):
Looks all good from my side! @sooperset what do you think?

sooperset:
Thank you for this contribution, @tsondergaard! The truststore integration is clean and addresses a real pain point for enterprise users. Both of the earlier review concerns have been properly addressed.

Merged! We'll follow up with a small hardening fix for an edge case in the env var parsing and add some basic unit tests.

PR #985, opened 2026-02-24
fix: sync OAuth tokens to base keyring username for load_tokens() compatibility

### LINKED ISSUE #984: OAuth tokens not synced between keyring and file storage after re-authentication
## Description

After re-authenticating via `mcporter auth mcp-atlassian --reset`, the OAuth session configuration fails with:

```
Error calling tool 'search': Failed to configure OAuth session
```

## Root Cause

The token loading logic in `OAuthConfig.load_tokens()` prioritizes keyring over file storage:

1. User re-authenticates → new tokens saved to **both** keyring and file
2. However, the keyring entry for the cloud-specific username (`oauth-{client_id}-cloud-{cloud_id}`) may not be updated correctly
3. On next MCP call, `load_tokens()` retrieves **stale tokens from keyring**
4. `ensure_valid_token()` tries to refresh but fails (403 Forbidden - refresh token already rotated)
5. `configure_oauth_session()` returns False

## Evidence

```python
# Keyring has old expires_at
>>> keyring.get_password("mcp-atlassian-oauth", "oauth-{client_id}")
{"expires_at": 1771648288, ...}  # Expired

# File has fresh tokens
>>> json.load(open("~/.mcp-atlassian/oauth-{client_id}.json"))
{"expires_at": 1771934383, ...}  # Valid
```

## Workaround

Manually sync file tokens to keyring:

```python
import keyring, json
tokens = json.load(open("~/.mcp-atlassian/oauth-{client_id}.json"))
keyring.set_password("mcp-atlassian-oauth", "oauth-{client_id}", json.dumps(tokens))
```

## Suggested Fix

In `_save_tokens()`, ensure both keyring entries are updated:
- `oauth-{client_id}` (used by `load_tokens()`)
- `oauth-{client_id}-cloud-{cloud_id}` (used by `_get_keyring_username()`)

Or consolidate to a single keyring key pattern.

## Environment

- mcp-atlassian version: 0.16.1
- Python: 3.13
- OS: macOS (Sequoia)
- Keyring backend: macOS Keychain

Shahid-LT:
## Summary

Fixes #984

When saving tokens, `_save_tokens()` uses `_get_keyring_username()` which includes `cloud_id` or `base_url` hash in the key (e.g., `oauth-{client_id}-cloud-{cloud_id}`). However, `load_tokens()` uses a simpler pattern (`oauth-{client_id}`) without the context suffix.

This mismatch causes `load_tokens()` to return stale tokens from keyring after re-authentication, because the new tokens are saved under a different key.

## Changes

In `_save_tokens()`, also save tokens to the base `oauth-{client_id}` keyring entry when the context-specific username differs. This ensures `load_tokens()` always finds the latest tokens.

## Testing

1. Authenticate with OAuth: `mcporter auth mcp-atlassian`
2. Wait for token to expire or manually expire it
3. Re-authenticate: `mcporter auth mcp-atlassian --reset`
4. Verify MCP calls work without "Failed to configure OAuth session" error

Before this fix, step 4 would fail because `load_tokens()` would retrieve stale tokens from the `oauth-{client_id}` keyring entry.

### DIFF
--- a/src/mcp_atlassian/utils/oauth.py
+++ b/src/mcp_atlassian/utils/oauth.py
@@ -374,7 +374,9 @@ class OAuthConfig:
             logger.debug(f"Saved OAuth tokens to keyring for {username}")
 
             # Also save to base username for compatibility with load_tokens()
-            # which uses the simpler oauth-{client_id} pattern
+            # which uses the simpler oauth-{client_id} pattern.
+            # Note: If the same client_id is used for both Cloud and DC (rare),
+            # the base key will be overwritten by whichever saves last.
             if username != base_username:
                 keyring.set_password(KEYRING_SERVICE_NAME, base_username, token_json)
                 logger.debug(f"Saved OAuth tokens to keyring for {base_username}")
--- a/tests/unit/utils/test_oauth.py
+++ b/tests/unit/utils/test_oauth.py
@@ -379,6 +379,44 @@ class TestOAuthConfig:
 
     @patch("keyring.set_password")
     @patch.object(OAuthConfig, "_save_tokens_to_file")
+    def test_save_tokens_keyring_success_dc(self, mock_save_to_file, mock_set_password):
+        """Test _save_tokens with successful keyring storage for Data Center."""
+        config = OAuthConfig(
+            client_id="test-client-id",
+            client_secret="test-client-secret",
+            redirect_uri="https://example.com/callback",
+            scope="WRITE",
+            base_url="https://jira.example.com",
+            refresh_token="test-refresh-token",
+            access_token="test-access-token",
+            expires_at=1234567890,
+        )
+        config._save_tokens()
+
+        # Verify keyring was used - should be called twice:
+        # 1. For context-specific key (oauth-{client_id}-dc-{url_hash})
+        # 2. For base key (oauth-{client_id}) for load_tokens() compatibility
+        assert mock_set_password.call_count == 2
+
+        # Check first call (context-specific DC key)
+        first_call = mock_set_password.call_args_list[0]
+        assert first_call[0][0] == KEYRING_SERVICE_NAME
+        assert first_call[0][1].startswith("oauth-test-client-id-dc-")
+        assert "test-refresh-token" in first_call[0][2]
+        assert "test-access-token" in first_call[0][2]
+
+        # Check second call (base key for load_tokens() compatibility)
+        second_call = mock_set_password.call_args_list[1]
+        assert second_call[0][0] == KEYRING_SERVICE_NAME
+        assert second_call[0][1] == "oauth-test-client-id"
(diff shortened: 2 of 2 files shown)


src/mcp_atlassian/utils/oauth.py
```diff
@@ -366,11 +367,18 @@ def _save_tokens(self) -> None:
...
-            # Store the token data in the system keyring
-            keyring.set_password(KEYRING_SERVICE_NAME, username, json.dumps(token_data))
+            token_json = json.dumps(token_data)
 
+            # Store the token data in the system keyring using context-specific key
+            keyring.set_password(KEYRING_SERVICE_NAME, username, token_json)
             logger.debug(f"Saved OAuth tokens to keyring for {username}")
 
+            # Also save to base username for compatibility with load_tokens()
+            # which uses the simpler oauth-{client_id} pattern
```
sooperset: (minor) If the same `client_id` is used for both Cloud and DC (unusual but possible), the base key gets overwritten by whichever saved last. This is fine for the standard single-instance setup, but a brief inline comment noting this would help future readers.

tests/unit/utils/test_oauth.py
```diff
@@ -355,16 +355,24 @@ def test_save_tokens_keyring_success(self, mock_save_to_file, mock_set_password)
...
-        mock_set_password.assert_called_once()
-        service_name = mock_set_password.call_args[0][0]
-        username = mock_set_password.call_args[0][1]
-        token_json = mock_set_password.call_args[0][2]
-
-        assert service_name == KEYRING_SERVICE_NAME
-        assert username == "oauth-test-client-id-cloud-test-cloud-id"
-        assert "test-refresh-token" in token_json
-        assert "test-access-token" in token_json
+        # Verify keyring was used - should be called twice:
```
sooperset, line 358: (minor) This validates the dual-write for Cloud context. Consider adding a similar case for DC context to verify the `dc-{url_hash}` username also triggers the base key write.

sooperset (dismissed):
Good catch on the keyring key mismatch. The dual-write approach is simple and brings keyring into parity with the file storage path (which already uses the base `oauth-{client_id}` pattern). Tests and lint pass cleanly.

src/mcp_atlassian/utils/oauth.py
Shahid-LT: Added inline comment noting the Cloud/DC overwrite behavior for the rare edge case. Thanks for the review!

tests/unit/utils/test_oauth.py
Shahid-LT: Added `test_save_tokens_keyring_success_dc` to verify the dual-write works for Data Center context with the `oauth-{client_id}-dc-{url_hash}` pattern. Good suggestion!

PR #991, opened 2026-02-24
feat(confluence): convert ac:image tags to markdown images

sooperset:
## Summary
- Convert Confluence proprietary `<ac:image>` XML tags to standard markdown images during preprocessing
- Handle `<ri:attachment>` (page-attached files) with attachment list lookup and fallback URL construction
- Handle `<ri:url>` (external images) with filename extraction from URL path
- Preserve `ac:width`/`ac:height` attributes on generated `` tags
- Pass `content_id` and `attachments` to `process_html_content()` at all call sites in `pages.py`

## Test plan
- [x] 9 parametrized test cases covering: basic attachment, external URL, width attributes, mixed content, no content_id fallback, attachment lookup, URL-encoded filenames, unknown inner tags, no-image passthrough
- [x] 4 additional targeted tests for width preservation, mixed content parts, download URL preference, backward compatibility
- [x] Updated 2 existing unit test assertions for new `content_id`/`attachments` params
- [x] All 1967 tests pass (unit + preprocessing)
- [x] Pre-commit clean (Ruff + mypy)

### DIFF
--- a/src/mcp_atlassian/preprocessing/base.py
+++ b/src/mcp_atlassian/preprocessing/base.py
@@ -279,6 +279,9 @@ class BasePreprocessor:
                 filename = ri_att.get("ri:filename", "")
                 alt = filename
 
+                # Check if this references a different page
+                is_cross_page = ri_att.find("ri:page") is not None
+
                 # Try attachment list lookup first
                 url = self._find_attachment_url(filename, attachments)
                 if url:
@@ -287,7 +290,7 @@ class BasePreprocessor:
                         src = f"{self.base_url}{url}"
                     else:
                         src = url
-                elif content_id:
+                elif content_id and not is_cross_page:
                     encoded = urllib.parse.quote(filename, safe="")
                     src = f"{self.base_url}/download/attachments/{content_id}/{encoded}"
                 else:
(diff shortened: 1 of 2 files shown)


src/mcp_atlassian/preprocessing/base.py
```diff
@@ -223,6 +233,95 @@ def _use_fallback_user_mention(self, user_element: Tag, account_id: str) -> None
...
+                if url:
+                    # Prepend base_url if relative path
+                    if url.startswith("/") and self.base_url:
+                        src = f"{self.base_url}{url}"
+                    else:
+                        src = url
+                elif content_id:
+                    encoded = urllib.parse.quote(filename, safe="")
+                    src = f"{self.base_url}/download/attachments/{content_id}/{encoded}"
+                else:
```
sooperset, line 296: (minor) Cross-page attachment references — when `<ri:attachment>` contains a nested `<ri:page ri:content-title="..." ri:space-key="..."/>`, this fallback constructs a URL with the *current* page's content_id rather than the referenced page. Not a regression (these images were previously dropped entirely), but worth noting for a future improvement: could detect `ri:page` child and either skip the fallback or use the filename-only reference instead.

PR #992, opened 2026-02-24
feat(confluence): add get_page_images tool for inline image retrieval

sooperset:
## Summary
- Add `confluence_get_page_images` MCP tool that returns page images as `ImageContent` for direct client rendering
- Two-tier image detection: explicit MIME type check (`image/png`, `image/jpeg`, etc.) + filename extension fallback for `application/octet-stream`
- Uses `fetch_attachment_content()` (from 5.3b) for downloading, with 50MB size limit enforcement
- Non-image attachments are filtered out; summary includes counts and any failures

## Test plan
- [x] `test_get_page_images_basic` — filters images from mixed attachments, returns ImageContent
- [x] `test_get_page_images_octet_stream_fallback` — detects `.jpg` extension when MIME is `application/octet-stream`
- [x] `test_get_page_images_no_images` — returns summary only when no images found
- [x] `test_get_page_images_size_limit` — skips images exceeding 50MB with error in summary
- [x] `test_get_page_images_fetch_failure` — handles `fetch_attachment_content()` returning None
- [x] All 1975 unit tests pass, pre-commit clean (Ruff + mypy)

### DIFF
--- a/src/mcp_atlassian/servers/confluence.py
+++ b/src/mcp_atlassian/servers/confluence.py
@@ -7,7 +7,7 @@ import mimetypes
 from typing import Annotated
 
 from fastmcp import Context, FastMCP
-from mcp.types import BlobResourceContents, EmbeddedResource, TextContent
+from mcp.types import BlobResourceContents, EmbeddedResource, ImageContent, TextContent
 from pydantic import BeforeValidator, Field
 
 from mcp_atlassian.exceptions import MCPAtlassianAuthenticationError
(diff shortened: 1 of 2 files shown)


src/mcp_atlassian/servers/confluence.py
```diff
@@ -1627,3 +1627,206 @@ async def delete_attachment(
...
+    Returns:
+        A list with a text summary followed by one ImageContent per
+        successfully downloaded image.
+    """
+    confluence_fetcher = await get_confluence_fetcher(ctx)
+    contents: list[TextContent | ImageContent] = []
+
+    attachments_result = confluence_fetcher.get_content_attachments(content_id)
+
+    if not attachments_result.get("success"):
```
sooperset, line 1716: (minor) Same pagination limitation as `download_content_attachments` (line 1455) — default `limit=50` means pages with 50+ attachments won't have all images scanned. This is a pre-existing pattern in the codebase, not a regression. Could be addressed as a follow-up for both tools.

sooperset (commented):
Self-review: Solid implementation following existing patterns (download_content_attachments). Two-tier MIME detection works correctly. One known limitation noted inline.

PR #998, opened 2026-02-25
  On Jira Server/DC, GET /rest/api/2/user?username= matches only the …

kalabj:
…login

    name, not email address. When an email identifier is given, _determine_user_api_params
    now calls _lookup_user_directly first (GET /rest/api/2/user/search?username=),
    which searches by email, display name, and login, and then uses the resolved
    username or key for the actual profile fetch. A fallback to passing the email
    directly as username is kept for cases where the login name is the email itself.
    Regression tests for the Server/DC email flow are added.

## Description

Fix email lookup when login is provided instead of username.
Adds fallback handling and regression tests.

## Changes

- Added direct email lookup before resolving username
- Added regression tests for Server/DC flow
- Improved fallback logic

## Testing

- [x] Manual checks performed: `tested throught mcp-inspector`

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/users.py
+++ b/src/mcp_atlassian/jira/users.py
@@ -259,10 +259,27 @@ class UsersMixin(JiraClient):
         # Server/DC: username, key, or email
         elif not self.config.is_cloud:
             if "@" in identifier:
-                api_kwargs["username"] = identifier
-                logger.debug(
-                    f"Determined param: username='{identifier}' (Server/DC email - might not work)"
-                )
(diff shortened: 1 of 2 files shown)


sooperset (approved):
Good fix for a real Server/DC limitation. The approach of resolving email via `_lookup_user_directly` first is sound and the fallback behavior is sensible. Tests cover the three key paths.

Two minor items to address in a follow-up:
1. Test patch path uses `src.` prefix instead of canonical `mcp_atlassian.`
2. Key-vs-username heuristic could be more robust if `_lookup_user_directly` returned field info.

PR #1010, opened 2026-02-25
fix: return descriptive error messages from tool handlers

### LINKED ISSUE #1009: create_issue tool returns generic error instead of Jira API error details
## Problem

When `jira_create_issue` fails due to a Jira API validation error (e.g. a required custom field is missing), the MCP tool returns a generic `"Error calling tool 'create_issue'"` with no details about what went wrong. This makes it impossible for the LLM client to diagnose or recover from the error.

For example, a project requires `customfield_12621` ("Exchange (Multiple)"). The Jira API returns a clear 400 response:

```json
{"errorMessages":[],"errors":{"customfield_12621":"Exchange (Multiple) is required."}}
```

But the MCP tool just returns `"Error calling tool 'create_issue'"`.

## Root Cause

The write tool handlers in `src/mcp_atlassian/servers/jira.py` do not catch exceptions from the underlying Jira API calls. When an exception propagates up to the MCP framework, it is caught and returned as a generic error message with no details.

## Suggested Fix

Add a `handle_tool_errors` decorator in `src/mcp_atlassian/utils/decorators.py` that catches exceptions and returns a JSON object with the error details. Integrate it into the existing `check_write_access` decorator so all write tools automatically get error handling in one place, without per-callsite try/except blocks.

PR: #1010

## Environment

- mcp-atlassian version: 0.11.9
- Jira Server/Data Center (not Cloud)

baumgold:
## Summary

Fixes #1009

When tool handlers raise exceptions, FastMCP's `mask_error_details` setting (if enabled) strips the original error message, producing generic errors like `"Error calling tool 'create_issue'"` with no actionable details.

This PR adds a `handle_tool_errors` decorator that catches exceptions and re-raises them as `ToolError`. Since `ToolError` bypasses FastMCP's error masking, descriptive error messages always reach MCP clients regardless of server configuration.

## Changes

- **`handle_tool_errors` decorator** — Catches non-`ToolError` exceptions, logs them, and re-raises as `ToolError(str(e))` with `isError=true` properly set in the MCP response
- **Integrated into `check_write_access`** — All write tools automatically get error handling via decorator composition
- **Tool name captured before `@wraps`** — Ensures correct tool name in log messages
- **New unit tests** — Tests for exception wrapping, `ToolError` passthrough, and normal return value preservation

## Test plan

- [x] All 2113 existing tests pass
- [x] New `handle_tool_errors` decorator tests cover exception wrapping, `ToolError` passthrough, and normal return
- [x] `check_write_access` tests verify read-only mode raises `ToolError`
- [x] Jira server tests verify invalid JSON inputs raise `ToolError` with descriptive messages

### DIFF
--- a/src/mcp_atlassian/utils/decorators.py
+++ b/src/mcp_atlassian/utils/decorators.py
@@ -1,4 +1,3 @@
-import json
 import logging
 from collections.abc import Awaitable, Callable
 from functools import wraps
@@ -6,6 +5,7 @@ from typing import Any, TypeVar
 
 import requests
 from fastmcp import Context
+from fastmcp.exceptions import ToolError
 from requests.exceptions import HTTPError
 
 from mcp_atlassian.exceptions import MCPAtlassianAuthenticationError
@@ -18,25 +18,26 @@ F = TypeVar("F", bound=Callable[..., Awaitable[Any]])
 
 def handle_tool_errors(func: F) -> F:
     """
-    Decorator for FastMCP tool handlers that catches exceptions and returns
-    descriptive JSON error messages instead of letting them propagate to the
-    MCP framework (which would produce a generic error).
+    Decorator for FastMCP tool handlers that catches exceptions and re-raises
+    them as ToolError with the original error message preserved.
+
+    ToolError bypasses FastMCP's mask_error_details setting, ensuring that
+    descriptive error messages are always sent to MCP clients regardless of
+    server configuration.
 
     Assumes the decorated function is async.
     """
+    tool_name = func.__name__
 
     @wraps(func)
     async def wrapper(*args: Any, **kwargs: Any) -> Any:
         try:
             return await func(*args, **kwargs)
+        except ToolError:
+            raise
         except Exception as e:
-            tool_name = func.__name__
             logger.error(f"Error in tool '{tool_name}': {e}", exc_info=True)
-            return json.dumps(
-                {"error": str(e), "tool": tool_name},
-                indent=2,
-                ensure_ascii=False,
-            )
+            raise ToolError(str(e)) from e
 
     return wrapper  # type: ignore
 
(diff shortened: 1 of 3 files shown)


sooperset:
Thank you for this contribution, @baumgold! Better error messages for LLM clients is a valuable goal. I've done a thorough review and found a few issues worth discussing:

**1. Errors are returned as successful results (`isError=false`)**

The `handle_tool_errors` decorator catches exceptions and returns a JSON string, which causes FastMCP to emit `CallToolResult(isError=false, content=[TextContent(...)])`. MCP clients that rely on the standard `isError` field will treat these errors as successful tool executions.

The correct approach would be to raise `ToolError` from `fastmcp.exceptions`, which FastMCP converts to `CallToolResult(isError=true)` with the error message preserved.

**2. `tool_name` resolves to `"wrapper"` instead of the actual tool name**

Due to decorator application order in `check_write_access`, `func.__name__` inside `handle_tool_errors` resolves to `"wrapper"`, not the actual tool name like `"create_issue"`. The `@wraps(func)` decorator runs after `@handle_tool_errors` in the execution order.

**3. FastMCP already provides detailed error messages by default**

FastMCP's `mask_error_details` setting defaults to `False`. With the default, `tool_manager.call_tool()` wraps exceptions as `ToolError(f"Error calling tool {key!r}: {e}")` — the original exception message is already included. The problem described in the PR ("the MCP client only sees 'Error calling tool create_issue'") only occurs when `mask_error_details=True`, which this project does not configure. Could you verify whether your specific setup enables this setting?

**Suggested fix:** If the goal is to ensure descriptive errors even when `mask_error_details=True`, the right approach would be:
```python
from fastmcp.exceptions import ToolError

# In the decorator:
raise ToolError(str(e)) from e
```
`ToolError` bypasses error masking in FastMCP and properly sets `isError=true`.

Would love to discuss further. Thanks again for the thoughtful PR!

baumgold:
Thank you for the thorough review, @sooperset! All three points were spot on. I've pushed a new commit addressing each:

1. **`ToolError` instead of JSON strings** — The decorator now raises `ToolError(str(e))`, which FastMCP properly converts to `CallToolResult(isError=true)` with the original error message preserved.

2. **`tool_name` fix** — `func.__name__` is now captured _before_ `@wraps` runs, so the logged tool name is correct (e.g. `"jira_create_issue"` instead of `"wrapper"`).

3. **`ToolError` passthrough** — Existing `ToolError` exceptions are re-raised directly to avoid double-wrapping.

Regarding `mask_error_details`: you're right that it defaults to `False`. In our deployment we're hitting the masked path, so this decorator ensures descriptive errors reach clients regardless of that setting.

I also added dedicated unit tests for `handle_tool_errors` covering the three key behaviors (exception wrapping, `ToolError` passthrough, and normal return). All 2113 tests pass.

Please let me know if there's anything else you'd like adjusted!

sooperset:
Great work addressing all the feedback, @baumgold! The ToolError approach, tool_name capture fix, and passthrough logic all look solid. Merging now — thank you for this contribution!

Issue #860, opened 2026-01-19
Bearer token auth overridden by .netrc credentials


hteichmann-strato:
## Bug Description

When a user has credentials for their Jira Server in `~/.netrc`, the Python `requests` library automatically uses those credentials instead of the explicitly provided `JIRA_PERSONAL_TOKEN`. This causes 401 authentication errors even though the token is valid.

## Root Cause

The `requests` library defaults to `trust_env=True`, which reads `~/.netrc` for authentication. When netrc credentials exist for the Jira host, they override the Bearer token header that mcp-atlassian sets.

## Steps to Reproduce

1. Have a valid `JIRA_PERSONAL_TOKEN` configured
2. Have an entry in `~/.netrc` for the same Jira host:
   ```
   machine jira.example.org
   login myuser
   password oldpassword
   ```
3. Try to use any Jira MCP tool
4. Get 401 Unauthorized error

## Debug Evidence

When inspecting the actual request headers sent:
```python
# Expected:
'Authorization': 'Bearer <your-configured-token>'

# Actual (due to netrc override):
'Authorization': 'Basic <base64-of-netrc-credentials>'
```

## Proposed Fix

Set `trust_env=False` on the session when explicit credentials are provided:

```python
# In client.py, after creating the Jira client
if self.config.personal_token or self.config.api_token:
    self.jira._session.trust_env = False
```

This ensures explicit configuration takes precedence over implicit netrc configuration.

## Workaround

Remove or comment out the Jira host entry from `~/.netrc`.

## Environment

- mcp-atlassian: latest (main branch)
- Jira Server v10.3.15
- Python requests library auto-reads netrc by default

PR #1012, opened 2026-02-25
Fixes corporate CA bundle with PAT auth type

bbartels:
## Description
After setting `trust_env=False` on the requests session for PAT and OAuth auth types (to prevent `.netrc` from overriding explicit credentials, per #860), the `REQUESTS_CA_BUNDLE` and `CURL_CA_BUNDLE` environment variables are also silently disabled. This breaks SSL verification in environments that rely on custom or internal Certificate Authorities (e.g. corporate environments with internal CAs not in Mozilla's public CA bundle).
This PR restores the CA bundle from environment variables after `trust_env=False` is set, by explicitly setting `session.verify` to the CA bundle path. This preserves the `.netrc` protection from #860 while fixing SSL verification for internal CAs.
Fixes: #860 (unintended side effect)
## Changes
- After setting `trust_env=False` in the Jira client, explicitly restore `session.verify` from `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` if available
- Same fix applied to the Confluence client
## Testing
- [x] Manual checks performed: Deployed mcp-atlassian in a Docker container with an internal CA (GRIssuingCA2) provided via `REQUESTS_CA_BUNDLE`. Confirmed that PAT auth with `trust_env=False` now completes SSL handshake successfully against an internal Jira Server instance, whereas previously it failed with `SSLCertVerificationError: certificate verify failed: self-signed certificate in certificate chain`.
## Checklist
- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/client.py
+++ b/src/mcp_atlassian/jira/client.py
@@ -124,6 +124,13 @@ class JiraClient:
         # explicit credentials (#860). Basic auth can safely use .netrc.
         if self.config.auth_type in ("pat", "oauth"):
             self.jira._session.trust_env = False
+            # Restore CA bundle -- trust_env=False also disables REQUESTS_CA_BUNDLE
+            # which breaks SSL in environments with internal/corporate CAs.
+            # See: https://requests.readthedocs.io/en/latest/api/#requests.Session.verify
+            if self.config.ssl_verify:
+                ca_bundle = os.environ.get("REQUESTS_CA_BUNDLE")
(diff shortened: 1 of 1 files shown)


bbartels:
@sooperset https://github.com/sooperset/mcp-atlassian/pull/951 broke corporate ca bundles, this is a fix if you dont mind having a look :)

bbartels:
ignore, https://github.com/sooperset/mcp-atlassian/pull/976 is a better fix

Issue #896, opened 2026-02-03
[Bug]: fixVersions field not returned when explicitly requested in fields parameter


FrogDevelopper:
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/sooperset/mcp-atlassian/issues) to make sure this bug has not already been reported.
- [x] I have checked the [README](https://github.com/sooperset/mcp-atlassian/blob/main/README.md) for relevant information.

### Bug Description

The `jira_get_issue` and `jira_search` tools do not return the `fixVersions` field when it is explicitly requested in the `fields` parameter. Other fields in the same request are returned correctly.

### Steps to Reproduce

1. Call `jira_get_issue` with explicit fields including `fixVersions`:
  ```python
  jira_get_issue(
      issue_key="KEY-1234",
      fields="summary,description,status,assignee,fixVersions"
  )
```

2. Observe the response - `fixVersions` is missing

### Expected Behavior

The response should include the fixVersions field:
```json
  {
    "id": "16650465",]
    "key": "KEY-1234",
    "summary": "...",
    "description": "...",
    "status": {...},
    "assignee": {...},\
    "fixVersions": [...],  // Should be present
  }
```

### Actual Behavior

The fixVersions field is silently omitted from the response, even though it was explicitly requested:
```json
  {
    "id": "16650465",
    "key": "KEY-1234",
    "summary": "...",
    "description": "...",
    "status": {...},
    "assignee": {...},
    // fixVersions missing
  }
```

### mcp-atlassian Version

Latest version (installed via uvx mcp-atlassian)

### Installation Method

Other

### Operating System

macOS

### Python Version

3.12

### Atlassian Instance Type

Jira Server / Data Center

### Client Application

Claude Code (skill)

### Additional Context

Additional Context 

  - The JIRA REST API directly returns `fixVersions` when requested
  - Other array fields work correctly (e.g., labels, components)
  - Both `fixVersions` (camelCase) and `fix_versions` (snake_case) fail when explicitly requested
  - Only `*all` returns the field as `fix_versions`

### Impact

We cannot efficiently retrieve fix version information without fetching all fields, which are numerous, making it impractical to validate release versions in automated workflows.

arthurmmedeiros:
I am having the same problem with the `versions` field. Is it possible to add it to the #[issue](https://github.com/sooperset/mcp-atlassian/blob/main/src/mcp_atlassian/models/jira/issue.py) model?

PR #1027, opened 2026-02-26
fix(jira): fix "fixVersions" in get jira issue

yliu:
The jira field key is 'fixVersions',  so it should use "fixVersions" instead of "fix_versions" when check "should include field".

## Description

Fixes: #896

## Changes

- use "fixVersions" instead of "fix_versions"

## Testing

- [x] Manual checks performed: `with fixVersions in "Get Issue" fields argument, it could return with "fix_versions"`

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/models/jira/issue.py
+++ b/src/mcp_atlassian/models/jira/issue.py
@@ -571,7 +571,7 @@ class JiraIssue(ApiModel, TimestampMixin):
         if self.components and should_include_field("components"):
             result["components"] = self.components
 
-        if self.fix_versions and should_include_field("fix_versions"):
+        if self.fix_versions and should_include_field("fixVersions"):
             result["fix_versions"] = self.fix_versions
 
         # Add epic fields if available and requested


src/mcp_atlassian/models/jira/issue.py
```diff
@@ -571,7 +571,7 @@ def should_include_field(field_name: str) -> bool:
         if self.components and should_include_field("components"):
             result["components"] = self.components
 
-        if self.fix_versions and should_include_field("fix_versions"):
+        if self.fix_versions and should_include_field("fixVersions"):
```
sooperset, line 574: The fix is correct — nice catch.

However, the same naming mismatch exists at line 520: `should_include_field("issue_type")` should be `should_include_field("issuetype")` since the Jira API field name is `issuetype`. That's pre-existing and not your fault, but since you're already here, it would be great to fix it in the same PR.

A more robust approach would be to normalize field names in `should_include_field` so it accepts both the Jira API name and the model-internal name:

```python
FIELD_NAME_ALIASES = {
    "fixVersions": "fix_versions",
    "fix_versions": "fixVersions",
    "issuetype": "issue_type",
    "issue_type": "issuetype",
}

def should_include_field(field_name: str) -> bool:
    if self.requested_fields == "*all" or not isinstance(self.requested_fields, list):
        return True
    if field_name in self.requested_fields:
        return True
    alias = FIELD_NAME_ALIASES.get(field_name)
    return alias is not None and alias in self.requested_fields
```

This would prevent this class of bug from recurring and handle both naming conventions.

sooperset, line 574: A regression test is missing. The PR checklist says "Unit tests added/updated" but the diff only touches the source file. Per project conventions, bug fixes need regression tests.

Consider adding a test like:

```python
def test_fix_versions_with_explicit_field_request(self, jira_issue_data):
    """Test that fixVersions is returned when explicitly requested (issue #896)."""
    issue = JiraIssue.from_api_response(
        jira_issue_data, requested_fields=["fixVersions"]
    )
    simplified = issue.to_simplified_dict()
    assert "fix_versions" in simplified
    assert simplified["fix_versions"] == ["v1.0"]
```

sooperset:
Thank you for this fix, @yliu! The one-line change correctly maps `fixVersions` to its API field name, and I've confirmed the logic is sound.

However, per project conventions (AGENTS.md), bug fixes need regression tests. Could you add a test case that:

1. Creates a `JiraIssue` with `fix_versions=["v1.0"]` and `requested_fields=["fixVersions"]`
2. Asserts `"fix_versions"` appears in `to_simplified_dict()` output

A test like this in `tests/unit/models/jira/test_issue.py` would prevent regressions.

Also worth noting: the same pattern mismatch exists at line 520 where `should_include_field("issue_type")` should use `"issuetype"` (the Jira API name). If you'd like to fix both in this PR, that would be great — otherwise we can handle it separately.

Thanks again for the contribution!

sooperset:
Thank you for the contribution, @yliu! This has been reimplemented in #1072 with regression tests and updated project conventions. Your work is credited via Reported-by trailer.

The fix also audited all other `should_include_field()` calls to confirm no similar mismatches exist.

PR #1030, opened 2026-02-26
fix(jira):fix allow underscores in issue and project key patterns

alvaroEset:
## Description

Some Jira Server/Data Center instances use non-standard project key formats that include underscores (e.g., D_DEV-123). The previous patterns only allowed uppercase letters and digits, causing all MCP tools that accept an issue_key parameter (jira_get_issue, jira_update_issue, jira_transition_issue, etc.) to reject these keys at validation time — before any API call was even made.

## Changes

- Allow Underscores in Jira Issue and Project Key Patterns

## Testing

- [X] Unit tests added/updated
- [X] Integration tests passed

## Checklist

- [X] Code follows project style guidelines (linting passes).
- [X] Tests added/updated for changes.
- [X] All tests pass locally.


sooperset:
Thank you for this fix, @alvaroEset! The regex update is clean and the included tests cover the key edge cases well.

Merged! We'll follow up separately to update a couple of similar patterns in the preprocessing module that have the same limitation.

Issue #527, opened 2025-06-18
[Feature]: OAuth support for Confluence Data Center on prem


sergiobank:
### Is your feature request related to a problem? Please describe.

Hey
First of all, thanks for the amazing work put into the project!
I would like to know if the existing OAuth solution for Confluence Cloud could be used for Confluence Data Center on premises OAuth authentication flow, or if it could be easily adapted to support this important feature.
I think it would be a terrific feature for enterprise use
Thanks in advance!

### Describe the solution you'd like

Hey
First of all, thanks for the amazing work put into the project!
I would like to know if the existing OAuth solution for Confluence Cloud could be used for Confluence Data Center on premises OAuth authentication flow, or if it could be easily adapted to support this important feature.
I think it would be a terrific feature for enterprise use
Thanks in advance!

sooperset:
Hi @sergiobank,

Thanks for the request! Unfortunately, this would be quite complicated to implement. Confluence Cloud uses OAuth 2.0, but Data Center requires OAuth 1.0a which is a completely different protocol. 

Given the complexity and maintenance burden, I'm not confident I can implement this properly. Would recommend using PAT tokens instead 🙏

sergiobank:
Hey @sooperset
Actually Confluence Data Center does support OAuth 2.0 (https://confluence.atlassian.com/confkb/oauth-2-0-configuration-for-confluence-1224638905.html)
Because of that, I'm confident this features is supported by minimal changes, and it would satisfy most of the enterprises authentication needs
Let me know what do you think about that and what would it take
Thanks in advance, and thanks again for the amazing work!

sooperset:
@sergiobank Ah, you're right! thanks for the correction and the link.

I don't have a Data Center test environment right now, but I can work on implementing this. Would you be able to help test it once I have a branch ready?

sergiobank:
Sure!! Amazing!!

sooperset:
@sergiobank I've implemented OAuth 2.0 support for Confluence Data Center in branch `feature/confluence-datacenter-oauth-support`. Ready for you to test!

Quick Setup:

1. Clone and checkout the branch:
   ```bash
   git clone https://github.com/sooperset/mcp-atlassian.git
   cd mcp-atlassian
   git checkout feature/confluence-datacenter-oauth-support
   ```

2. Set environment variables:
   ```bash
   export ATLASSIAN_OAUTH_CLIENT_ID="your-client-id"
[...]

PR #1054, opened 2026-02-27
feat(auth): add opt-in OAuth proxy + DCR/discovery endpoints

### LINKED ISSUE #610: [Feature]: Enabling multi-user OAuth flow for self-hosted remote MCP (like Atlassian-hosted remote MCP)

Hi team,

We’re trying to replicate the behavior described in [this Atlassian community post](https://community.atlassian.com/forums/Atlassian-Platform-articles/Using-the-Atlassian-Remote-MCP-Server-beta/ba-p/3005104) using our **own self-hosted remote MCP server**, which we currently host as an ECS service.

When using Atlassian’s hosted remote MCP at `https://mcp.atlassian.com/v1/sse`, users are automatically redirected to the OAuth 2.0 wizard, where they can authenticate and grant access. This per-user OAuth flow works great — each user is identified individually, and tokens are handled dynamically.

Our goal is to enable **the same behavior** in our self-hosted remote MCP, where:

- Users accessing the MCP (via **Claude Desktop**) are automatically redirected to the Atlassian OAuth wizard if they haven’t authenticated yet
- Each user goes through the OAuth 2.0 flow and obtains their own access token
- Those tokens are stored and scoped based on their Atlassian permissions (projects, spaces, etc.)
- Authorization is **per user**, not globally tied to the deployer

Currently, when setting up a remote MCP, the workflow relies on a one-time manual auth to obtain tokens. We’d like to know if the MCP codebase supports (or plans to support) this **multi-user OAuth model** for self-hosted deployments.

Is there a way to configure the app so that self-hosted remote MCPs behave the same way as `https://mcp.atlassian.com/v1/sse`, prompting each user with their own auth flow?

Any guidance or best practices would be appreciated!

Thanks in advance.

_No response_

### Use Case

_No response_

### Additional Context

_No response_

Poggen:
## Summary
This adds an **opt-in** OAuth proxy mode for remote MCP deployments that need spec-aligned OAuth onboarding (discovery + DCR + auth code flow), without changing existing default auth behavior.

When `ATLASSIAN_OAUTH_PROXY_ENABLE=true`, the server now exposes:
- `/.well-known/oauth-protected-resource`
- `/.well-known/oauth-authorization-server`
- `POST /register`
- `GET /authorize`
- `POST /token`
- callback route derived from `ATLASSIAN_OAUTH_REDIRECT_URI`

## What is included
- New `HardenedOAuthProxy` wrapper with DCR hardening for allowed grant types/scopes.
- Opt-in wiring in main server bootstrap (`auth=_build_auth_provider()` only when enabled).
- Cloud/DC upstream endpoint resolution:
  - Cloud -> `auth.atlassian.com` endpoints (+ `audience`/`prompt` authorize params)
  - Data Center -> `<instance>/rest/oauth2/latest/{authorize,token}`
- FastMCP auth-context token resolution in dependencies (`get_access_token()`), so OAuth tool calls use the upstream token when available.
- Env/docs updates for proxy mode and DCR controls.
- Unit coverage for provider construction/hardening and token resolution behavior.

## Backward compatibility
- Proxy mode is disabled by default.
- Existing API token, PAT, and header-based OAuth flows continue to work unchanged unless enabled.

## Validation
- `uv run pytest -q`

Closes #610
Refs #527

### DIFF
--- a/src/mcp_atlassian/servers/dependencies.py
+++ b/src/mcp_atlassian/servers/dependencies.py
@@ -11,7 +11,7 @@ from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
 from fastmcp import Context
-from fastmcp.server.dependencies import get_http_request
+from fastmcp.server.dependencies import get_access_token, get_http_request
 from starlette.requests import Request
 
 from mcp_atlassian.confluence import ConfluenceConfig, ConfluenceFetcher
@@ -27,6 +27,29 @@ if TYPE_CHECKING:
     from mcp_atlassian.jira.config import JiraConfig as UserJiraConfigType
 
 logger = logging.getLogger("mcp-atlassian.servers.dependencies")
+
+
+def _resolve_oauth_access_token(fallback_token: str, service: str) -> str:
+    """Resolve upstream OAuth token from FastMCP auth context with safe fallback."""
+    try:
+        access_token = get_access_token()
+    except Exception as e:
+        logger.debug(
+            "OAuth token resolution via FastMCP context failed for %s; "
+            "using request token (%s)",
+            service,
+            e,
+        )
+        return fallback_token
+
+    if access_token and access_token.token:
+        if access_token.token != fallback_token:
+            logger.debug(
+                "Resolved upstream %s token from FastMCP auth context", service
+            )
+        return access_token.token
+
+    return fallback_token
 
 
 def _make_ssrf_safe_hook(
@@ -407,7 +430,9 @@ async def get_jira_fetcher(ctx: Context) -> JiraFetcher:
 
             credentials = {"user_email_context": user_email}
             if resolved_auth_type == "oauth":
-                credentials["oauth_access_token"] = user_token
+                credentials["oauth_access_token"] = _resolve_oauth_access_token(
+                    user_token, "Jira"
+                )
             else:
                 credentials["personal_access_token"] = user_token
 
@@ -641,7 +666,9 @@ async def get_confluence_fetcher(ctx: Context) -> ConfluenceFetcher:
 
             credentials = {"user_email_context": user_email}
             if resolved_auth_type == "oauth":
-                credentials["oauth_access_token"] = user_token
+                credentials["oauth_access_token"] = _resolve_oauth_access_token(
+                    user_token, "Confluence"
+                )
             else:
                 credentials["personal_access_token"] = user_token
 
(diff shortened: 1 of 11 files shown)


src/mcp_atlassian/servers/dependencies.py
```diff
@@ -29,6 +29,29 @@
...
+    try:
+        access_token = get_access_token()
+    except Exception as e:
+        logger.debug(
+            "OAuth token resolution via FastMCP context failed for %s; "
+            "using request token (%s)",
+            service,
+            e,
+        )
+        return fallback_token
```
sooperset, line 43: (minor) Catching bare `Exception` here is very broad. In proxy mode, if `get_access_token()` fails for an unexpected reason (e.g. `AttributeError` from a FastMCP version mismatch), the fallback silently sends the wrong token to Atlassian — the MCP client's local session token instead of the upstream Atlassian token. The resulting 401 would be hard to diagnose since this logs at DEBUG level.

Consider narrowing to `(RuntimeError, LookupError)` or bumping to WARNING when the fallback is used:
```python
except (RuntimeError, LookupError) as e:
    logger.debug(...)
```

src/mcp_atlassian/servers/oauth_proxy.py
```diff
@@ -0,0 +1,75 @@
...
+            forced_scope = " ".join(self._forced_scopes).strip()
+            updates["scope"] = forced_scope or None
+            if client_info.scope and client_info.scope != forced_scope:
+                logger.warning(
+                    "DCR requested scope '%s'; enforcing '%s'",
+                    client_info.scope,
+                    forced_scope,
+                )
+
+        if updates:
```
sooperset, line 69: (minor) `updates` is initialized with `{"response_types": ["code"]}` on line 47, so this guard is always True. Not a bug, just dead code — could remove the conditional and always run `model_copy`.

sooperset (approved):
Well-structured PR that cleanly adds opt-in OAuth proxy support. The `HardenedOAuthProxy` subclass is a good pattern for enforcing DCR restrictions without modifying the base FastMCP behavior. Test coverage is solid (16 new tests for proxy build, 5 for token resolution, 2 for verifier). All 2345 unit tests pass with no regressions, pre-commit clean.

Two small items to consider (non-blocking):

1. `_resolve_oauth_access_token` catches bare `Exception`. In proxy mode, if `get_access_token()` fails unexpectedly, the fallback sends the MCP client's local session token to Atlassian instead of the upstream token — resulting in a confusing 401 rather than a clear auth-context error. Consider narrowing to `(RuntimeError, LookupError)` or logging at WARNING level.

2. (minor) In `HardenedOAuthProxy.register_client()`, the `updates` dict is always non-empty (`response_types` is set unconditionally), so the `if updates:` guard is always True.

Issue #1008, opened 2026-02-25
[Feature]: `DISABLED_TOOLS`/`--disabled-tools`


vrmiguel99:
### Is your feature request related to a problem? Please describe.

Hi there!

I know that we have the opposite: `ENABLED_TOOLS`/`--enabled-tools`, (and the read-only mode) which works fine. In my case, though, I'd like to only disallow `delete_issue`, while keeping all the rest. Would be nice to be able to just deny that one and keep all the rest.

### Describe the solution you'd like

`DISABLED_TOOLS`/`--disabled-tools`

### Describe alternatives you've considered

```ts
// Allow everything but `delete_issue`
const allowedTools = [
	'jira_search',
	'jira_get_issue',
	'jira_get_all_projects',
	'jira_get_project_issues',
	'jira_get_worklog',
	'jira_get_transitions',
	'jira_search_fields',
	'jira_get_agile_boards',
	'jira_get_board_issues',
	'jira_get_sprints_from_board',
	'jira_get_sprint_issues',
	'jira_get_issue_link_types',
	'jira_batch_get_changelogs',
	'jira_get_user_profile',
	'jira_download_attachments',
	'jira_get_project_versions',
	'jira_get_issue_development_info',
	'jira_get_issues_development_info',
	'jira_get_project_components',
	'jira_get_service_desk_for_project',
	'jira_get_service_desk_queues',
	'jira_get_queue_issues',
	'jira_create_issue',
	'jira_update_issue',
	'jira_batch_create_issues',
	'jira_add_comment',
	'jira_transition_issue',
	'jira_add_worklog',
	'jira_link_to_epic',
	'jira_create_sprint',
	'jira_update_sprint',
	'jira_create_issue_link',
	'jira_remove_issue_link',
	'jira_create_version',
	'jira_batch_create_versions',
];
```

sooperset:
Thank you for the suggestion! After review, we believe the existing configuration options already cover this use case:

- **`ENABLED_TOOLS`**: An allowlist that restricts available tools to only those specified. This is more secure than a denylist approach since new tools are blocked by default.
- **`READ_ONLY_MODE=true`**: Blocks all write operations at the server level.

Together these provide fine-grained control over tool availability. We're closing this as covered by existing functionality. If you have a specific scenario that these options don't address, please let us know!

PR #1088, opened 2026-03-01
feat(server): add ALLOW_DELETE_TOOLS safety filter for destructive operations

Troubladore:
## Description

Adds a safety layer for destructive delete operations, complementary to `READ_ONLY_MODE` and the `DISABLED_TOOLS` feature from #1008.

While `DISABLED_TOOLS` lets users disable any specific tool, `ALLOW_DELETE_TOOLS` provides a simpler UX for the common safety case: allow all writes (create, update) but gate deletes behind an explicit opt-in. This is useful for teams that want agents to create and modify content but not permanently destroy it.

**New env var:** `ALLOW_DELETE_TOOLS` (default: `false`)

When `false`, the following tools are excluded from the tool list:
- `jira_delete_issue`
- `confluence_delete_page`
- `confluence_delete_attachment`

This is independent from `READ_ONLY_MODE` — you can have writes enabled but deletes blocked.

## Changes

- Add `ALLOW_DELETE_TOOLS` env var parsing in `utils/io.py` (uses existing `is_env_extended_truthy` helper)
- Add `allow_delete` field to `MainAppContext`
- Tag delete tools with `"delete"` in server tool definitions (`jira.py`, `confluence.py`)
- Filter delete-tagged tools in `_list_tools_mcp()` when `allow_delete=False`
- Log delete tool status at server startup
- Update `.env.example` with documented variable

## Testing

- [x] Parametrized tests for env var parsing (truthy/falsy values)
- [x] Tag validation test ensuring all delete tools have the `"delete"` tag
- [x] All existing tests pass (2538 passed)

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -145,6 +145,12 @@ CONFLUENCE_API_TOKEN=your_confluence_api_token_here
 # Disables all write operations (create, update, delete). Default is false.
 #READ_ONLY_MODE=false
 
+# --- Delete Tools Safety Filter ---
+# Controls whether destructive delete tools (delete issue, delete page, etc.)
+# are exposed. When false (default), delete tools are hidden from the tool listing.
+# Set to true to allow delete operations.
+#ALLOW_DELETE_TOOLS=false
+
 # --- Ignore Proxy Auth Headers ---
 # Set to 'true' to ignore Authorization headers from reverse proxies
 # (e.g., GCP Cloud Run, AWS ALB). When enabled, the server uses only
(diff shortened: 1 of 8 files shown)


sooperset:
Thank you for the PR @Troubladore! After consideration, we think the existing READ_ONLY_MODE provides sufficient safety for destructive operations at the server level. Adding a separate ALLOW_DELETE_TOOLS filter creates overlapping configuration that increases complexity without meaningful safety gain. Closing for now, but appreciate the thoughtfulness around safety controls.

PR #1090, opened 2026-03-01
feat(confluence): add get_space_page_tree tool for page hierarchy discovery

### LINKED ISSUE #1085: [Feature]: Add tool to discover Confluence space page hierarchy

  When an AI agent needs to create a page in the right place, or understand how a Confluence space is organized, there's no way to get the full hierarchy in one call. The existing confluence_get_page_children tool only returns one level of children, so mapping out a space requires multiple round-trips — each consuming tokens and adding latency.

  As Jesse Vincent describes in https://blog.fsck.com/2025/10/19/mcps-are-not-like-other-apis/, MCP tools should support progressive disclosure and minimize the number of calls an agent needs to understand its environment. Right now, understanding a space structure requires the agent to recursively explore it one level at a time, which is is sometimes fragile or expensive.

Add a new read-only tool confluence_get_space_page_tree that returns a flat list of all pages in a space with parent relationship metadata. This gives agents the full picture in a single call.

Parameters: space_key (required), limit (optional, default 100)

Response schema:
  {
    "space_key": "DOC",
    "total_pages": 42,
    "has_more": false,
    "next_start": null,
    "pages": [
      {"id": "12345", "title": "Documentation Home", "parent_id": null, "position": 0, "depth": 0},
      {"id": "12346", "title": "Getting Started", "parent_id": "12345", "position": 0, "depth": 1}
    ]
  }

The flat list with parent_id and depth is deliberately more token-efficient than rendering ASCII trees. The position field preserves sibling ordering so agents can reconstruct the visual layout if needed. Pagination metadata (has_more, next_start) lets agents handle large spaces without surprise truncation.

This fits naturally in the confluence_pages toolset alongside the existing page tools.

- Add recursive=true to confluence_get_page_children — This mixes the "get one level" and "get everything" use cases in one tool. Separate tools are clearer for agents.
- Return an ASCII tree — More human-readable, but significantly more tokens for the same information. A flat list with depth/parent_id is cheaper to transmit and easier for agents to filter or query programmatically.
- Keep requiring multiple calls — This is the current state. It works, but it's the kind of "make the agent figure it out" approach that the MCP philosophy argues against.

### Use Case

- Page creation placement: Before creating a page, an agent can see the full hierarchy and pick the right parent. No guessing, no retry loops.
- Space audits: Reviewing space structure (depth, orphan pages, organization) in a single call.
- Page move planning: Understanding current structure before reorganizing pages.
- Onboarding context: When an agent is asked to "add documentation to the project space," it can immediately see what's already there.

### Additional Context

I have an implementation ready and will submit a PR. The implementation uses get_all_pages_from_space with ancestor expansion, builds the flat list server-side, and sorts breadth-first by depth then position. No new dependencies required.

Troubladore:
## Description

Implements #1085. Adds a new read-only tool `confluence_get_space_page_tree` that returns a flat, token-efficient representation of all pages in a Confluence space with parent relationships.

As discussed in [MCPs Are Not Like Other APIs](https://blog.fsck.com/2025/10/19/mcps-are-not-like-other-apis/), MCP tools should support progressive disclosure and minimize the calls an agent needs to understand its environment. Currently, mapping a space hierarchy requires recursive `get_page_children` calls — fragile and expensive. This tool gives agents the full picture in one call.

**Response:**
```json
{
  "space_key": "DOC",
  "total_pages": 42,
  "has_more": false,
  "next_start": null,
  "pages": [
    {"id": "12345", "title": "Home", "parent_id": null, "position": 0, "depth": 0},
    {"id": "12346", "title": "Getting Started", "parent_id": "12345", "position": 0, "depth": 1}
  ]
}
```

Fixes #1085

## Changes

- Add `get_space_page_tree()` method to `PagesMixin` — fetches pages with ancestor expansion, builds flat list with `parent_id`/`depth`/`position`, sorts breadth-first
- Add `get_space_page_tree` server tool in `confluence_pages` toolset (read-only)
- Include pagination metadata (`has_more`, `next_start`, `hint`) when results are truncated at limit

## Testing

- [x] 4 unit tests for client method (empty space, single root, children with depth, position sorting)
- [x] 1 server tool test
- [x] Tool count assertion updated (23 → 24)
- [x] All 2531 tests pass

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/confluence/pages.py
+++ b/src/mcp_atlassian/confluence/pages.py
@@ -937,6 +937,8 @@ class PagesMixin(ConfluenceClient):
                 result["next_start"] = start
             return result
 
+        except HTTPError:
+            raise  # let @handle_auth_errors decorator handle auth errors
         except Exception as e:
             logger.error(f"Error fetching page tree for space '{space_key}': {e}")
             raise Exception(f"Failed to fetch page tree: {e}") from e


src/mcp_atlassian/confluence/pages.py
```diff
@@ -704,6 +704,92 @@ def get_page_children(
...
+            - Note: parent_id is None for root pages
+
+        Raises:
+            Exception: If there is an error fetching pages
+        """
+        try:
+            # Fetch all pages from the space
+            pages = self.confluence.get_all_pages_from_space(
+                space=space_key,
+                start=0,
```
sooperset: **BLOCKER: No pagination — silently returns incomplete results for large spaces.**

`get_all_pages_from_space()` from atlassian-python-api makes a single `GET /rest/api/content` call. Confluence enforces a server-side max of ~200-250 results per request, regardless of the `limit` parameter. A space with 300 pages and `limit=500` will silently return only ~200-250 pages.

This breaks the tool's core value proposition ("get the full picture in one call"). Need a pagination loop:

```python
all_pages = []
start = 0
batch_size = min(limit, 200)
while len(all_pages) < limit:
    fetch_limit = min(batch_size, limit - len(all_pages))
    batch = self.confluence.get_all_pages_from_space(
        space=space_key, start=start, limit=fetch_limit,
        expand="ancestors",
    )
    if not batch:
        break
    all_pages.extend(batch)
    if len(batch) < fetch_limit:
        break
    start += len(batch)
pages = all_pages
```

src/mcp_atlassian/servers/confluence.py
```diff
@@ -345,6 +345,66 @@ async def get_page_children(
...
+    tree_data = confluence_fetcher.get_space_page_tree(space_key=space_key, limit=limit)
+
+    result: dict[str, object] = (
+        tree_data if isinstance(tree_data, dict) else {"data": tree_data}
+    )
+
+    # Add pagination metadata
+    total_pages = tree_data.get("total_pages", 0) if isinstance(tree_data, dict) else 0
+    if total_pages >= limit:
+        result["has_more"] = True
```
sooperset: **BLOCKER: `has_more` heuristic is unreliable.**

If the API caps at 200 but `limit=500`, you get 200 pages → `200 >= 500` is `False` → `has_more` reports no more data when it IS truncated. The correct signal would be `_links.next` from the raw API response, but `get_all_pages_from_space()` discards it.

With proper pagination (see other comment), `has_more` can be determined by whether the loop hit the `limit` ceiling.

src/mcp_atlassian/confluence/pages.py
```diff
@@ -704,6 +704,92 @@ def get_page_children(
...
+                limit=limit,
+                expand="ancestors,metadata.properties.position",
```
sooperset: NIT: `metadata.properties.position` in expand is unnecessary — position is read from `extensions.position` (line 754) which is auto-included by the v1 API. The extra expand just increases response payload. `expand="ancestors"` is sufficient.

src/mcp_atlassian/confluence/pages.py
```diff
@@ -704,6 +704,92 @@ def get_page_children(
             logger.debug("Full exception details:", exc_info=True)
             return []
 
+    def get_space_page_tree(
```
sooperset, line 826: NIT: Missing `@handle_auth_errors("Confluence API")` decorator. All recently-added public methods in `PagesMixin` have it (`move_page`, `get_page_version_diff`, `get_page_history`).

src/mcp_atlassian/servers/confluence.py
```diff
@@ -345,6 +345,66 @@ async def get_page_children(
...
+
+    Returns:
+        JSON with space_key, total_pages, and pages array containing
+        {id, title, parent_id, position, depth} for each page.
+        Root pages have parent_id: null and depth: 0.
+    """
+    confluence_fetcher = await get_confluence_fetcher(ctx)
+    tree_data = confluence_fetcher.get_space_page_tree(space_key=space_key, limit=limit)
+
+    result: dict[str, object] = (
```
sooperset: NIT: `isinstance(tree_data, dict)` check and `{"data": tree_data}` fallback are dead code — `get_space_page_tree()` always returns a dict.

sooperset (changes requested):
Thanks for the contribution @Troubladore! The concept is solid — a flat list with `parent_id`/`depth` is much more token-efficient than recursive `get_page_children` calls.

However, there are two correctness issues that need to be addressed before we can merge:

1. **No pagination** — `get_all_pages_from_space()` makes a single API call. Confluence caps responses at ~200-250 results, so spaces with more pages silently return incomplete data.
2. **Unreliable `has_more`** — The `total_pages >= limit` heuristic is wrong when the API cap is lower than the requested limit.

See inline comments for details and suggested fixes.

src/mcp_atlassian/confluence/pages.py
Troubladore: Addressed. Switched to `get_all_pages_from_space_raw()` with a manual pagination loop (`page_size=200`, `_links.next` for continuation), matching the pattern in `confluence/search.py:160-195`.

We chose `_raw` over wrapping `get_all_pages_from_space()` because the library's internal loop has a broken termination condition (`len(results) <= limit`) that exits after one batch when the server cap is below the requested limit — wrapping it would create a double-pagination layer without solving the root issue. Happy to switch to the suggested approach if preferred.

src/mcp_atlassian/servers/confluence.py
Troubladore: Addressed. `has_more` is now computed in the fetcher: `has_more = len(all_pages) >= limit and bool(next_link)`, where `next_link` comes from `_links.next` in the raw API response. The server tool reads it directly instead of guessing.

src/mcp_atlassian/confluence/pages.py
Troubladore: Addressed. Expand changed to `"ancestors"` only.

src/mcp_atlassian/confluence/pages.py
Troubladore: Addressed. Added `@handle_auth_errors("Confluence API")`.

src/mcp_atlassian/servers/confluence.py
Troubladore: Addressed. Simplified to `dict(tree_data)`.

Troubladore:
All 5 items addressed — replied inline to each. Summary of changes:

- **Pagination**: Switched to `get_all_pages_from_space_raw()` with manual pagination loop (`page_size=200`, `_links.next` detection), matching the pattern in `confluence/search.py`. Chose `_raw` over wrapping `get_all_pages_from_space()` to avoid double-pagination and get reliable truncation detection. Happy to adjust if you'd prefer the other approach.
- **`has_more`**: Derived from `_links.next` signal instead of the `total_pages >= limit` heuristic. Also returns `next_start` when truncated so callers can paginate programmatically.
- **Expand**: Removed `metadata.properties.position`, using only `ancestors`.
- **Decorator**: Added `@handle_auth_errors("Confluence API")`.
- **Dead code**: Removed `isinstance(tree_data, dict)` check.

Test coverage includes: multi-batch pagination, API cap smaller than limit, `has_more` in both truncated/exhausted cases, empty batch termination, `limit=1` boundary, `_links` present without `next` key, and `start` offset propagation across batches.

---

**One thing worth flagging:** the current sort order is `(depth, position, title)`, which is breadth-first-ish but doesn't group siblings under the same parent — pages from different parents at the same depth can interleave when they share position values. The `parent_id` field is there so clients can reconstruct proper grouping, but if you'd prefer the response itself to preserve parent-grouped ordering (e.g. sorting by ancestor chain + position), happy to change that.

src/mcp_atlassian/confluence/pages.py
```diff
@@ -822,6 +822,125 @@ def get_page_children(
...
+                "space_key": space_key,
+                "total_pages": len(result_pages),
+                "has_more": has_more,
+                "pages": result_pages,
+            }
+            if has_more:
+                result["next_start"] = start
+            return result
+
+        except Exception as e:
```
sooperset, line 942: The broad `except Exception` catches `HTTPError` (401/403) before `@handle_auth_errors` can convert it to `MCPAtlassianAuthenticationError`. Need to re-raise `HTTPError` explicitly:

```python
        except HTTPError:
            raise  # let @handle_auth_errors decorator handle auth errors
        except Exception as e:
            logger.error(f"Error fetching page tree for space '{space_key}': {e}")
            raise Exception(f"Failed to fetch page tree: {e}") from e
```

This matches the pattern used by other methods in this file (e.g., `delete_page`, `move_page`).

Troubladore: Fixed in 394f803 — added `except HTTPError: raise` before the generic handler, matching the pattern in `get_page_ancestors`, `get_page_history`, `move_page`, and `delete_page`.

sooperset:
Thanks @Troubladore! Great job addressing all the feedback — the pagination loop with `_links.next` detection is solid. Merged.

Issue #486, opened 2025-06-05
[Feature]: Informative Error Messages


TowelDude:
### Is your feature request related to a problem? Please describe.

As remote MCP servers become more prodominent, logging error messages for debugging purposes - creates a bad UX. 

## Current Behavior
Cursor accesses tool -> tool fails (For example, Server URL is incorrect) -> Cursor gets MCP Response of `Error calling tool 'search'` , MCP Server logs `Jira client (fetcher) not available. Ensure server is configured correctly` (which is very informative, and if Cursor would get that he could inform the user about it) 

## Expected Behavior
Informative error messages should propagate to the MCP client. 

## Proposed Fix
That happens because FastMCP catches all Exceptions, logs them, and returns that non-informative error message to the client, to prevent data leakage (see recent PR https://github.com/jlowin/fastmcp/pull/434 ) 
Recently, FastMCP have referenced that issue and suggests raising informative error messages using `fastmcp.exceptions.ToolError(msg="...")`

Since there are a lot of (very informative) error messages raised as `ValueError` and `TypeError` etc, the proposed fix is to gradually start migrating them into an Exception class that inherits `fastmcp.exceptions.ToolError`. 

I would be happy to contribute with a PR to start things off. 

 

### Describe the solution you'd like

..

### Describe alternatives you've considered

..

orinciog:
👍 I also need this feature. 
My usecase is the following:
* When I try to close a JIRA issue, a `TimeSpent` field is required. From jira I receive a good error message, which unfortunately is not passed to MCP Client:

```
HTTP: POST rest/api/2/issue/EC-2398/transitions -> 400 Bad Request
DEBUG - atlassian.rest_client - HTTP: Response text -> {"errorMessages":["Field Time Spent is required."],"errors":{}}
ERROR - mcp-jira - HTTP error during API call: Field Time Spent is required.

requests.exceptions.HTTPError: Field Time Spent is required.
DEBUG - mcp.server.lowlevel.server - Response sent
DEBUG - sse_starlette.sse - ping: b': ping - 2025-06-06 15:10:42.228696+00:00\r\n\r\n'
DEBUG - mcp.server.streamable_http - Closing SSE writer
DEBUG - sse_starlette.sse - chunk: b'event: message\r\ndata: {"jsonrpc":"2.0","id":7,"result":{"content":[{"type":"text","text":"Error calling tool \'transition_issue\'"}],"isError":true}}\r\n\r\n'
```

[...]

PR #1092, opened 2026-03-01
feat: add intelligent error recovery with fuzzy matching and suggestions

Troubladore:
## Description

Builds on #486 (Informative Error Messages). When agents provide slightly wrong input, MCP tools should recover internally rather than failing and forcing retry loops. As [MCPs Are Not Like Other APIs](https://blog.fsck.com/2025/10/19/mcps-are-not-like-other-apis/) argues, tools should be designed "like scripts you're handing to an undertrained ops kid" — forgiving, recoverable, and helpful.

This PR adds three layers of error recovery:

**1. Space key recovery:** When an invalid `space_key` is provided, the tool fetches available spaces and fuzzy-matches. A single confident match (e.g., case mismatch `eruditis` → `ERUDITIS`) is auto-corrected with a note. Multiple matches return suggestions.

**2. Page title recovery:** When `get_page` can't find a title, it searches via CQL for similar titles and auto-corrects or suggests alternatives.

**3. Toolset-aware error messages:** When a filtered-out tool is called, the error explains *why* it's unavailable and *how* to enable it (e.g., "belongs to 'jira_agile' toolset — set TOOLSETS=default,jira_agile").

References #486

## Changes

- New `src/mcp_atlassian/utils/suggestions.py` module:
  - `fuzzy_match()` — case-insensitive `difflib.get_close_matches` + substring fallback
  - `format_suggestions()` — structured error response builder
  - `suggest_spaces()` — fetches spaces and fuzzy-matches
- `servers/confluence.py`:
  - `_try_correct_space_key()` / `_space_key_error_response()` shared helpers
  - Space key recovery in `get_page`, `create_page`, `search`
  - Title recovery in `get_page` via CQL search
- `servers/main.py`:
  - `_call_tool_mcp()` override with toolset-aware error messages
- `utils/toolsets.py`:
  - `find_tool_toolset_from_registry()` helper

## Testing

- [x] 14 tests for suggestion utilities (fuzzy match, format, suggest_spaces)
- [x] 12 Confluence server tests (space key auto-correction/suggestions/no-match, title auto-correction/suggestions/fallthrough, create_page recovery, search filter correction)
- [x] 6 toolset error message tests (unknown tool, write blocked, disabled toolset, config hints)
- [x] 3 toolset registry lookup tests
- [x] All 2556 tests pass

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/servers/confluence.py
+++ b/src/mcp_atlassian/servers/confluence.py
@@ -4,7 +4,7 @@ import base64
 import json
 import logging
 import mimetypes
-from typing import Annotated
+from typing import TYPE_CHECKING, Annotated
 
 from fastmcp import Context, FastMCP
 from mcp.types import BlobResourceContents, EmbeddedResource, ImageContent, TextContent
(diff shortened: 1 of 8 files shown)


sooperset:
Thank you for the comprehensive PR @Troubladore\! While fuzzy matching for error recovery is an interesting concept, we think this adds significant complexity for a feature that MCP clients (Claude, Cursor, etc.) already handle well — they can interpret error messages and retry with corrections. We'd rather keep the server focused on clean error messages and let clients handle recovery logic. Closing for now, but thanks for the creative approach.

PR #1113, opened 2026-03-03
test(jira): cover dynamic Epic Link field use

### LINKED ISSUE #454: [Feature]:

We encountered a problem when using the `jira_link_to_epic` functionality in the MCP server: the custom field for the Epic Link in our Jira instance was not the default (`customfield_10008`), but a different one (`customfield_10001`). This caused the epic linking to silently fail or use the wrong field.

To make the MCP server more robust and compatible with different Jira configurations, I suggest using the `/customFields` API endpoint provided by Jira Data Center. This endpoint returns the correct custom field IDs for all issue types, including the Epic Link field, for the current Jira instance.

By querying `/customFields` at runtime, the MCP server can dynamically determine the correct field to use for epic linking, instead of relying on a hardcoded field ID. This would prevent similar issues for other users whose Jira instances have non-standard field IDs.

**Suggestion:**
- On startup or before linking to an epic, call the `/customFields` endpoint to fetch the correct custom field ID for the Epic Link.
- Use this field ID in the `jira_link_to_epic` implementation.
- Optionally, cache the result for performance.

This change would make the MCP server more flexible and user-friendly across different Jira environments.

Thank you!

_No response_

### Use Case

_No response_

### Additional Context

_No response_

Troubladore:
## Summary

Issue #454 reported that `jira_link_to_epic` could miss Jira Data Center instances whose Epic Link field uses a non-standard ID such as `customfield_10001`. Runtime field discovery is already implemented, so this PR adds focused regression coverage for the complete linking path.

- Returns `customfield_10001` from the mocked Jira field definitions API.
- Exercises the real `get_field_ids_to_epic()` discovery logic through `link_issue_to_epic()`.
- Verifies the discovered field ID is sent after the parent-field strategy is unavailable.
- Avoids source inspection and hardcoded implementation assertions.

## Test plan

- [x] `uv run pytest tests/unit/jira/test_epics.py -q` — 26 passed
- [x] `uv run pytest tests/integration/ --integration -q` — 8 passed, 15 skipped
- [x] `uv run pytest -xq` — 3205 passed, 229 skipped
- [x] `uv run pre-commit run --all-files` — all hooks passed, including Ruff and mypy

Closes #454

### DIFF
--- a/tests/unit/jira/test_epics.py
+++ b/tests/unit/jira/test_epics.py
@@ -1,5 +1,6 @@
 """Tests for the Jira Epics mixin."""
 
+import inspect
 from unittest.mock import MagicMock, call
 
 import pytest
@@ -48,7 +49,8 @@ class TestEpicsMixin:
             "fields": {
                 "issuetype": {"name": "Epic"},
                 "summary": "Test Epic",
-                "customfield_10011": "Epic Name Value",  # This should be discovered as epic_name
+                # Should be discovered as epic_name
+                "customfield_10011": "Epic Name Value",
             },
         }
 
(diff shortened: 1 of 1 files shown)


Troubladore:

sooperset:
Thanks @Troubladore, the focused regression test covers the full field-discovery path and proves `customfield_10001` is used when the parent strategy fails. That's exactly the coverage this Jira DC case needed.

Issue #904, opened 2026-02-09
[Bug]: markdown_to_jira converts Jira numbered list syntax to headings when processing mixed format


hteichmann-strato:
# [Bug]: markdown_to_jira converts Jira numbered list syntax to headings when processing mixed format

## Prerequisites

- [x] I have searched the existing issues - this is NOT a duplicate of #786
- [x] I have checked the README

## Bug Description

The `markdown_to_jira()` converter incorrectly transforms valid Jira numbered list syntax (`# Item`) into heading syntax (`h1. Item`) when the input contains mixed markdown and Jira wiki markup, or when already-converted content is re-processed.

**Impact:** Numbered list items render as large bold headings instead of numbered lists in Jira.

**Difference from #786:** PR #827 fixed `#item` (no space) → `h1.` conversion, but didn't handle `# Item` (with space), which is the correct Jira numbered list syntax. Since both Jira numbered lists and markdown headings use `# ` (hash + space), they're indistinguishable to the current regex.

## Reproduction

**Test case 1: Mixed markdown + Jira wiki markup**
```python
input = """h3. Next Steps

1. Wait for mirrors
1. Apply patch"""

result = preprocessor.markdown_to_jira(input)
# Expected: h3. Next Steps\n\n# Wait for mirrors\n# Apply patch
# Actual:   h3. Next Steps\n\nh1. Wait for mirrors\nh1. Apply patch
```

**Test case 2: Already-converted Jira wiki markup**
```python
input = """h3. Title

# First item
# Second item"""

result = preprocessor.markdown_to_jira(input)
# Expected: (unchanged - already Jira format)
# Actual:   h3. Title\n\nh1. First item\nh1. Second item
```

## When This Occurs

- AI assistants generate mixed format content (Jira wiki markup headings + markdown lists)
- Users have templates with Jira wiki markup and add markdown content
- Content is copied from Jira and re-submitted with modifications
- Accidental double-processing of already-converted content

## Real-World Example

JIRA comment that triggered this issue:

**What was generated:**
```
h3. Next Steps (Tomorrow)

1. Wait for CentOS Stream 9 mirrors to stabilize
1. If issue persists, apply Dockerfile patch
1. Retry build once infrastructure stabilizes
```

**What rendered in Jira:**
```
[Large heading] Next Steps (Tomorrow)

[Huge heading] Wait for CentOS Stream 9 mirrors to stabilize
[Huge heading] If issue persists, apply Dockerfile patch
[...]

PR #1127, opened 2026-03-03
fix(preprocessing): preserve Jira ordered list items when Jira wiki markers present

hteichmann-strato:
## Summary

Fixes the remaining write-path half of #904.

`markdown_to_jira()` converts `# text` to `h1. text` for Markdown headings, but Jira wiki markup uses the same `# ` syntax for ordered list items. When mixed or already-converted Jira wiki content contains an unambiguous Jira marker, ordered list items must remain lists.

The converter now detects Jira wiki headings and complete Jira block-macro markers after protecting Markdown code spans. It skips only the ambiguous hash-heading conversion; the remaining Markdown-to-Jira conversions still run for mixed input.

## Scope

This update keeps the original ordered-list fix focused. The unrelated formatting conversions from the follow-up commit were removed while rebasing onto current `main`.

## Changes

- Detect `h1.`-`h6.` headings and complete `{code}`, `{noformat}`, and `{panel}` markers.
- Preserve `# item` and nested `## item` Jira ordered lists when those markers are present.
- Continue converting Markdown numbered lists in mixed Jira/Markdown input.
- Avoid false positives from macro-like words and Jira marker examples inside fenced Markdown code.
- Add regression coverage for mixed input, nested lists, valid macros, lookalikes, fenced code, pure Markdown, and the unavoidable bare `# Item` ambiguity.

## Verification

- `uv run pre-commit run --all-files` — passed.
- `uv run pytest tests/unit/preprocessing/test_preprocessing.py -q` — 127 passed.
- `uv run pytest tests/unit/ -q` — 3,034 passed, 5 skipped.
- Jira DC E2E subset — 42 passed, 1 skipped.
- Direct Jira DC create/read/delete regression — stored `# First item` / `# Second item`, with no `h1.` corruption.
- Full Cloud E2E — all Jira write checks passed. Two transient media/DNS failures passed on rerun; one unrelated Confluence search-model failure remains outside this Jira-only diff.

### DIFF
--- a/src/mcp_atlassian/preprocessing/jira.py
+++ b/src/mcp_atlassian/preprocessing/jira.py
@@ -107,6 +107,13 @@ class JiraPreprocessor(BasePreprocessor):
         "cmake": "bash",
     }
 
+    # Regex that detects Jira wiki markup markers unambiguous enough to conclude
+    # the input is Jira wiki format rather than Markdown.  Used to skip the
+    # "# heading" → "h1. heading" conversion when the # is a Jira ordered list.
+    _JIRA_WIKI_MARKER_RE = re.compile(
+        r"(?:^h[1-6]\.\s|^\{code|^\{noformat|^\{panel)", re.MULTILINE
+    )
+
     def __init__(
         self, base_url: str = "", disable_translation: bool = False, **kwargs: Any
     ) -> None:
@@ -471,14 +478,20 @@ class JiraPreprocessor(BasePreprocessor):
             flags=re.MULTILINE,
         )
 
-        # Headers with # prefix - require space after #
-        # to distinguish from Jira lists (issue #786)
-        output = re.sub(
-            r"^([#]+) (.*)$",
-            lambda match: f"h{len(match.group(1))}. " + match.group(2),
-            output,
-            flags=re.MULTILINE,
-        )
+        # Headers with # prefix - only convert when input looks like Markdown, not
+        # Jira wiki markup.  Jira wiki uses "h1. Heading" for headings and "# Item"
+        # for ordered lists.  If the (code-stripped) text already contains a Jira
+        # heading marker (h1.-h6.) or a Jira macro opening brace, the caller is
+        # sending Jira wiki markup and "# Item" should be treated as an ordered-list
+        # entry, not converted to "h1. Item".
+        # Fixes issue #904: # item must not become h1. when Jira wiki markers present
+        if not self._JIRA_WIKI_MARKER_RE.search(output):
+            output = re.sub(
+                r"^([#]+) (.*)$",
+                lambda match: f"h{len(match.group(1))}. " + match.group(2),
+                output,
+                flags=re.MULTILINE,
+            )
 
         # Bold and italic - skip lines starting with
         # asterisks+space (Jira list syntax, issue #786)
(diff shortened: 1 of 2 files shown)


hteichmann-strato:
While working on the test suite I expanded coverage for `markdown_to_jira` and also fixed several missing conversions along the way.

**Root cause fix (this PR):**
- Jira ordered list items (`# item`) were being corrupted to `h1. item` headings when the input already contained Jira wiki markup - because `markdown_to_jira()` ran unconditionally. Fixed via Option A: detect Jira wiki markers at function entry and pass input through unchanged.

**Additional conversions added:**
- Bold+italic (`***x***`, `___x___`, `__*x*__`, `**_x_**`) → `*_x_*`
- Blockquotes (`> text`, multi-paragraph with bare `>`) → `{quote}..{quote}`
- Horizontal rule (`---`, `***`, `___`) → `----`
- Fixed setext heading regex that was corrupting standalone `---` into `h2.`

**Test scheme introduced:** `passthrough_` (Jira wiki in/out), `converts_` (Markdown in, Jira wiki out), `ambiguous_` (known limitation) - makes it easier to see what's covered.

---

**Broader concern I'd like to raise:**

The current dual-support model (auto-detecting whether input is Jira wiki or Markdown) is inherently fragile. The detection relies on a small set of markers (`h[1-6].`, `{code`, `{noformat`, `{panel`) - but there are inputs that will silently fall through to the wrong path, either getting converted when they shouldn't or passed through when conversion was expected.

The tool spec says input should be Markdown. If we commit to that, we can drop the detection entirely and always convert - simpler, predictable, no edge cases. The tradeoff is breaking callers who currently pass Jira wiki markup directly.

Worth discussing: should `markdown_to_jira` be Markdown-only, or do we want to keep the auto-detection and accept the edge case risk?

hteichmann-strato:
@sooperset ping, any update here from your side or something is missing?

hteichmann-strato:
@sooperset would you pls take a look here? the formatting on ordered-lists on non-cloud JIRA Instances is still broken with the mcp-atlassian MCP.

src/mcp_atlassian/preprocessing/jira.py
```diff
@@ -471,14 +479,20 @@ def _md_inline_to_jira(
...
-            flags=re.MULTILINE,
-        )
+        # Headers with # prefix - only convert when input looks like Markdown, not
+        # Jira wiki markup.  Jira wiki uses "h1. Heading" for headings and "# Item"
+        # for ordered lists.  If the (code-stripped) text already contains a Jira
+        # heading marker (h1.-h6.) or a Jira macro opening brace, the caller is
+        # sending Jira wiki markup and "# Item" should be treated as an ordered-list
+        # entry, not converted to "h1. Item".
+        # Fixes issue #904: # item must not become h1. when Jira wiki markers present
+        if not self._JIRA_WIKI_MARKER_RE.search(output):
```
sooperset, line 489: **BLOCKER:** This searches `output` after the Setext conversion above has already generated `h1.` or `h2.` markers. For pure Markdown such as `Title\n=====\n\n# Section`, the generated `h1.` makes this condition skip ATX conversion, so `# Section` is sent to Jira as an ordered-list item. Required for merge: determine whether Jira markers were present in the protected input before any heading transformation, and add this mixed-heading case as a regression test.

sooperset (changes requested):
Thanks for the contribution @hteichmann-strato! Keeping this focused on the ordered-list corruption and covering marker lookalikes was the right direction.

1. The marker check must run before Setext heading conversion so generated Jira syntax can't alter how later Markdown headings are classified.
2. Add a regression test mixing Setext and ATX headings.

PR #1128, opened 2026-03-03
feat(auth): add SSO cookie support for Jira and Confluence

jkiddo:
## Summary

- Add `JIRA_COOKIE` and `CONFLUENCE_COOKIE` environment settings.
- Attach the configured raw cookie to the corresponding HTTP session for all auth modes.
- Document the cookie format and secret-handling requirements.
- Cover both environment parsing and session application for Jira and Confluence.
- Synchronize the contributor branch with current `main`.

## Motivation

Some SSO gateways and reverse proxies require a session cookie in addition to the normal Jira or Confluence credentials. Dedicated cookie settings make that deployment pattern available without embedding the value in general custom-header configuration.

## Verification

- Focused Jira and Confluence config/client tests: 81 passed
- Full unit suite: 3,067 passed, 5 skipped
- Integration suite: 8 passed, 15 service-specific cases skipped
- Pre-commit: Ruff formatting, Ruff lint, and strict mypy passed
- Cloud e2e: 82 passed, 15 optional OAuth cases skipped
- Data Center e2e: 94 passed, 1 skipped

### DIFF
--- a/src/mcp_atlassian/confluence/config.py
+++ b/src/mcp_atlassian/confluence/config.py
@@ -40,6 +40,7 @@ class ConfluenceConfig:
     client_key: str | None = None  # Client private key file path (.pem)
     client_key_password: str | None = None  # Password for encrypted private key
     timeout: int = 75  # Connection timeout in seconds
+    cookie: str | None = None  # Raw cookie string to attach to all requests
 
     @property
     def is_cloud(self) -> bool:
@@ -187,6 +188,9 @@ class ConfluenceConfig:
         ):
             timeout = int(os.getenv("CONFLUENCE_TIMEOUT", "75"))
 
+        # Cookie configuration
+        cookie = os.getenv("CONFLUENCE_COOKIE")
+
         return cls(
             url=url or "",
             auth_type=auth_type,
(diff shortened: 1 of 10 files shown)


Troubladore:

src/mcp_atlassian/confluence/config.py
```diff
@@ -40,6 +40,7 @@ class ConfluenceConfig:
     client_key: str | None = None  # Client private key file path (.pem)
     client_key_password: str | None = None  # Password for encrypted private key
     timeout: int = 75  # Connection timeout in seconds
+    cookie: str | None = None  # Raw cookie string to attach to all requests
```
sooperset, line 43: **BLOCKER:** Required for merge: move `cookie` after `attachment_download_use_v1`. v0.22.0 already exposes `attachment_download_use_v1` as the last positional argument. Inserting `cookie` before it silently rebinds an existing `ConfluenceConfig(..., 75, True)` call to `cookie=True`, leaves the attachment setting at `None`, and makes `requests` raise `InvalidHeader` when it prepares a request. Please preserve the old parameter order and add a regression test for that positional call.

sooperset (changes requested):
Thanks for the contribution @jkiddo! The session-level Jira and Confluence tests close the original HTTP-application coverage gap well.

1. Preserve the v0.22.0 positional constructor contract for `ConfluenceConfig`.

PR #1133, opened 2026-03-04
fix(confluence): add ancestors to expand params in get_page_content (closes #1131)

### LINKED ISSUE #1131: [Feature]: Add parentId field to confluence_get_page response

 Description
The `confluence_get_page` function does not return the `parentId` field in its response, which makes it difficult to determine the parent page of a Confluence page.

 Problem
When querying a Confluence page via `confluence_get_page`, the returned metadata does not include information about the page's parent. This makes it impossible to:
- Know which page is the parent of a given page
- Build the page hierarchy programmatically
- Move pages to the correct parent

 Expected Behavior
The response should include a `parentId` field that indicates the ID of the parent page (if any).

 Workaround
Currently, users have to:
1. Search for keywords to find potential parent pages
2. Use `get_page_children` to iterate through candidate pages
3. Ask the user directly
These workarounds are inefficient and error-prone.

 Additional Request
It would also be helpful to have a new tool like `confluence_get_page_ancestors` that returns the full ancestor chain of a page (parent, grandparent, etc.).
 
Environment
- MCP Server: mcp-atlassian
- Confluence: Self-hosted (Server/Data Center)

 Example Use Case
I need to create a page to the same parent as another page, but I cannot determine the parent ID of the reference page through the API.

 Expected Behavior
The response  `confluence_get_page` function should include a `parentId` field that indicates the ID of the parent page (if any).

 Additional Request
It would also be helpful to have a new tool like `confluence_get_page_ancestors` that returns the full ancestor chain of a page (parent, grandparent, etc.).

_No response_

### Use Case

_No response_

### Additional Context

_No response_

Troubladore:
## Summary

- Add `ancestors` to the expand parameter in `get_page_content()` so `confluence_get_page` responses include the ancestor chain (parent, grandparent, etc.)
- Same fix applied to the version-history code path (`get_page_by_version`)
- The `ConfluencePage` model already has an `ancestors` field and `to_simplified_dict()` already formats it — only the expand param was missing

## Test plan

- [x] Unit tests updated to assert new expand string
- [x] E2E test added: creates parent→child hierarchy, fetches child, verifies ancestors populated with parent ID
- [ ] Verify on Cloud instance
- [ ] Verify on Server/DC instance

Closes #1131

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/mcp_atlassian/confluence/pages.py
+++ b/src/mcp_atlassian/confluence/pages.py
@@ -63,7 +63,7 @@ class PagesMixin(ConfluenceClient):
                 )
                 page = v2_adapter.get_page(
                     page_id=page_id,
-                    expand="body.storage,version,space,children.attachment",
+                    expand="body.storage,version,space,children.attachment,ancestors",
                 )
             else:
                 logger.debug(
@@ -72,7 +72,7 @@ class PagesMixin(ConfluenceClient):
                 )
                 page = self.confluence.get_page_by_id(
                     page_id=page_id,
-                    expand="body.storage,version,space,children.attachment",
+                    expand="body.storage,version,space,children.attachment,ancestors",
                 )
 
             # Check if API returned an error string
@@ -1031,7 +1031,7 @@ class PagesMixin(ConfluenceClient):
                 page = v2_adapter.get_page_by_version(
                     page_id=page_id,
                     version=version,
-                    expand="body.storage,version,space,children.attachment",
+                    expand="body.storage,version,space,children.attachment,ancestors",
                 )
             else:
                 logger.debug(
@@ -1043,7 +1043,7 @@ class PagesMixin(ConfluenceClient):
                     page_id=page_id,
                     status="historical",
                     version=version,
-                    expand="body.storage,version,space,children.attachment",
+                    expand="body.storage,version,space,children.attachment,ancestors",
                 )
 
             if isinstance(page, str):
--- a/tests/e2e/cloud/test_confluence_cloud_operations.py
+++ b/tests/e2e/cloud/test_confluence_cloud_operations.py
@@ -137,3 +137,44 @@ class TestConfluenceCloudComments:
 
         comments = confluence_fetcher.get_page_comments(page.id)
         assert len(comments) > 0
+
+
+class TestConfluencePageAncestors:
+    """get_page_content should populate the ancestors field.
+
+    Regression for https://github.com/sooperset/mcp-atlassian/issues/1131
+    Root cause: 'ancestors' not included in expand params in get_page_content().
+    """
+
+    def test_child_page_has_ancestors(
+        self,
+        confluence_fetcher: ConfluenceFetcher,
+        cloud_instance: CloudInstanceInfo,
+        resource_tracker: CloudResourceTracker,
+    ) -> None:
+        uid = uuid.uuid4().hex[:8]
+        parent = confluence_fetcher.create_page(
+            space_key=cloud_instance.space_key,
+            title=f"Ancestor Parent {uid}",
+            body="<p>Parent page.</p>",
+            is_markdown=False,
+        )
+        resource_tracker.add_confluence_page(parent.id)
+
+        child = confluence_fetcher.create_page(
+            space_key=cloud_instance.space_key,
+            title=f"Ancestor Child {uid}",
+            body="<p>Child page.</p>",
+            is_markdown=False,
+            parent_id=parent.id,
+        )
+        resource_tracker.add_confluence_page(child.id)
+
+        fetched = confluence_fetcher.get_page_content(child.id)
+        assert len(fetched.ancestors) > 0, (
+            "ancestors is empty — 'ancestors' missing from expand params"
+        )
+        parent_ids = [a.get("id") for a in fetched.ancestors]
+        assert parent.id in parent_ids, (
+            f"parent {parent.id} not in ancestors: {parent_ids}"
+        )
--- a/tests/e2e/cloud/test_confluence_cloud_operations.py
+++ b/tests/e2e/cloud/test_confluence_cloud_operations.py
@@ -178,3 +178,17 @@ class TestConfluencePageAncestors:
         assert parent.id in parent_ids, (
             f"parent {parent.id} not in ancestors: {parent_ids}"
         )
+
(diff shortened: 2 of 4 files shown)


Troubladore:

sooperset:
Maintainer sync: merged current `main` into this branch (no conflicts) to clear the stale/behind state. All required checks are green on the updated head `8eaa117`: pytest on Python 3.10–3.13, lint, and build-and-push.

Note: the added `cloud_e2e` ancestors regression test was not executed here (requires live Confluence Cloud credentials); it is auto-skipped in standard CI. The change is a low-risk expand-param addition and is fully covered by unit tests exercising the API-response → model → `to_simplified_dict()` path. Ready for review.

sooperset:
The v1/basic side of this PR is in good shape: adding `ancestors` to the v1 expand string plus the model-to-`to_simplified_dict()` test proves that path.

I can't merge it yet because Cloud OAuth still won't return ancestors. In this repo, Cloud OAuth uses `ConfluenceV2Adapter`; that adapter accepts `expand` only as a compatibility argument, sends only `body-format=storage`, and never maps `ancestors` into the v1-compatible response.

Required for merge:

1. Make `ConfluenceV2Adapter.get_page()` populate `ancestors` for Cloud OAuth. If v2 needs a separate call for ancestors, call it and map the result into the existing v1-compatible shape.
2. Cover the OAuth/v2 path with a test that fails today. A mock assertion that `expand` was passed isn't enough because the adapter currently ignores it.
3. Either make `get_page_by_version()` support ancestors too, or narrow the version-history part of the PR.
4. Clarify the e2e coverage. The new Cloud e2e uses the basic-auth `confluence_fetcher`, so it doesn't cover the OAuth/v2 path.

Happy to merge once this covers the OAuth path and checks are rerun on the latest `main`.

sooperset:
The new DC e2e assertion is useful, and it confirms the Server/DC path for this fix.

I still can't merge this because the Cloud OAuth path is unchanged. Cloud OAuth uses `ConfluenceV2Adapter`, and that adapter still treats `expand` as a compatibility argument only. It sends `body-format=storage` and doesn't map `ancestors` into the v1-compatible response.

Required for merge:

1. Make `ConfluenceV2Adapter.get_page()` populate `ancestors` for Cloud OAuth.
2. If v2 needs a separate ancestors call, call it and map the result into the existing v1-compatible shape.
3. Either make `get_page_by_version()` support ancestors too, or narrow that part of the PR.
4. Add a test that fails against the current v2 adapter behavior. Checking that `expand` was passed to a mock isn't enough because the real adapter ignores it.

Happy to merge once the OAuth/v2 path is covered and the checks are rerun.

src/mcp_atlassian/confluence/pages.py
```diff
@@ -64,7 +64,7 @@ def get_page_content(
                 )
                 page = v2_adapter.get_page(
                     page_id=page_id,
-                    expand="body.storage,version,space,children.attachment",
+                    expand="body.storage,version,space,children.attachment,ancestors",
```
sooperset, line 67: **BLOCKER:** This passes `ancestors` into the OAuth path, but the real `ConfluenceV2Adapter.get_page()` treats `expand` as compatibility-only. It still calls v2 pages with only `body-format=storage` and `_convert_v2_to_v1_format()` doesn't add `ancestors`, so OAuth users still get an empty parent list. Required for merge: fetch and map ancestors in the adapter, or narrow the PR away from OAuth.

src/mcp_atlassian/confluence/pages.py
```diff
@@ -1176,7 +1176,7 @@ def get_page_history(
                 page = v2_adapter.get_page_by_version(
                     page_id=page_id,
                     version=version,
-                    expand="body.storage,version,space,children.attachment",
+                    expand="body.storage,version,space,children.attachment,ancestors",
```
sooperset, line 1179: **IMPORTANT:** The version-history path has the same problem. `ConfluenceV2Adapter.get_page_by_version()` accepts `expand`, but it only sends `body-format=storage` and doesn't map ancestors. Please either implement ancestors for this adapter method too, or remove the history-path claim from this PR.

tests/e2e/cloud/test_confluence_cloud_operations.py
```diff
@@ -191,3 +191,58 @@ def test_add_and_get_comments(
...
+class TestConfluencePageAncestors:
+    """get_page_content should populate the ancestors field.
+
+    Regression for https://github.com/sooperset/mcp-atlassian/issues/1131
+    Root cause: 'ancestors' not included in expand params in get_page_content().
+    """
+
+    def test_child_page_has_ancestors(
+        self,
+        confluence_fetcher: ConfluenceFetcher,
```
sooperset, line 205: **IMPORTANT:** This Cloud e2e uses the default `confluence_fetcher` fixture, which is basic auth. It confirms Cloud basic auth, but it doesn't exercise the OAuth/v2 adapter that still drops ancestors.

tests/unit/confluence/test_pages.py
```diff
@@ -1453,7 +1491,8 @@ def test_get_page_content_oauth_uses_v2_api(self, oauth_pages_mixin):
 
             # Assert that v2 API was used instead of v1
             mock_v2_adapter.get_page.assert_called_once_with(
-                page_id=page_id, expand="body.storage,version,space,children.attachment"
+                page_id=page_id,
+                expand="body.storage,version,space,children.attachment,ancestors",
```
sooperset, line 1495: **IMPORTANT:** This only proves `PagesMixin` passed a string to a mock. The bug is below this layer: the real v2 adapter ignores the string. Please add coverage against `ConfluenceV2Adapter.get_page()` or an OAuth-path fixture where `ConfluencePage.ancestors` and `to_simplified_dict()` are populated from the adapter output.

sooperset (changes requested):
Thanks for the contribution @Troubladore! The v1/basic expand change is small, and the model-to-`to_simplified_dict()` regression proves that path.

1. Cloud OAuth still won't return ancestors because the v2 adapter doesn't use or map the new `expand` value.
2. The current OAuth and Cloud tests don't cover that adapter behavior, so the required `unit+cloud` evidence is incomplete.

PR #1140, opened 2026-03-06
fix(jira): allow null values to clear fields in update_issue

legard:
## Description

When calling `jira_update_issue` with `null` for a field (e.g., `customfield_10105: null` to clear Sprint), `_process_additional_fields` passes the value through `_format_field_value_for_write`, which returns `None` for both formatting errors **and** intentional `null` values. The subsequent `if formatted_value is not None` guard then silently drops the update.

This means there is currently no way to clear a custom or standard field via `jira_update_issue` — the `null` is indistinguishable from a formatting failure.

**Workaround before this fix:** Direct REST API call `PUT /rest/api/2/issue/{key}` with `{"fields": {"customfield_10105": null}}` works (HTTP 204), confirming this is a client-side filtering issue.

## Changes

- Add an early-return guard in `_process_additional_fields` that passes `None` directly to the fields dict, bypassing `_format_field_value_for_write` (`src/mcp_atlassian/jira/issues.py`)
- Document `None` clearing behavior in `update_issue` docstring
- Add 4 unit tests covering:
  - `None` for custom field IDs (e.g., `customfield_10013=None`)
  - `None` for named fields (e.g., `priority=None`)
  - Full `update_issue()` → API call path with `None` kwargs
  - Regression: invalid (non-None) values are still filtered out

This is consistent with the existing `assignee=None` pattern already in `update_issue` (line 1047–1048).

## Testing

- [x] All 46 tests pass (`uv run pytest tests/unit/jira/test_issues.py`)
- [x] Manual integration test: cleared Sprint (`customfield_10105`) on a real Jira Server issue via MCP tool call, confirmed field was set to `null`

### Test results

```
$ uv run pytest tests/unit/jira/test_issues.py -v
============================== 46 passed in 0.13s ==============================
```

```
$ uv run pre-commit run --files src/mcp_atlassian/jira/issues.py tests/unit/jira/test_issues.py
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check for added large files..............................................Passed
debug statements (python)................................................Passed
ruff-format..............................................................Passed
ruff.....................................................................Passed
mypy.....................................................................Passed
```

## Checklist

- [x] Code follows project style guidelines (linting passes)
- [x] All tests pass locally
- [x] Documentation updated (docstring)

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -986,6 +986,14 @@ class IssuesMixin(
                 logger.debug(f"Identified field '{key}' as standard system field ID.")
 
             if api_field_id:
+                # Allow None values to pass through for clearing fields
+                if value is None:
+                    fields[api_field_id] = None
+                    logger.debug(
+                        f"Setting field '{api_field_id}' to None from kwarg '{key}' (clearing field)."
+                    )
+                    continue
+
                 # Get the full field definition for formatting context if needed
                 field_definition = self.get_field_by_id(
                     api_field_id
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -1058,6 +1058,7 @@ class IssuesMixin(
                 - assignee: New assignee for the issue
                 - parent: Parent issue key (str or {"key": "..."} dict)
                 - epicKey/epic_link/epicLink: Epic link alias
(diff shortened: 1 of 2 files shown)


Troubladore:

sooperset:
Thanks Ruslan, this is a focused fix for preserving null updates when Jira fields need to be cleared. The regression coverage for both direct custom field IDs and named fields made the change straightforward to review.

PR #1152, opened 2026-03-10
feat(jira): add dedicated assign_issue tool

JeanBaptisteRenard:
## Summary

Add a new `jira_assign_issue` MCP tool that uses the dedicated Jira assignment endpoint (`PUT /rest/api/3/issue/{key}/assignee`) instead of relying on the fields update in `update_issue`.

### Motivation

On some Jira Cloud configurations, setting the assignee field via `update_issue` (which uses the generic fields update endpoint) is **silently ignored** — the API returns 204 but the assignee doesn't change. The dedicated assignment endpoint works reliably in all configurations.

### Changes

- **`src/mcp_atlassian/jira/issues.py`**: Add `assign_issue()` method to `IssuesMixin`
- **`src/mcp_atlassian/servers/jira.py`**: Add `jira_assign_issue` tool with `@check_write_access` decorator
- **`tests/unit/jira/test_issues.py`**: Unit tests for mixin (assign, unassign via None, unassign via empty string, error handling)
- **`tests/unit/servers/test_jira_server.py`**: Server-layer tests (assign, unassign, error propagation)

### Features

- Accepts email, display name, or account ID as user identifier
- Supports unassignment (pass `null` or empty string)
- Resolves user identifiers via `_get_account_id` (Cloud/Server compatible)
- Returns the updated issue after assignment
- Tagged with `jira`, `write`, `toolset:jira_issues`
- Respects `READ_ONLY_MODE` via `@check_write_access`

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -1163,6 +1163,48 @@ class IssuesMixin(
             logger.error(f"Error updating issue {issue_key}: {error_msg}")
             raise ValueError(f"Failed to update issue {issue_key}: {error_msg}") from e
 
+    def assign_issue(
+        self,
+        issue_key: str,
+        assignee: str | None,
+    ) -> JiraIssue:
+        """
+        Assign a Jira issue to a user using the dedicated assignment endpoint.
+
+        Unlike update_issue (which sets assignee via the fields update and is
+        silently ignored by some Jira configurations), this method calls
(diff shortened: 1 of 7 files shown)


Troubladore:

sooperset:
Merged this because the dedicated assignee endpoint closes the silent no-op path cleanly, and the null/unassign handling is covered. The MCP-level Cloud/DC tests are the right guardrail for this tool.

PR #1153, opened 2026-03-11
fix(jira): use correct case-sensitive applicationType values for dev-status API

sfc-gh-hdibani:
## Summary

The Atlassian dev-status API (`/rest/dev-status/1.0/issue/detail`) requires the `applicationType` query parameter to use the **exact casing** registered by each DVCS connector plugin. For GitHub integrations the correct value is `GitHub` (capital G), not `github`.

**Current behaviour:**
- Passing `applicationType=github` (lowercase) returns a **500 Internal Server Error** on Atlassian Cloud.
- When no `application_type` is specified, the fallback loop in `DevelopmentMixin` tries lowercase `"github"`, catches the resulting exception, and returns **empty results** — silently hiding all linked PRs/branches/commits.

**After this fix:**
- The fallback list uses the correct casing: `["stash", "bitbucket", "GitHub", "GitLab"]`
- Docstrings and MCP tool `Field` descriptions now show `'GitHub'` (not `'github'`) and note that values are case-sensitive, so LLMs and users pass the right value on the first attempt.

## Changes

- `src/mcp_atlassian/jira/development.py` — fix fallback `app_types` list + update docstrings
- `src/mcp_atlassian/servers/jira.py` — update `Field(description=...)` for `application_type` in both `get_issue_development_info` and `get_issues_development_info`

## How I found this

While building a support agent that calls `jira_get_issue_development_info` via MCP, the tool consistently returned empty results. Tracing the HTTP calls revealed the 500 from the dev-status endpoint when `applicationType=github` was sent. Changing to `applicationType=GitHub` immediately returned the expected PR/branch data.

## Test plan

- [x] Verified `applicationType=GitHub` returns linked PRs on Atlassian Cloud (tested against a real Jira issue)
- [x] Confirmed `applicationType=github` (lowercase) returns 500
- [x] Confirmed omitting `applicationType` now falls back to the correct casing and returns results

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/mcp_atlassian/jira/development.py
+++ b/src/mcp_atlassian/jira/development.py
@@ -26,7 +26,9 @@ class DevelopmentMixin(JiraClient):
         Args:
             issue_key: The issue key (e.g., PROJECT-123)
             application_type: Filter by application type
-                (e.g., 'stash', 'github', 'bitbucket').
+                (e.g., 'stash', 'GitHub', 'bitbucket').
+                Values are case-sensitive (the dev-status API
+                returns empty results or 500 errors on mismatch).
                 If None, tries common application types.
             data_type: Filter by data type
                 (e.g., 'pullrequest', 'branch', 'repository').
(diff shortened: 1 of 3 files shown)


Troubladore:

sooperset:
Merged, Hassan. The applicationType casing fix is a good catch, especially because the old fallback could hide linked GitHub development data instead of failing loudly. The added regression test around auto-discovery order is exactly the right coverage for this.

PR #1156, opened 2026-03-11
feat(jira): add human-readable custom field keys

### LINKED ISSUE #1154: [Feature]: Use display names as keys for custom fields instead of customfield_XXXXX IDs

Summary

Custom fields are returned (and must be referenced) using opaque internal IDs like `customfield_12345`. This makes it hard for both humans and AI agents to work with custom fields without first discovering what each ID means.

The infrastructure for display-name mapping **already exists** in the codebase (`FieldsMixin._generate_field_map()` builds a `name → id` map), but it isn't used on the read path to present custom fields by their human-readable names.

## Current behavior

When retrieving an issue with `fields="*all"`, custom fields are returned keyed by internal IDs:

```json
{
  "customfield_10028": {"value": 5, "name": "Story Points"},
  "customfield_12345": {"value": "Critical", "name": "Severity"},
  "customfield_12400": {"value": "Passed", "name": "Test Status"}
}
```

## Describe the solution you'd like

Custom fields should be keyed by their human-readable display name instead of opaque `customfield_XXXXX` IDs in all MCP tool responses, and display names should be accepted as input on write operations.

**Read path**: When `jira_get_issue` or `jira_search` returns custom fields, the output should look like:

```json
{
  "Story Points": {"value": 5, "field_id": "customfield_10028"},
  "Severity": {"value": "Critical", "field_id": "customfield_12345"}
}
```

instead of the current:

```json
{
  "customfield_10028": {"value": 5, "name": "Story Points"},
  "customfield_12345": {"value": "Critical", "name": "Severity"}
}
```

**Write path**: `jira_update_issue` and `jira_create_issue` should accept display names in `fields` / `additional_fields`:

```json
{"Severity": "Critical", "Story Points": 5}
```

instead of requiring:

```json
{"customfield_12345": "Critical", "customfield_10028": 5}
```

The codebase already has the infrastructure for this. `FieldsMixin._generate_field_map()` builds a `name → id` lookup from Jira's `/rest/api/latest/field` endpoint, and `_process_additional_fields()` already resolves names to IDs on writes. The main changes needed are:

1. `JiraIssue.to_simplified_dict()` - use the stored display name as the output key, keep `field_id` inside for reference
2. `JiraIssue.from_api_response()` - already stores the name from the `names` expand, just needs to use it as the primary key

Backward compatibility is preserved by including `field_id` in each entry so users using raw IDs can still look them up.

1. **Keep current behavior and rely on `jira_search_fields`**: Users/agents call `jira_search_fields` first to discover field IDs, then use those IDs in subsequent calls. This works but adds an extra round-trip for every interaction and forces agents to maintain a mental mapping between names and IDs.

### Use Case

 AI agents and automation tools that interact with Jira through MCP.

### Additional Context

_No response_

faizbawa:
## Description

Adds an opt-in `use_display_names` parameter to `jira_get_issue` and `jira_search`.
When enabled, custom field keys use human-readable names such as `Story Points`
instead of opaque IDs such as `customfield_10243`. The tools automatically request
the Jira `names` expansion.

If a display name conflicts with a standard output key, another custom field, or a
requested `include` section, the custom field keeps its raw ID so no value is silently
lost.

Fixes #1154.

## Changes

- Add display-name lookup and serialization helpers to `JiraIssue` and `JiraSearchResult`.
- Preserve and combine field-name maps across Jira Cloud search pages.
- Preserve `fields="*all"` in `jira_get_issue` so custom fields reach the model.
- Add `use_display_names: bool = false` to `jira_get_issue` and `jira_search`.
- Merge the automatic `names` expansion with caller-provided expansions.
- Reserve requested `include` output keys during display-name selection.
- Add collision, client, server, Cloud E2E, and Data Center E2E regression coverage.
- Document the opt-in output and collision fallback.

## Verification

- `uv run pre-commit run --all-files` — passed (Ruff, formatting, mypy).
- Affected unit suites — 310 passed.
- `uv run pytest tests/integration/ --integration -q` — 8 passed, 15 skipped.
- Full Cloud E2E — 84 passed, 15 skipped.
- Full Data Center E2E — 96 passed, 100 Cloud-only tests skipped.

### DIFF
--- a/src/mcp_atlassian/servers/jira.py
+++ b/src/mcp_atlassian/servers/jira.py
@@ -348,6 +348,18 @@ async def get_issue(
             default=True,
         ),
     ] = True,
+    use_display_names: Annotated[
+        bool,
+        Field(
+            description=(
+                "When true, custom field keys in the output use human-readable "
+                "display names (e.g. 'Story Points') instead of opaque IDs "
+                "(e.g. 'customfield_10243'). The 'names' expansion is added "
+                "automatically. Standard fields are unaffected."
+            ),
+            default=False,
+        ),
+    ] = False,
 ) -> str:
     """Get details of a specific Jira issue including its Epic links and relationship information.
 
@@ -359,6 +371,7 @@ async def get_issue(
         comment_limit: Maximum number of comments.
         properties: Issue properties to return.
         update_history: Whether to update issue view history.
+        use_display_names: Opt into human-readable custom field keys.
 
     Returns:
         JSON string representing the Jira issue object.
@@ -371,6 +384,10 @@ async def get_issue(
     if fields and fields != "*all":
         fields_list = [f.strip() for f in fields.split(",")]
 
+    # Automatically include 'names' expansion when display names are requested
+    if use_display_names and (not expand or "names" not in expand):
+        expand = f"{expand},names" if expand else "names"
+
     issue = jira.get_issue(
         issue_key=issue_key,
         fields=fields_list,
@@ -379,7 +396,10 @@ async def get_issue(
         properties=properties.split(",") if properties else None,
         update_history=update_history,
     )
-    result = issue.to_simplified_dict()
+    if use_display_names:
+        result = issue.to_display_name_dict()
+    else:
+        result = issue.to_simplified_dict()
     return json.dumps(result, indent=2, ensure_ascii=False)
 
 
@@ -451,6 +471,18 @@ async def search(
             default=None,
         ),
     ] = None,
+    use_display_names: Annotated[
+        bool,
+        Field(
+            description=(
+                "When true, custom field keys in the output use human-readable "
+                "display names (e.g. 'Story Points') instead of opaque IDs "
+                "(e.g. 'customfield_10243'). The 'names' expansion is added "
+                "automatically. Standard fields are unaffected."
+            ),
+            default=False,
+        ),
+    ] = False,
 ) -> str:
     """Search Jira issues using JQL (Jira Query Language).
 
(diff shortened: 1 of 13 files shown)


Troubladore:

src/mcp_atlassian/servers/jira.py
```diff
@@ -597,7 +612,10 @@ async def get_issue(
         properties=properties.split(",") if properties else None,
         update_history=update_history,
     )
-    result = issue.to_simplified_dict()
+    if use_display_names:
+        result = issue.to_display_name_dict()
```
sooperset: **BLOCKER:** This converts custom fields before the `include` output keys exist. A custom field named `remote_links`, `transitions`, `watchers`, or `worklogs` is silently overwritten by the later enrichment assignment. With empty `comments` or `changelogs`, `setdefault` instead leaves the custom field under the reserved key and the requested section has the wrong shape. I reproduced this through `jira_get_issue` with `use_display_names: true` and `include: "remote_links"`, where `customfield_10001` disappeared. Required for merge: reserve every requested include key during display-name selection (or re-key after enrichment), then add server regression tests proving both the custom value and the included section survive.

sooperset (changes requested):
Thanks for the contribution @faizbawa! Preserving the Cloud names map across pages and adding Cloud/DC end-to-end cases target the right failure modes.

1. `jira_get_issue` still loses or masks data when a custom field display name collides with an `include` output key.

src/mcp_atlassian/servers/jira.py
faizbawa: Thanks for the review. Will check this out.

faizbawa: Fixed in ccc2c15. to_display_name_dict() now takes extra_reserved_keys.

Server now passes the include output keys so custom fields can't claim them. On collision the raw ID is preserved; both the custom value and the enrichment section survive. Added 8 parameterized regression tests covering all include sections.

sooperset:
The collision fallback is well done: custom field values now survive clashes with standard keys, duplicate display names, and all 6 `jira_get_issue` include outputs. The Cloud and Data Center coverage is exactly what this shared Jira path needed, thanks @faizbawa!

PR #1158, opened 2026-03-12
ATLASSIAN_OAUTH_ENABLE check to require service URL

harsh1netapp:
When `ATLASSIAN_OAUTH_ENABLE=true` is set without a corresponding `CONFLUENCE_URL` or `JIRA_URL`, the service is incorrectly marked as available. This causes downstream failures when per-user fetchers are created without a base URL.

This is particularly problematic in multi-user or gateway deployments where only one service URL is configured and users provide credentials through authorization headers. Minimal OAuth availability is now scoped to each service URL, so configuring Confluence no longer exposes unusable Jira tools, and vice versa.

The regression coverage includes no URLs, both URLs, each individual service URL, accepted enable values, and disabled values.

## Testing

- `uv run pytest tests/unit/ -q` (3270 passed, 5 skipped)
- `uv run pytest tests/integration/ -q` (23 environment-gated skips)
- `uv run pytest tests/e2e/cloud --cloud-e2e -q` (89 passed, 15 skipped)
- `uv run pytest tests/e2e/ --dc-e2e -q` (99 passed, 105 skipped)

### DIFF
--- a/src/mcp_atlassian/utils/environment.py
+++ b/src/mcp_atlassian/utils/environment.py
@@ -109,16 +109,15 @@ def get_available_services(
             pat_env="CONFLUENCE_PERSONAL_TOKEN",
         )
 
-    if not confluence_is_setup and os.getenv("ATLASSIAN_OAUTH_ENABLE", "").lower() in (
-        "true",
-        "1",
-        "yes",
-    ):
-        confluence_is_setup = True
-        logger.info(
-            "Using Confluence minimal OAuth configuration "
-            "- expecting user-provided tokens via headers"
-        )
+        # OAuth enable check with known URL — expecting user-provided tokens via headers
+        if not confluence_is_setup and os.getenv(
+            "ATLASSIAN_OAUTH_ENABLE", ""
+        ).lower() in ("true", "1", "yes"):
+            confluence_is_setup = True
+            logger.info(
+                "Using Confluence OAuth with CONFLUENCE_URL "
+                "- expecting user-provided tokens via headers"
+            )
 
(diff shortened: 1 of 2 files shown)


Troubladore:

src/mcp_atlassian/utils/environment.py
```diff
@@ -109,16 +114,12 @@ def get_available_services(
...
-        "true",
-        "1",
-        "yes",
-    ):
-        confluence_is_setup = True
-        logger.info(
-            "Using Confluence minimal OAuth configuration "
-            "- expecting user-provided tokens via headers"
-        )
+        if not confluence_is_setup and oauth_enabled:
```
sooperset, line 117: **BLOCKER:** Moving this fallback inside `if confluence_url` also disables the existing URL-less minimal BYOT flow for both services. `.env.example` documents `ATLASSIAN_OAUTH_ENABLE=true` without URLs for gateways that provide the Bearer token and Cloud ID per request, and both config factories still accept that setup. With this change, startup creates neither base config, so those requests can't build user fetchers and all Jira and Confluence tools disappear.

Required for merge: when either service URL is configured, enable only services with their matching URL. When neither URL is configured, preserve the existing behavior that enables both services for minimal BYOT. Please cover both states at the server lifecycle or tool-listing level.

sooperset (changes requested):
Thanks for the contribution @harsh1netapp! The service-specific URL test matrix is useful and covers the cross-service exposure case well.

1. Preserve the documented URL-less minimal BYOT mode while keeping the new per-service behavior when either URL is configured.

Issue #164, opened 2025-03-28
Feature Request: Return counts of search result too


lawwu:
Currently `search_issues` only returns the list of JiraIssues. It would be nice if there was a method to return the counts too. I think you already have this in JiraSearchResult: https://github.com/sooperset/mcp-atlassian/blob/main/src/mcp_atlassian/jira/search.py#L82

This would enable answering questions like "How many tickets are there in .... ?" Right now this relies on the AI to correctly count by reasoning over the returned text of issues.

sooperset:
Hi @lawwu, Thanks for reporting! I've added total, start_at, and max_results to the search results with PR #203. Please test with these new changes and reopen this issue if you still have the problem.

lawwu:
Thank you @sooperset for adding this!

PR #1164, opened 2026-03-13
fix(jira): recognize legacy Cloud account IDs by shape

NickLoeza:
## Summary

- recognize legacy 24-character hexadecimal Jira Cloud account IDs regardless of their first character
- preserve the modern `<digits>:<uuid>` form and documented `accountid:` prefix already supported on current `main`
- use full-shape matching so malformed IDs, email addresses, display names, and usernames still use normal user lookup
- add regression coverage for accepted account IDs and lookup fallbacks

## Background

`_get_account_id()` previously treated an unprefixed value as an account ID only when it started with `5` and had at least 10 characters. That rejected known legacy IDs such as `606b8fb83a516300764cb19d`, while also accepting malformed values such as the former 17-character test fixture.

The branch is rebased onto current `main`, which added the modern colon form and the explicit prefix for other opaque account IDs. This PR keeps both behaviors and replaces only the fragile legacy heuristic.

## Verification

- [x] `uv run pytest tests/unit/jira/test_users.py -xvs` — 69 passed
- [x] `uv run pytest tests/unit/ -xq` — 3,068 passed, 5 skipped
- [x] `uv run pytest tests/integration/ -xvs` — 23 environment-gated tests skipped as designed
- [x] `uv run pytest tests/e2e/cloud/ --cloud-e2e -xvs` — 82 passed, 15 skipped
- [x] `uv run pytest tests/e2e/ --dc-e2e -xvs` — 94 passed, 98 skipped
- [x] `uv run pre-commit run --all-files` — Ruff formatting/lint and mypy passed

### DIFF
--- a/src/mcp_atlassian/jira/users.py
+++ b/src/mcp_atlassian/jira/users.py
@@ -121,7 +121,7 @@ class UsersMixin(JiraClient):
             ValueError: If the account ID could not be found.
         """
         # If it looks like an account ID already, return it
-        if assignee.startswith("5") and len(assignee) >= 10:
+        if re.match(r"^[0-9a-f]{24}$", assignee) or re.match(r"^\d+:", assignee):
             return assignee
 
         account_id = self._lookup_user_directly(assignee)
--- a/src/mcp_atlassian/jira/users.py
+++ b/src/mcp_atlassian/jira/users.py
@@ -121,16 +121,16 @@ class UsersMixin(JiraClient):
             ValueError: If the account ID could not be found.
         """
         # If it looks like an account ID already, return it.
-        # Cloud account IDs come in two shapes: the legacy 24-char hex format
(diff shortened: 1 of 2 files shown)


Troubladore:
We identified the same bug — the `startswith("5")` heuristic rejects valid Atlassian account IDs starting with other hex digits. We've implemented our own fix in [`community/main`](https://github.com/Troubladore/mcp-atlassian/tree/community/main) ([PR #164](https://github.com/Troubladore/mcp-atlassian/pull/164)) using regex matching for both 24-char hex and digits:uuid formats.

Your PR's CI is currently failing because the new regex doesn't match the test fixture `5abcdef1234567890` (17 chars — not a valid 24-char Atlassian account ID). The test fixture needs updating to a realistic ID length.

You're welcome to use our community branch until this upstream catches up.

sooperset:
The full-shape matcher and parametrized regression cases cover both legacy hex IDs and lookup fallbacks cleanly. Thanks for tightening this path, @NickLoeza.

PR #1170, opened 2026-03-16
fix: resolve issue type IDs for all types on localized Jira instances

seongjinlim:
## Problem

On localized Jira instances (e.g., Korean, Japanese, German), `create_issue()` fails with "Invalid issue type" for any type other than Epic or Subtask.

**Root cause (two issues):**

1. **`get_project_issue_types()`** checks for `"values"` key in the createmeta response, but `atlassian-python-api`'s `issue_createmeta_issuetypes()` returns `"issueTypes"` key — causing it to return an empty list.

2. **`create_issue()`** only resolves issue type name→ID for Epic and Subtask. For other types like Story, Task, or Bug, it sends `{"name": "스토리"}` directly, which Jira rejects because it expects the ID.

**Reproduction:** On a Korean Jira Cloud instance, call `create_issue(project_key="PROJ", issue_type="스토리", ...)` → fails with "선택한 이슈 유형이 올바르지 않습니다" (Invalid issue type).

### `projects.py` — `get_project_issue_types()`
Add `"issueTypes"` key fallback between the existing `"values"` and `"projects"` fallbacks:

```python
issue_types = meta.get("values", [])
if not issue_types:
    issue_types = meta.get("issueTypes", [])  # <-- NEW
if not issue_types:
    # existing fallback for older format
```

### `issues.py` — `create_issue()` + `_find_issue_type_id()`
- Add `else` branch in `create_issue()` to resolve IDs for all non-Epic/Subtask issue types
- Add `_find_issue_type_id()` method that matches by `name` (case-insensitive) with `untranslatedName` fallback

## Testing

Verified on a Korean Jira Cloud instance (doosanrobotics.atlassian.net):
- `issue_createmeta_issuetypes("PAP")` returns `{"issueTypes": [...]}` (not `"values"`)
- After fix: `스토리` resolves to `id=10115`, issue created successfully
- Epic and Subtask resolution unchanged (existing code paths)

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -605,7 +605,7 @@ class IssuesMixin(
                     "Provide issue_type like 'Task', 'Story', or 'Bug'."
                 )
 
-            # Handle Epic and Subtask issue type names across different languages
+            # Handle issue type names across different languages
             actual_issue_id = None
             if self._is_epic_issue_type(issue_type) and issue_type.lower() == "epic":
                 # If the user provided "Epic" but we need to find the localized name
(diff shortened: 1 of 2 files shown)


Troubladore:
This PR has been integrated into [`community/main`](https://github.com/Troubladore/mcp-atlassian/tree/community/main) on our actively maintained community fork, with additional tests added for the new _find_issue_type_id method and the issueTypes key fallback. You are welcome to use that branch until this upstream catches up.

Issue #1104, opened 2026-03-02
tracking: outcome-oriented tool restructuring roadmap (v0.22 → v1.0)


sooperset:
## Roadmap: Outcome-Oriented Tool Restructuring

Reference: https://www.philschmid.de/mcp-best-practices

Incremental path from 73 tools to ~25 outcome-oriented tools, targeting v1.0.0 as the stable API.

### Principle
Each release independently releasable. New params are additive, deprecations log warnings, removals happen in dedicated cleanup releases.

### Phases

| Release | Phase | Focus | Tool count | Status |
|---------|-------|-------|-----------|--------|
| **v0.22** | A | `include` on get_issue/get_page, orchestration on update_issue | 73 (additive) | In progress |
| **v0.23** | B | `discover` consolidation (merge 10 discovery tools into 1), deprecate absorbed tools | ~63 | Planned |
| **v0.24** | C | Merge attachment/media tools, merge sprint/version management tools | ~50 | Planned |
| **v0.25** | D | Remove all deprecated tools from previous phases | ~35 | Planned |
| **v1.0.0** | — | Stable API: clean ~25-tool surface as public contract | ~25 | Planned |

### Phase A issues (v0.22)
- [x] #1100 — Shared helpers
- [ ] #1101 — `jira_get_issue` include param
- [ ] #1102 — `jira_update_issue` orchestration
- [ ] #1103 — `confluence_get_page` include param

### Phase B targets (v0.23)
- `jira_discover` meta-tool: merge `get_all_projects`, `get_agile_boards`, `get_sprints_from_board`, `search_fields`, `get_field_options`, `get_link_types`, `get_project_versions`, `get_project_components`, `get_service_desk_for_project`, `get_service_desk_queues`
- Deprecation warnings on absorbed tools

### Phase C targets (v0.24)
- Merge `download_attachments` + `get_issue_images` → unified media tool
- Merge `create_sprint` + `update_sprint` + `add_issues_to_sprint` → `manage_sprint`
- Merge `create_version` + `batch_create_versions` → `manage_version`
- Merge Confluence attachment tools (7 → 2)

### Phase D targets (v0.25)
- Remove all deprecated v1 tools
- Remove deprecated toolset definitions
- Update tool counts in `test_toolsets.py`

sooperset:
### Proposal: Reduce multi-tool chaining with inline params

Currently, a simple "update issue and transition to Done with a comment" requires 3 separate tool calls:

[...]

PR #1174, opened 2026-03-16
feat(confluence): add content properties tools (get/set page display width and metadata)

### LINKED ISSUE #1173: feat(confluence): add content properties tools (get/set page display width and metadata)
## Problem

There is currently no way to read or write **Confluence content properties** through mcp-atlassian. Content properties are key-value metadata stored against a page via the Confluence REST API and control important page-level behaviours — most visibly **page display width** (full-width vs fixed-width layout).

### Concrete use case that surfaced the gap

While updating a Confluence page via `confluence_update_page`, I wanted to switch the page to wide/full-width display mode. This is not a storage-format setting — it is a content property:

| Property key | Value |
|---|---|
| `content-appearance-published` | `"fixed-width"` or `"full-width"` |
| `content-appearance-draft` | `"fixed-width"` or `"full-width"` |

Because no MCP tool existed, I had to fall back to raw `curl` calls — three separate HTTP requests that an LLM assistant should be able to handle through the MCP server directly.

## Proposed solution

Two new tools:

### `confluence_get_content_properties`

Returns all content properties for a page as `{ key: value }`, or a single property when `key` is provided.

**Confluence API:** `GET /wiki/rest/api/content/{id}/property` and `GET /wiki/rest/api/content/{id}/property/{key}`

### `confluence_set_content_property`

Creates or updates a content property. Handles the version increment automatically (reads current version, writes `version + 1`), so callers never need to manage version numbers.

**Confluence API:** `GET` to check existence/version → `POST` to create or `PUT` to update.

## Example

```python
# Set page to full-width display
confluence_set_content_property(
    page_id="1491730434",
    key="content-appearance-published",
    value='"full-width"'
)

# Read all properties on a page
confluence_get_content_properties(page_id="1491730434")
# → { "content-appearance-published": "full-width", ... }
```

## Other use cases this unlocks

- **Editor version** (`editor` key with `{"version": 2}`)
- **Custom metadata** — arbitrary key-value pairs for automation
- **App integrations** — Confluence apps store state as content properties

## Implementation notes

- Works for both Confluence Cloud and Server/Data Center (same API path)
- `409 Conflict` (version mismatch) is avoided by the auto-version upsert pattern
- Value is accepted as a JSON string and parsed before sending to the API

---

I have a working implementation ready and will submit a PR linked to this issue.

kbichave:
## Description

Fixes #1173

Adds MCP support for reading and writing arbitrary Confluence content properties. These key-value metadata entries are used by Confluence and apps for settings such as editor versions, layout behavior, and custom automation state.

## Changes

- Adds `PropertiesMixin.get_content_properties(page_id, key=None)` to return all properties or one property by key.
- Adds `PropertiesMixin.set_content_property(page_id, key, value)` to create or update a property with automatic version increments.
- Registers `confluence_get_content_properties` as a read-only tool.
- Registers `confluence_set_content_property` as a write-protected tool that accepts JSON values.
- Routes requests through the configured Atlassian client and the correct Confluence v1 base URL for Cloud, Server/Data Center, and Cloud OAuth gateway authentication.
- URL-encodes property keys and validates Confluence response shapes.
- Updates tool documentation, generated-doc mappings, and tool-count assertions.

## Example

```json
{
  "page_id": "1491730434",
  "key": "editor",
  "value": "{\"version\": 2}"
}
```

## Verification

- [x] `pre-commit run --all-files`
- [x] Full unit suite: 3,245 passed, 5 skipped
- [x] Focused Confluence/server/toolset suite: 586 passed
- [x] Cloud e2e: 87 passed, 15 skipped; one unrelated Jira cleanup race passed on immediate isolated retry
- [x] Data Center e2e: 99 passed, 104 skipped
- [x] Integration suite collected successfully; 23 environment-gated tests skipped

## Checklist

- [x] Code follows project style and typing checks.
- [x] Regression and MCP tool tests are included.
- [x] Cloud and Data Center behavior is verified against live instances.

### DIFF
--- a/src/mcp_atlassian/confluence/properties.py
+++ b/src/mcp_atlassian/confluence/properties.py
@@ -42,14 +42,9 @@ class PropertiesMixin(ConfluenceClient):
                 )
                 response.raise_for_status()
                 data = response.json()
-                return {
-                    item["key"]: item["value"]
-                    for item in data.get("results", [])
-                }
+                return {item["key"]: item["value"] for item in data.get("results", [])}
         except Exception as e:
-            logger.error(
-                f"Failed to get content properties for page {page_id}: {e}"
-            )
+            logger.error(f"Failed to get content properties for page {page_id}: {e}")
             raise Exception(
                 f"Failed to get content properties for page {page_id}: {e}"
             ) from e
--- a/src/mcp_atlassian/confluence/properties.py
+++ b/src/mcp_atlassian/confluence/properties.py
@@ -1,0 +1,113 @@
+"""Module for Confluence content property operations."""
+
+import logging
+from typing import Any
+
+from .client import ConfluenceClient
+
+logger = logging.getLogger("mcp-atlassian")
+
+
+class PropertiesMixin(ConfluenceClient):
+    """Mixin for Confluence content property operations."""
+
+    def get_content_properties(
+        self, page_id: str, key: str | None = None
+    ) -> dict[str, Any]:
+        """Get content properties for a Confluence page.
+
+        Args:
+            page_id: The ID of the page.
+            key: Optional property key. If provided, returns only that property.
+                If omitted, returns all properties as a ``{key: value}`` dict.
+
+        Returns:
+            Dict mapping property key(s) to their values.
+
+        Raises:
+            Exception: If the API request fails.
+        """
+        try:
+            base_url = self.confluence.url.rstrip("/")
+            if key:
+                url = f"{base_url}/rest/api/content/{page_id}/property/{key}"
+                response = self.confluence._session.get(url)
+                response.raise_for_status()
+                data: dict[str, Any] = response.json()
+                return {data["key"]: data["value"]}
+            else:
+                url = f"{base_url}/rest/api/content/{page_id}/property"
+                response = self.confluence._session.get(
+                    url, params={"expand": "version"}
+                )
+                response.raise_for_status()
+                data = response.json()
+                return {
+                    item["key"]: item["value"]
+                    for item in data.get("results", [])
+                }
+        except Exception as e:
+            logger.error(
+                f"Failed to get content properties for page {page_id}: {e}"
+            )
+            raise Exception(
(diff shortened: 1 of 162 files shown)


Troubladore:
Integrated into the [community fork](https://github.com/Troubladore/mcp-atlassian/tree/community/main) via [Troubladore/mcp-atlassian#168](https://github.com/Troubladore/mcp-atlassian/pull/168).

The mixin capability was preserved but the tool exposure was refactored to align with the [tool restructuring roadmap (#1104)](https://github.com/sooperset/mcp-atlassian/issues/1104):
- **Read:** `get_page(include="properties")` instead of a standalone getter tool
- **Write:** Single `confluence_set_content_property` tool (kept standalone since property-only writes can't be expressed as a param on `update_page`)
- Methods added to existing `PagesMixin` using atlassian-python-api library methods rather than a new mixin with raw `_session` calls

Net result: +1 tool (vs +2), 13 tests (vs 9). Full capability preserved.

src/mcp_atlassian/confluence/properties.py
```diff
@@ -0,0 +1,122 @@
...
+                self._content_property_url(page_id, key),
+                absolute=True,
+            )
+            if not isinstance(data, dict) or "key" not in data or "value" not in data:
+                raise ValueError("Confluence returned an invalid content property")
+            return {str(data["key"]): data["value"]}
+
+        data = self.confluence.get(
+            self._content_property_url(page_id),
+            params={"expand": "version", "limit": 200},
```
sooperset, line 50: **BLOCKER:** This method promises all properties, but this request reads only the first page. Confluence allows an unlimited number of properties per content item and paginates bulk responses, so pages beyond the server's per-request limit are silently omitted. Required for merge: follow `_links.next` or advance `start` until exhausted, then add a unit test with a second response page.

sooperset (changes requested):
Thanks for the contribution @kbichave! The live Cloud and Data Center create/read/update coverage is exactly the right foundation.

1. The all-properties read must consume every response page instead of returning only the first page.

PR #1183, opened 2026-03-17
fix(confluence): preserve storage output and page width for round-trip updates

slusset:
## Summary

This PR improves Confluence page round-tripping for non-OAuth flows, especially PAT/basic/token-based usage where callers need to fetch and retransmit storage content without losing Confluence-specific metadata.

## Description

Converting Confluence content to markdown is useful for readability, but it is lossy for rich pages. Macro-heavy pages can carry metadata that must be retransmitted unchanged to retain formatting and render behavior, including:

- embedded Jira display mode details such as card vs list rendering
- task metadata
- page layout details

For PAT/basic/token workflows, callers often rely on the storage path rather than OAuth-specific alternatives, so preserving storage fidelity is important for safe page updates.

## Changes

- Return exact storage XHTML when markdown conversion is disabled, without preprocessing macro-heavy content.
- Preserve custom page width on update unless the caller explicitly overrides it.
- Keep default-width updates from attempting unsupported `fixed-width` property writes.
- Clarify tool descriptions and docs so agents know when to use storage mode for safe round-tripping.
- Cover the Cloud v2 child-page path and both Cloud and Data Center round trips.

## Testing

- `uv run pytest tests/unit/confluence/test_pages.py tests/unit/servers/test_confluence_server.py -q` — 203 passed
- `uv run pytest tests/integration/ -xvs` — 23 skipped because the legacy suite requires its separate real-data flag; live coverage is below
- `uv run pytest tests/e2e/cloud --cloud-e2e -xvs` — 87 passed, 15 optional OAuth cases skipped
- `uv run pytest tests/e2e --dc-e2e -xvs` — 66 passed, 135 non-DC/optional-auth cases skipped
- Final Cloud and Data Center storage-fidelity regressions — passed in both editions
- Final Cloud and Data Center width-preservation regressions — passed in both editions
- `uv run pre-commit run --all-files` — passed (Ruff formatting/lint and mypy)

## Checklist

- [x] Regression tests cover the changes.
- [x] Relevant unit and live end-to-end tests pass.


Troubladore:
Integrated into the [community fork](https://github.com/Troubladore/mcp-atlassian/tree/community/main) via [Troubladore/mcp-atlassian#285](https://github.com/Troubladore/mcp-atlassian/pull/285). Applied as-is — clean refactor, good tests. The storage passthrough and width preservation fixes are solid improvements.

sooperset:
Thanks for the contribution @slusset! The exact storage passthrough and the Cloud and Data Center regressions give this change strong coverage.

1. Please mark the PR ready for review when you're comfortable with the maintainer follow-up commit and want us to merge it.

Once it's out of draft, I'll do the final state check and merge it.

eeichinger:
Independent confirmation that the page-width part of this PR fixes a real problem, in case it helps justify a rebase.

On v0.23.0 (Confluence Cloud, Basic auth), `confluence_update_page_section` rewrites **both** width properties even when no width argument is passed:

```
content-appearance-published : max (v1)  ->  fixed-width (v2)
content-appearance-draft     : max (v1)  ->  fixed-width (v2)
```

That matches this PR's "Keep default-width updates from attempting unsupported `fixed-width` property writes" exactly.

Two details that may be useful for the regression tests here:

- **It is not Confluence's save path.** Editing and saving the same page in the web editor left both properties untouched at `max` (v1). So the write originates in this library, not in Confluence normalising on save.
- **Both properties are affected**, not just `content-appearance-published` — worth asserting on `-draft` too.

Separately, the storage-fidelity half of this PR also lines up with what I measured: `confluence_get_page(convert_to_markdown=False)` currently re-serialises rather than returning the stored XHTML byte-for-byte — a `data-datasource="{&quot;…}"` attribute comes back as `data-datasource='{"…"}'`. Semantically equivalent and macros still render, but it means the output can't be used as a byte-exact basis for a round-trip, which is what "Return exact storage XHTML when markdown conversion is disabled" addresses.

Worth noting the section-scoped update otherwise did its job well: a Jira datasource anchor and two `status` macros outside the target section survived, and the `<ac:structured-macro>` blocks came back byte-identical including their `ac:macro-id`s. The width reset is the only thing that leaked outside the section.

I filed #1547 for the width behaviour so it's discoverable as a bug, and pointed it here. Happy to re-test against this branch if that's useful. Thanks for the work on this.

PR #1187, opened 2026-03-17
fix(jira): get field options for cascading select fields

### LINKED ISSUE #1186: [Bug]: jira get field options doesn't handle cascading select lists
### Prerequisites

### Bug Description

When you call the jira_get_field_options on Jira Data center for a cascading select list only the parent items are returned.

There appear to be two distinct bugs, one per platform.

### Cloud: cascading options has a distinct API

  File: src/mcp_atlassian/jira/field_options.py:119-130

  The Cloud path calls:
  GET /rest/api/3/field/{field_id}/context/{context_id}/option

  The Jira Cloud REST API v3 returns cascading select options as a flat list. Child options contain an optionId field referencing their parent's ID — there is no cascadingOptions nesting in this response:
```
  { "values": [
      {"id": "10001", "value": "Americas"},
      {"id": "10002", "value": "US",     "optionId": "10001"},
      {"id": "10003", "value": "Canada", "optionId": "10001"}
  ]}
```
  Jira Cloud has a separate cascade-specific endpoint that must be called per parent:
  GET /rest/api/3/field/{fieldId}/context/{contextId}/option/{optionId}/cascadingOption

  The code never calls this. The effect is that all options (parent + child) come back as flat sibling FieldOption objects with empty child_options, because the parser in field_option.py:82 will never find a cascadingOptions key:
  # field_option.py:80-84

```
 children = [
      cls.from_api_response(c)
      for c in data.get("cascadingOptions", [])   # ← key never present in Cloud response
      if isinstance(c, dict)
  ]
```

### Cause 2 — Server/DC: wrong key name for children

src/mcp_atlassian/models/jira/field_option.py:82

  Jira Server/DC createmeta allowedValues for cascading select fields nests children under the key children, not cascadingOptions:
```
  { "allowedValues": [
      {"id": "10001", "value": "Americas", "children": [
          {"id": "10002", "value": "US"},
          {"id": "10003", "value": "Canada"}
      ]}
  ]}
```
  The parser looks for cascadingOptions — a key that does not exist in Server/DC responses, so children will always be []. Only the parent FieldOption objects are returned.

### Steps to Reproduce

Execute the jira_get_field_options  tool specifying the project and fieldId for a cascading select list with child items

### Expected Behavior

The child items for cascading select lists should be returned

### Actual Behavior

```shell
On Jira Datacenter 9.x only the parent items are returned
```

### mcp-atlassian Version

0.21.0

### Installation Method

Docker

### Operating System

Other

### Python Version

3.10

### Atlassian Instance Type

Jira Server / Data Center

### Client Application

Claude code

### Additional Context

_No response_

ffsetit:
## Description

Updates get_field_options to correctly handle the Jira api response for cascading select lists.

Fixes: #1186

## Changes

**Model fixes**

- FieldOption.from_api_response:
  - Accepts both cascadingOptions and children as child keys.
  - Handles string-valued option entries and string children (creates FieldOption for string entries).

**Cloud flow fixes**

- _get_field_options_cloud:
  - Collects the Cloud main endpoint flat list of options.
  - Identifies parent option IDs and, per parent, attempts the cascade-specific endpoint:
    - `GET /rest/api/3/field/{fieldId}/context/{contextId}/option/{optionId}/cascadingOption`
  - Falls back to grouping child items by optionId when cascade endpoint fails or returns nothing.
  - Attaches `child_options` to parent `FieldOption` objects.

**Server/DC fixes**

- Parser now recognizes nested children under `children` and parses string children.

**Tests**

- Updated cloud cascading unit test to mock realistic API shapes:
- Main flat options response, then cascade-specific endpoint returning children.
- Ran targeted and full test suites; all tests passed locally.

## Testing

- [x] Manual checks performed: Executed tool using local server against Jira Data Center, sucessfully fetched custom cascade select list with multiple parent & child values

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/field_options.py
+++ b/src/mcp_atlassian/jira/field_options.py
@@ -109,8 +109,8 @@ class FieldOptionsMixin(JiraClient):
             global_ctx = next((c for c in contexts if c.is_global_context), None)
             context_id = global_ctx.id if global_ctx else contexts[0].id
 
-        # Paginate through all options
-        all_options: list[FieldOption] = []
+        # Paginate through all option entries (Cloud returns a flat list).
+        raw_items: list[dict] = []
         start_at = 0
         max_results = 100
 
@@ -127,7 +127,7 @@ class FieldOptionsMixin(JiraClient):
                 values = response.get("values", [])
                 for item in values:
                     if isinstance(item, dict):
-                        all_options.append(FieldOption.from_api_response(item))
+                        raw_items.append(item)
 
                 total = response.get("total", len(values))
                 start_at += len(values)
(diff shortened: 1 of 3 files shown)


Troubladore:
Integrated into the [community fork](https://github.com/Troubladore/mcp-atlassian/tree/community/main) via [Troubladore/mcp-atlassian#286](https://github.com/Troubladore/mcp-atlassian/pull/286). Applied as-is with conflict resolution. Good fix — the cascade-specific endpoint with pagination fallback is well-designed.

src/mcp_atlassian/jira/field_options.py
```diff
@@ -141,7 +141,70 @@ def _get_field_options_cloud(
...
+            child_items: list[dict] = []
+            # True once the cascade endpoint responds (even with empty results),
+            # so an empty authoritative response is not confused with a failure.
+            cascade_fetched = False
+            cascade_start = 0
+            cascade_max = 100
+
+            while True:
+                try:
+                    cascade_resp = self.jira.get(
```
sooperset, line 167: **BLOCKER:** This adds one `/option/{id}/cascadingOption` request for every parent option. Atlassian's current Cloud REST docs list `GET /rest/api/3/field/{fieldId}/context/{contextId}/option` with `optionId`, `onlyOptions`, `startAt`, and `maxResults` query params, and don't list this `/cascadingOption` route: https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-custom-field-options/#api-rest-api-3-field-fieldid-context-contextid-option-get

For a regular Cloud select field with 100 options, `jira_get_field_options` now makes 101 requests and 100 of them are expected 404s before fallback. Required for merge: build the hierarchy from the flat `optionId` list, or use the documented query params, without probing an undocumented endpoint for every option. Please add a regression assertion for the call count too.

sooperset (changes requested):
Thanks for the contribution @ffsetit! The Server/DC `children` parser and the precedence tests are a solid part of the fix.

1. The Cloud path needs to stop probing an undocumented `/cascadingOption` endpoint for every parent option.
2. Please add a call-count regression for a normal Cloud select field so the existing one-request path doesn't become N+1 again.

PR #1191, opened 2026-03-22
fix(utils): honor NO_PROXY env var when session proxies are explicitly set

### LINKED ISSUE #677: [Bug]: Proxy configuration issue, getting 502 in MCP server logs
### Prerequisites

### Bug Description

I am getting 502 issue when trying to se `ghcr.io/sooperset/mcp-atlassian:latest` MCP.

I have proxy restrictions on my environment so I can't directly call JIRA APIs using CURL. but as soon as I add `--noproxy '*'` in the CURL request I get the response.
I try adding the configuration on the container for enabling the proxy setting, but it's not working.
I tried with:
- `--network=host`
-  `"-e", "http_proxy"`
- `"-e", "https_proxy"`
- `"-e", "no_proxy"`
but it's still getting me same error.

### Steps to Reproduce

Enable some proxy configuration on the machine so you have to add `--noproxy '*'` when calling JIRA APis, then try to enable noproxy/proxy settings on the MCP server.

### Expected Behavior

MCP container should respect the noproxy/proxy environment variables while accessing JIRA APIs

### Actual Behavior

```shell
MCP container is not considering noproxy/proxy environment variables while accessing JIRA APIs
```

### mcp-atlassian Version

latest

### Installation Method

Other

### Operating System

### Python Version

Python is not installed in the VM

### Atlassian Instance Type

Jira Cloud

### Client Application

Cline

### Additional Context

Installation Method: Podman
Python Version: Python is not enabled on the VM

abhinavgaba:
## Description

The `requests` library ignores `NO_PROXY` when proxies are explicitly configured on `session.proxies`. This means hosts listed in `NO_PROXY` were still routed through the proxy when `HTTP_PROXY`/`HTTPS_PROXY` were set.

May fix #677.

## Changes

- Add `NoProxyAdapter` — an HTTP adapter that calls `should_bypass_proxies` on each request and clears proxies when the URL matches `NO_PROXY`
- Capture the no-proxy list on the adapter at mount time (threaded from each client's config through `configure_ssl_verification`/`configure_proxy_bypass`) instead of re-reading `NO_PROXY` per request, so Jira and Confluence with different `JIRA_NO_PROXY`/`CONFLUENCE_NO_PROXY` values stay isolated; the environment remains a fallback
- Preserve the DNS-pinning SSRF protection from the current transport when a domain-specific no-proxy adapter is mounted
- Add `configure_proxy_bypass()` — mounts `NoProxyAdapter` on a session when `NO_PROXY` is set
- `SSLIgnoreAdapter` now inherits from `NoProxyAdapter` so SSL-disabled sessions also respect `NO_PROXY`
- `configure_ssl_verification(ssl_verify=True)` mounts `NoProxyAdapter` when `NO_PROXY` is set
- Fix `urlparse(url)` being called unconditionally (broke fixtures passing mock URLs)
- Fix `send()` override signatures to match the supertype (`Mapping`, `bytes | str`) for mypy

## Testing

- [x] Unit tests added for `NoProxyAdapter`, `configure_proxy_bypass`, and the `NO_PROXY`-aware path in `configure_ssl_verification`
- [x] Regression proving two adapters with different no-proxy lists stay isolated even when the shared `NO_PROXY` environment variable is overwritten
- [x] Security regression proving the no-proxy adapter preserves DNS pinning
- [x] `uv run pytest tests/unit/ -q` — 3084 passed, 5 skipped
- [x] `uv run pytest tests/integration/ --integration -q` — 8 passed, 15 skipped
- [x] `uv run pytest tests/e2e --dc-e2e -q` — 94 passed, 98 skipped
- [x] `uv run pytest tests/e2e/cloud --cloud-e2e -q` — 81 passed, 15 skipped; one Atlassian Cloud HTTP 500 passed on immediate targeted rerun
- [x] `uv run pre-commit run --all-files` — formatting, Ruff, and mypy clean

## Checklist

- [x] Relevant unit, integration, Cloud, and Data Center verification completed.

### DIFF
--- a/src/mcp_atlassian/confluence/client.py
+++ b/src/mcp_atlassian/confluence/client.py
@@ -127,6 +127,7 @@ class ConfluenceClient:
             client_cert=self.config.client_cert,
             client_key=self.config.client_key,
             client_key_password=self.config.client_key_password,
+            no_proxy=self.config.no_proxy,
         )
 
         # Proxy configuration
--- a/src/mcp_atlassian/confluence/client.py
+++ b/src/mcp_atlassian/confluence/client.py
@@ -32,6 +32,7 @@ class ConfluenceClient:
             MCPAtlassianAuthenticationError: If OAuth authentication fails
         """
         self.config = config or ConfluenceConfig.from_env()
+        transport_url = self.config.url
 
         # Initialize the Confluence client based on auth type
         if self.config.auth_type == "oauth":
@@ -64,6 +65,7 @@ class ConfluenceClient:
                 # Cloud: use the Atlassian Cloud API URL
                 api_url = f"https://api.atlassian.com/ex/confluence/{self.config.oauth_config.cloud_id}"
                 is_cloud = True
+            transport_url = api_url
 
             # Initialize Confluence with the session
             self.confluence = Confluence(
@@ -112,10 +114,14 @@ class ConfluenceClient:
         if self.config.auth_type in ("pat", "oauth"):
             self.confluence._session.trust_env = False
 
+        if self.config.no_proxy and isinstance(self.config.no_proxy, str):
+            os.environ["NO_PROXY"] = self.config.no_proxy
+            log_config_param(logger, "Confluence", "NO_PROXY", self.config.no_proxy)
+
         # Configure SSL verification using the shared utility
         configure_ssl_verification(
             service_name="Confluence",
-            url=self.config.url,
+            url=transport_url,
             session=self.confluence._session,
             ssl_verify=self.config.ssl_verify,
             client_cert=self.config.client_cert,
@@ -137,9 +143,6 @@ class ConfluenceClient:
                 log_config_param(
                     logger, "Confluence", f"{k.upper()}_PROXY", v, sensitive=True
                 )
-        if self.config.no_proxy and isinstance(self.config.no_proxy, str):
-            os.environ["NO_PROXY"] = self.config.no_proxy
-            log_config_param(logger, "Confluence", "NO_PROXY", self.config.no_proxy)
 
         # Apply custom headers if configured
         if self.config.custom_headers:
--- a/src/mcp_atlassian/utils/ssl.py
+++ b/src/mcp_atlassian/utils/ssl.py
@@ -1,21 +1,79 @@
-"""SSL-related utility functions for MCP Atlassian."""
+"""SSL and proxy utility functions for MCP Atlassian."""
 
 import logging
+import os
 import ssl
 from typing import Any
 from urllib.parse import urlparse
 
+from requests import PreparedRequest, Response
 from requests.adapters import HTTPAdapter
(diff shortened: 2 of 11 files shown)


Troubladore:
Integrated into the [community fork](https://github.com/Troubladore/mcp-atlassian/tree/community/main) via [Troubladore/mcp-atlassian#290](https://github.com/Troubladore/mcp-atlassian/pull/290). Applied as-is — clean fix with good test coverage. The NoProxyAdapter as a base class for SSLIgnoreAdapter is a nice design.

src/mcp_atlassian/utils/ssl.py
```diff
@@ -1,21 +1,85 @@
...
+            timeout: Request timeout
+            verify: SSL verification setting
+            cert: Client certificate
+            proxies: Proxy configuration
+
+        Returns:
+            The response from the server
+        """
+        # Check if we should bypass proxies for this URL
+        no_proxy = os.environ.get("NO_PROXY") or os.environ.get("no_proxy")
```
sooperset: **BLOCKER:** This reads `NO_PROXY` from the process environment each time a request is sent, so the adapter doesn't retain the no-proxy list it was mounted with. Jira and Confluence can run in the same server process with different `JIRA_NO_PROXY` and `CONFLUENCE_NO_PROXY` values. If Jira is mounted for `jira.internal` and Confluence later sets `NO_PROXY=confluence.internal`, the Jira adapter evaluates Jira requests against the Confluence list and keeps the proxy. Please capture the configured no-proxy value on the adapter or session when mounting it, then add a regression with two different no-proxy lists.

sooperset (changes requested):
Thanks for the contribution @abhinavgaba! The adapter approach is a good fit for the explicit `session.proxies` case, and the single-client tests cover the intended bypass behavior well.

1. The mounted adapter still reads process-wide `NO_PROXY` at send time, so Jira and Confluence configs can overwrite each other.
2. Please add a regression that proves two clients or two mounted adapters with different no-proxy lists stay isolated.

src/mcp_atlassian/confluence/client.py
```diff
@@ -112,15 +114,20 @@ def __init__(self, config: ConfluenceConfig | None = None) -> None:
         if self.config.auth_type in ("pat", "oauth"):
             self.confluence._session.trust_env = False
 
+        if self.config.no_proxy and isinstance(self.config.no_proxy, str):
+            os.environ["NO_PROXY"] = self.config.no_proxy
```
abhinavgaba, line 127: We could remove setting the global `NO_PROXY` altogether, but would it be better to leave that as a follow-on change?

sooperset:
Capturing the no-proxy list on each adapter fixed the Jira/Confluence isolation edge case cleanly, and the regression coverage across both clients is exactly what this transport change needed. Thanks for following through on the review feedback.

PR #1193, opened 2026-03-23
feat(auth): add service account support via Cloud ID

### LINKED ISSUE #781: [Feature]: Service Accounts

When I let the agent to write documentation on Confluence / handle issues on Jira,
I don't want my name/api token to be used by it.
I want that the stuff it does, to appear as done by a bot/agent.
-> Service Account

I'm trying to use Service Accounts in jira/confluence so it's clear who did the change.
(You probably know this, but I'm describing for newbies)
Creating a Service Account in Atlassian Administration -> Directory is easy and it can still work with standard API keys. And the flow in the mcp configuration would be the same... almost :)
Because Service Accounts are not using the standard url path like `http://awesome.atlassian.net/wiki` but instead they follow the OAuth path:
`https://api.atlassian.com/ex/confluence/<<cloudid>>/wiki `
then the rest is the same, for example `/api/v2/pages?limit=5`

In a real life test this would look like:
```
curl --request GET \
  --url 'https://api.atlassian.com/ex/confluence/<<cloudid>>/wiki/api/v2/pages?limit=5' \
  --user '<<sa.email@serviceaccount.atlassian.com:<<sa.token>>' \
  --header 'Accept: application/json' | jq
```

I'm not sure if you support this configuration. It's similar to BYOT method, but to me it looks like that it's somewhere defaulting back to the standard Confluence URL like `http://awesome.atlassian.net/wiki`

I tried to define the url as `https://api.atlassian.com/ex/confluence/<<cloudid>>/wiki` but it returned `Out of scope error`

_No response_

### Use Case

_No response_

### Additional Context

Ah sorry, I saw the rest of the fields when I've already described my point 😄

clwluvw:
## Description

Add Atlassian Cloud ID routing for scoped service-account API tokens so Jira and Confluence automation can run under a dedicated bot identity.

Fixes #781

## Changes

- Add shared and per-service Cloud ID configuration with service-specific precedence.
- Derive Jira and Confluence API gateway URLs when a Cloud ID is configured.
- Keep scoped-token requests on the gateway, including hand-built attachment, permission, user lookup, template, restriction, and development endpoints.
- Allow the site URLs to be omitted while recommending them for generated browser links.
- Document service-account setup, scoped tokens, and Cloud ID discovery.
- Add unit coverage and a live Cloud auth-matrix variant for Cloud ID gateway routing.

## Verification

- `pre-commit run --all-files`
- `uv run pytest -q tests/unit/` — 3179 passed, 5 skipped
- Live integration suite — 23 passed
- Cloud e2e — 101 passed, 15 skipped
- Data Center e2e — 98 passed, 117 skipped

### DIFF
--- a/src/mcp_atlassian/confluence/config.py
+++ b/src/mcp_atlassian/confluence/config.py
@@ -40,6 +40,7 @@ class ConfluenceConfig:
     client_key: str | None = None  # Client private key file path (.pem)
     client_key_password: str | None = None  # Password for encrypted private key
     timeout: int = 75  # Connection timeout in seconds
+    cloud_id: str | None = None  # Atlassian Cloud ID (for service accounts)
 
     @property
     def is_cloud(self) -> bool:
@@ -49,6 +50,10 @@ class ConfluenceConfig:
             True if this is a cloud instance (atlassian.net), False otherwise.
             Localhost URLs are always considered non-cloud (Server/Data Center).
         """
+        # cloud_id set directly on config means Cloud (service account or explicit)
+        if self.cloud_id:
+            return True
+
         # OAuth with cloud_id uses api.atlassian.com which is always Cloud
         if (
             self.auth_type == "oauth"
@@ -71,6 +76,17 @@ class ConfluenceConfig:
         return is_atlassian_cloud_url(self.url) if self.url else False
 
     @property
+    def api_url(self) -> str:
+        """Get the effective API URL, using the Cloud gateway when cloud_id is set.
+
+        Returns:
+            The api.atlassian.com gateway URL if cloud_id is set, otherwise the base URL.
+        """
+        if self.cloud_id:
+            return f"https://api.atlassian.com/ex/confluence/{self.cloud_id}"
+        return self.url
+
+    @property
     def verify_ssl(self) -> bool:
         """Compatibility property for old code.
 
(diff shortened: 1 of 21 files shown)


Troubladore:
Integrated into the [community fork](https://github.com/Troubladore/mcp-atlassian/tree/community/main) via [Troubladore/mcp-atlassian#292](https://github.com/Troubladore/mcp-atlassian/pull/292). Applied as-is with conflict resolution. We clarified the env var precedence (JIRA_CLOUD_ID > ATLASSIAN_CLOUD_ID) in the field comments and from_env logic so the behavior is more visible to readers.

clwluvw:
ping @sooperset

src/mcp_atlassian/confluence/config.py
```diff
@@ -40,6 +40,7 @@ class ConfluenceConfig:
     client_key: str | None = None  # Client private key file path (.pem)
     client_key_password: str | None = None  # Password for encrypted private key
     timeout: int = 75  # Connection timeout in seconds
+    cloud_id: str | None = None  # Atlassian Cloud ID (for service accounts)
```
sooperset, line 43: **BLOCKER:** Inserting `cloud_id` before the existing `attachment_download_use_v1` field changes this public dataclass's positional signature. A call valid on main with 18 arguments now puts the final boolean in `cloud_id`, then `__post_init__` calls `.strip()` and raises `AttributeError`. Required for merge: move `cloud_id` after `attachment_download_use_v1` and add a regression test that preserves the previous positional call shape.

sooperset (changes requested):
Thanks for the contribution @clwluvw! Routing every hand-built request through `api_url` closes the scoped-token gap, and the Cloud/DC coverage is strong.

1. Preserve the existing positional signature of `ConfluenceConfig` when adding `cloud_id`.

src/mcp_atlassian/confluence/config.py
sooperset, line 43: **BLOCKER:** Inserting `cloud_id` before the existing `attachment_download_use_v1` field changes this public dataclass's positional signature. A call valid on main with 18 arguments now puts the final boolean in `cloud_id`, then `__post_init__` calls `.strip()` and raises `AttributeError`. Required for merge: move `cloud_id` after `attachment_download_use_v1` and add a regression test that preserves the previous positional call shape.

sooperset (changes requested):
Thanks for the contribution @clwluvw! Routing every hand-built request through `api_url` closes the scoped-token gap, and the Cloud/DC coverage is strong.

1. Preserve the existing positional signature of `ConfluenceConfig` when adding `cloud_id`.

PR #1194, opened 2026-03-24
feat(ssl): enhance SSL compatibility for legacy corporate servers

arioko:
## Description

This PR enhances SSL compatibility for legacy corporate servers that use outdated SSL/TLS configurations. Many enterprise Jira/Confluence instances still run on older servers that only support TLS 1.0/1.1 and legacy cipher suites. The current implementation with `ssl.create_default_context()` fails to connect to these servers, causing connection issues in corporate environments.

Fixes: #

## Changes

- Replace `ssl.create_default_context()` with `ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)` for better compatibility
- Add `context.minimum_version = ssl.TLSVersion.MINIMUM_SUPPORTED` to enable TLS 1.0/1.1 support
- Add `context.set_ciphers('DEFAULT@SECLEVEL=0')` to support legacy cipher suites
- Update unit test [test_ssl_ignore_adapter_init_poolmanager](tests/unit/utils/test_ssl.py:33:0-64:56) to mock the correct SSL context creation
- Add verification for new SSL configuration methods in tests

## Testing

- [x] Manual checks performed: Verified all SSL unit tests pass with the new implementation. Successfully tested against corporate Confluence instance with legacy SSL/TLS configuration.

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/utils/ssl.py
+++ b/src/mcp_atlassian/utils/ssl.py
@@ -37,9 +37,14 @@ class SSLIgnoreAdapter(HTTPAdapter):
             pool_kwargs: Additional arguments for the pool manager
         """
         # Configure SSL context to disable verification completely
-        context = ssl.create_default_context()
+        # Use legacy mode for compatibility with older servers (TLS 1.0/1.1)
+        context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
         context.check_hostname = False
         context.verify_mode = ssl.CERT_NONE
+        # Enable all TLS versions for maximum compatibility
+        context.minimum_version = ssl.TLSVersion.MINIMUM_SUPPORTED
+        # Add legacy cipher suites for older corporate servers
+        context.set_ciphers('DEFAULT@SECLEVEL=0')
 
         self.poolmanager = PoolManager(
             num_pools=connections,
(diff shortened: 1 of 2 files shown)


Troubladore:
Integrated a reworked version into the [community fork](https://github.com/Troubladore/mcp-atlassian/tree/community/main) via [Troubladore/mcp-atlassian#301](https://github.com/Troubladore/mcp-atlassian/pull/301).

The use case is legitimate, but we gated the legacy TLS/cipher support behind a dedicated `SSL_LEGACY_TLS=true` env var rather than applying it implicitly to all `SSL_VERIFY=false` users. The project's pattern (see `.env.example`) is to make security-affecting behavior explicit with dedicated toggles — disabling cert verification is a different risk decision from widening protocol/cipher negotiation.

When `SSL_LEGACY_TLS=true` is set, the behavior matches your PR exactly (TLS 1.0/1.1 + `SECLEVEL=0`). When not set, existing behavior is unchanged. A warning is logged when legacy mode is active.

sooperset:
Your diagnosis is valid: some older corporate Jira and Confluence servers can fail during TLS negotiation even when certificate verification is disabled.

I can't merge this version as-is because it changes the security profile for every `SSL_VERIFY=false` user. Disabling certificate checks and allowing TLS 1.0/1.1 plus `DEFAULT@SECLEVEL=0` are separate risk decisions, and this PR couples them.

For this to be mergeable, it would need:

1. Gate legacy protocol and cipher support behind a dedicated opt-in env var, for example `SSL_LEGACY_TLS=true`.
2. Keep the current `SSL_VERIFY=false` behavior unchanged when that env var isn't set.
3. Log a clear warning when legacy TLS mode is active.
4. Add regression tests for both paths: default disabled-verification behavior and explicit legacy TLS behavior.
5. Document the option in `.env.example`.

I'm closing this PR in its current form. A narrowed follow-up with that opt-in behavior and tests would be a much better fit.

PR #1196, opened 2026-03-24
feat(auth): add mTLS client certificate authentication for Server/DC

keithxm23:
Description

Adds "cert" as a first-class authentication type for Jira and Confluence Server/Data
Center deployments that use mutual TLS (mTLS). In some corporate environments, a
client certificate/key pair is sufficient to authenticate with the API — no username,
password, or personal access token is needed.

Previously, setting JIRA_CLIENT_CERT / CONFLUENCE_CLIENT_CERT without also providing
a username/token caused the server to log "not configured" at startup, and the
services were silently disabled.

Changes

 - utils/environment.py: _check_service_auth() accepts a new optional client_cert_env
 param; returns True for Server/DC when the cert env var is set. Both Jira and 
Confluence call sites now pass it.
 - jira/config.py + confluence/config.py: add "cert" to the auth_type Literal; detect
 *_CLIENT_CERT before raising a missing-credentials error; is_auth_configured() 
handles the cert type.
 - jira/client.py + confluence/client.py: new elif auth_type == "cert": branch 
initialises the client without credentials, allowing the existing 
configure_ssl_verification() call to apply the cert/key to the session.

Testing

 - [X]  Unit tests added/updated — all 2547 existing tests pass (uv run pytest 
tests/unit/ -q)
 - [x]  Integration tests passed
 - [X]  Manual checks performed: server starts and logs Using Jira mTLS client 
certificate authentication / Using Confluence mTLS client certificate authentication;
 Jira and Confluence tools function correctly against a corporate Server/DC instance 
using a PEM client certificate.

Checklist

 - [X]  Code follows project style guidelines (linting passes).
 - [X]  Tests added/updated for changes.
 - [X]  All tests pass locally.
 - [x]  Documentation updated (if needed).


Troubladore:
Integrated into the [community fork](https://github.com/Troubladore/mcp-atlassian/tree/community/main) via [Troubladore/mcp-atlassian#299](https://github.com/Troubladore/mcp-atlassian/pull/299). Two changes from your PR:

- Downgraded cert path logging from plaintext to boolean presence (cert paths reveal filesystem layout)
- Added 14 tests covering config detection, precedence, `is_auth_configured`, client init, and `trust_env` behavior

The feature itself is clean and aligns well with the project's existing config model. Thanks for the contribution!

keithxm23:
Pulled in https://github.com/Troubladore/mcp-atlassian/pull/299 which added test coverage and cleaned up logging.

What can I do to help get this merged?

shuklapranjal:
Can somebody take a look at this PR and merge it please.

src/mcp_atlassian/jira/client.py
```diff
@@ -107,6 +107,19 @@ def __init__(self, config: JiraConfig | None = None) -> None:
                 verify_ssl=self.config.ssl_verify,
                 timeout=self.config.timeout,
             )
+        elif self.config.auth_type == "cert":
```
sooperset, line 115: **BLOCKER:** This adds cert-only auth to the main Jira session, but `_lookup_user_by_permissions()` still calls `requests.get()` directly. Under this auth type the fallback sends empty Basic credentials, never uses `self.jira._session`, and never supplies the client certificate. A strict mTLS endpoint will reject the connection, so user resolution and assignment can fail whenever direct lookup falls through to permission search. Required for merge: route that fallback through the configured Jira session, or explicitly carry its certificate and other session settings, and add a cert-auth regression test.

docs/troubleshooting.mdx
```diff
@@ -255,7 +255,12 @@ Header values containing sensitive information are automatically masked in logs.
     CONFLUENCE_SSL_VERIFY=false
     ```
 
-    For mTLS (mutual TLS) authentication:
+    For mTLS (mutual TLS) authentication, set either a combined cert/key PEM:
```
sooperset, line 258: **IMPORTANT:** This is the only user-facing mTLS setup text, while the authentication guide still says the project supports 3 methods and the configuration guide omits the certificate and key variables. Users shouldn't have to discover a first-class auth method under SSL troubleshooting. Required for merge: add Server/DC mTLS setup to the authentication guide and list both Jira and Confluence variables, combined-versus-split PEM behavior, and the encrypted-key limitation in the configuration guide.

sooperset (changes requested):
Thanks for the contribution @keithxm23! The cert-only startup detection, combined PEM wiring, and both-edition verification are now in good shape.

1. Route Jira's permission-search fallback through the mTLS-configured session.
2. Document mTLS as a first-class auth method in the primary auth and configuration guides.

src/mcp_atlassian/jira/client.py
keithxm23: Thank you for catching that. I pushed a fix

docs/troubleshooting.mdx
keithxm23: Thanks, I pushed an update to the docs

keithxm23:
Testing                                                                                               
                                                                                                         
   • Unit tests added/updated — all 3473 existing tests pass ( uv run pytest tests/unit/ -q )            
   • Integration tests passed ( uv run pytest tests/integration/test_ssl_verification.py --integration , 
      8 passed / 2 skipped requiring live  JIRA_URL / CONFLUENCE_URL )                                   
   • Manual checks performed: server starts and logs  Using Jira mTLS client certificate authentication  
      /  Using Confluence mTLS client certificate authentication  with a self-signed combined PEM cert;  
     Jira and Confluence clients initialize correctly with  AUTH_TYPE=cert .                             
   • Rebased onto latest  main , resolving conflicts with the new  external  auth passthrough and        
     NO_PROXY support (both features now coexist).                                                       
                                                                                                         
   Checklist                                                                                             
                                                                                                         
   ✓ Code follows project style guidelines (linting passes — ruff, mypy)                                 
   ✓ Tests added/updated for changes                                                                     
   ✓ All tests pass locally                                                                              
   ✓ Documentation updated

keithxm23:
After pulling in the latest main and resolving conflicts, I ran into a gap in the docs while
 testing the mTLS setup end-to-end with a local MCP client (OpenCode).

 What happened: With cert-only auth configured (JIRA_CLIENT_CERT / CONFLUENCE_CLIENT_CERT, no PAT
 or Basic credentials), the UserTokenMiddleware rejects any request that arrives without an
 Authorization header — because without explicit opt-in it refuses to silently fall back to
 operator-level credentials. A local MCP client like OpenCode sends no auth header by default, so
 every request was rejected with 401.

 The fix: Set ALLOW_GLOBAL_CRED_FALLBACK=true in the environment. This opts the server in to using
 the globally configured cert identity for unauthenticated requests. Without this, mTLS cert auth
 is effectively unusable from any MCP client that doesn't also supply its own per-request
 Authorization header.

 This commit (17a338c) adds a <Note> to the mTLS section of docs/authentication.mdx calling out
 this requirement and the security trade-off (only safe when the server isn't exposed to untrusted
 callers).

PR #1197, opened 2026-03-24
feat(jira): expose accountId through get_user_profile

clwluvw:
## Description

exposes account id of the user as well through get_user_profile(). it could be helpful as email is mostly empty in webhooks that mcp wants to be triggered by.

## Testing

## Checklist

- [ ] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/models/jira/common.py
+++ b/src/mcp_atlassian/models/jira/common.py
@@ -85,6 +85,8 @@ class JiraUser(ApiModel):
             "email": self.email,
             "avatar_url": self.avatar_url,
         }
+        if self.account_id:
+            result["account_id"] = self.account_id
         if self.user_key:
             result["key"] = self.user_key
         return result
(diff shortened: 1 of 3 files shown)


Troubladore:
Integrated into the [community fork](https://github.com/Troubladore/mcp-atlassian/tree/community/main) via [Troubladore/mcp-atlassian#297](https://github.com/Troubladore/mcp-atlassian/pull/297). Applied as-is — clean, tested change.

christianvogt:
Please fix this.

clwluvw:
ping @sooperset

sooperset:
Seena, your change pointed at the right missing piece: webhook-driven flows need `account_id` because email can be empty in Jira Cloud.

I rechecked this against current `main`, and the same behavior has already landed through #1347. Current `main` keeps `account_id` in the simplified user output even when it's `None`, which also covers Server/DC users.

For this PR to merge now, it would need to:

1. Rebase on current `main` cleanly.
2. Keep the current output contract: always include `account_id`, with `None` when Jira doesn't return one.
3. Drop the duplicate test changes that are already covered, or add only new coverage if there's still a gap.

Given that, I don't think this PR should merge as-is. My recommendation is to close it as superseded unless you see a separate behavior gap that #1347 didn't cover.

PR #1209, opened 2026-03-26
feat(jira): Allow to sort comments (ASC/DESC) before limiting

### LINKED ISSUE #653: [Bug]: MCP does not understand "get the last comment of Jira issue x"
### Prerequisites

### Bug Description

MCP always always uses jira_issue_get with argument comment_limit=1 which returns the oldest comment, not the last one. I have to explicitly instruct it to use comment_limit=100 and look at the timestamps, which does not work reliably and pollutes context with unnecessary information.

### Steps to Reproduce

Prompt the model to get the latest comment of Jira issue X. 

### Expected Behavior

MCP is supposed to use the correct api call to return the latest comment

### Actual Behavior

```shell
Model uses jira_issue_get tool with argument "comment_limit: 1" which returns the oldest comment.
```

### mcp-atlassian Version

0.11.9

### Installation Method

Docker

### Operating System

macOS

### Python Version

n/a

### Atlassian Instance Type

Jira Cloud

### Client Application

LM Studio

### Additional Context

_No response_

jkellerer:
## Description

The PR adds the ability to sort JIRA case comments in descending order (on create date) before applying the comment limit. This allows to fetch the newest comments instead of the oldest (current default). 

For backward compatibility, no sorting is applied when the parameter is omitted.

Fixes: #653

## Changes

- Added `comment_sort = ( None | 'asc' | 'desc' )` as input parameter to `get_issue`.
- Also added `sort` or `comment_sort` to all internal implementations to fetch comments.
- Test for `CommentsMixin` (the only location where comment limit is tested) are extended to test the sort order.

## Testing

- [x] Unit tests added/updated (Note: only where tests for comment limit existed)
- [x] Manual checks performed: 
      Using the PR with cl..-code. When asking for oldest 3 and newest 3 comments it works as expected (2 calls, asc and desc sorting, limit=3).

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/comments.py
+++ b/src/mcp_atlassian/jira/comments.py
@@ -1,7 +1,7 @@
 """Module for Jira comment operations."""
 
 import logging
-from typing import Any
+from typing import Any, Literal
 
 from ..models.jira.adf import adf_to_text
 from ..utils import parse_date
@@ -14,7 +14,10 @@ class CommentsMixin(JiraClient):
     """Mixin for Jira comment operations."""
 
     def get_issue_comments(
-        self, issue_key: str, limit: int = 50
+        self,
+        issue_key: str,
+        limit: int = 50,
+        sort: None | Literal["asc", "desc"] = None,
     ) -> list[dict[str, Any]]:
         """
         Get comments for a specific issue.
@@ -22,6 +25,7 @@ class CommentsMixin(JiraClient):
         Args:
             issue_key: The issue key (e.g. 'PROJ-123')
             limit: Maximum number of comments to return
+            sort: Sort order for comments ("asc" or "desc"). Default is no sorting.
 
         Returns:
             List of comments with author, creation date, and content
@@ -38,7 +42,7 @@ class CommentsMixin(JiraClient):
                 raise TypeError(msg)
 
             processed_comments = []
-            for comment in comments.get("comments", [])[:limit]:
+            for comment in comments.get("comments", []):
                 processed_comment = {
                     "id": comment.get("id"),
                     "body": self._clean_text(comment.get("body", "")),
@@ -48,7 +52,12 @@ class CommentsMixin(JiraClient):
                 }
                 processed_comments.append(processed_comment)
 
-            return processed_comments
+            if sort == "asc":
+                processed_comments.sort(key=lambda x: x["created"])
+            elif sort == "desc":
+                processed_comments.sort(key=lambda x: x["created"], reverse=True)
+
+            return processed_comments[:limit]
         except Exception as e:
             logger.error(f"Error getting comments for issue {issue_key}: {str(e)}")
             raise Exception(f"Error getting comments: {str(e)}") from e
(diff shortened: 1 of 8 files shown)


faizbawa (commented):
Nice work, this is a clean implementation and the CI is all green. One thing I noticed that might be worth addressing:

### Performance regression in `comments.py` for the default (`sort=None`) path

Before this PR, `get_issue_comments` sliced **before** the per-comment processing loop:

```python
for comment in comments.get("comments", [])[:limit]:
    # _clean_text, parse_date, etc.
```

After this PR, it processes **all** comments first, then sorts, then slices:

```python
for comment in comments.get("comments", []):   # iterates ALL
    # _clean_text, parse_date on every comment
processed_comments.sort(...)
return processed_comments[:limit]
```

When `sort=None` (the default for every existing caller), this means all comments are now processed even when only a small subset is needed. For issues with hundreds of comments and a small `limit`, that's unnecessary work on the hot path.

A possible fix - restore the early-slice optimization when no sort is requested:

```python
raw_comments = comments.get("comments", [])
if sort is not None:
    # Must process all before sorting
    processed = [self._process(c) for c in raw_comments]
    processed.sort(key=lambda x: x["created"], reverse=(sort == "desc"))
    return processed[:limit]
# Fast path: slice first, then process (original behavior)
return [self._process(c) for c in raw_comments[:limit]]
```

Same pattern applies to `_get_issue_comments_if_needed` in `issues.py`, though the impact is smaller there since there's no per-comment processing - just a sort over raw dicts that could be skipped when `comment_sort is None`.

faizbawa (changes requested):

src/mcp_atlassian/jira/comments.py
```diff
@@ -38,7 +42,7 @@ def get_issue_comments(
                 raise TypeError(msg)
 
             processed_comments = []
-            for comment in comments.get("comments", [])[:limit]:
+            for comment in comments.get("comments", []):
                 processed_comment = {
```
faizbawa, line 46: When sort=None (the default for every existing caller), this means all comments are now processed even when only a small subset is needed. For issues with hundreds of comments and a small limit, that's unnecessary work on the hot path.

A possible fix - restore the early-slice optimization when no sort is requested:


```
if sort is not None:
    if sort == "desc":
        comment_list = list(reversed(comment_list))
    comment_list = comment_list[:limit]
else:
    comment_list = comment_list[:limit]
```
```
processed_comments = []
for comment in comment_list:
    processed_comment = {
        "id": comment.get("id"),
        "body": self._clean_text(comment.get("body", "")),
        "created": str(parse_date(comment.get("created"))),
        "updated": str(parse_date(comment.get("updated"))),
        "author": comment.get("author", {}).get("displayName", "Unknown"),
    }
    processed_comments.append(processed_comment)
```

faizbawa:
Also closes #1214

jkellerer:
Guess I'm also closing this PR because there is an upcoming submission from https://github.com/Troubladore/mcp-atlassian/issues/307 which will fix the issue in a better way (server side limiting, paging and sorting). Had overlooked it at first.

Funny that timings are so close, but it really is a very important thing to fix to use this MCP for production cases (with long trails)...

jkellerer:
@faizbawa thanks a lot for the review. 

For the time being, I probably change the git url to Troubladore's version as it seems to be better maintained atm (and includes this feature already).

PR #1211, opened 2026-03-27
fix(server): honour ssl_verify and proxy settings in header-based PAT branch

### LINKED ISSUE #1159: [Bug]: HTTP proxy not used for header-based PAT auth in multi-tenant setups
### Prerequisites

## Summary

In multi-tenant setups using per-request PAT injection via headers, the fetcher is created with `http_proxy=None`, bypassing any outbound HTTP proxy and causing connection timeouts.

## Context

I am running mcp-atlassian in a **multi-tenant containerized setup** (AWS AgentCore Runtime) where:
- No global PAT is configured in the container environment
- Each user's PAT is injected per-request via custom HTTP headers (`X-Atlassian-Jira-Personal-Token` / `X-Atlassian-Jira-Url`)
- All outbound traffic routes through a corporate Squid proxy (`HTTP_PROXY` / `HTTPS_PROXY` env vars set)

`tools/list` works fine (no outbound network call), but every `tools/call` times out.

## Root cause

In `servers/dependencies.py`, the header-based PAT branch creates a new config from scratch with `http_proxy=None`:

```python
header_config = spec.config_class(
    url=url_header_val,
    auth_type="pat",
    personal_token=token_header_val,
    ssl_verify=True,
    http_proxy=None,   # ← proxy never set
    https_proxy=None,
    no_proxy=None,
    ...
)
```

The other auth branches (`basic`, `oauth/pat` with token) correctly inherit proxy settings from the global config via `_create_user_config_for_fetcher`. Only the header-based PAT branch does not.

Additionally, `trust_env=False` is set on the session (PR #860), which disables the `HTTP_PROXY` / `HTTPS_PROXY` env var fallback. Both conditions together mean no proxy is ever configured for header-based PAT requests.

## Proposed fix

Read proxy env vars when creating the header-based fetcher config:

```diff
 header_config = spec.config_class(
     url=url_header_val,
     auth_type="pat",
     personal_token=token_header_val,
     ssl_verify=True,
-    http_proxy=None,
-    https_proxy=None,
-    no_proxy=None,
+    http_proxy=os.getenv("HTTP_PROXY"),
+    https_proxy=os.getenv("HTTPS_PROXY"),
+    no_proxy=os.getenv("NO_PROXY"),
     socks_proxy=None,
     ...
 )
```

This mirrors what `_create_user_config_for_fetcher` already does for the other auth branches.

## References

- #860 (introduced `trust_env=False` and the header-based PAT auth path)

### Steps to Reproduce

1. Set `HTTP_PROXY` / `HTTPS_PROXY` env vars pointing to a proxy required to reach the Atlassian instance
2. Start mcp-atlassian with **no global PAT** (`JIRA_PERSONAL_TOKEN` / `CONFLUENCE_PERSONAL_TOKEN` unset)
3. Enable header-based multi-tenant auth (`ATLASSIAN_OAUTH_ENABLE=true`)
4. Send a `tools/list` request → succeeds (no outbound network call)
5. Inject `X-Atlassian-Jira-Personal-Token` + `X-Atlassian-Jira-Url` headers and call any tool (e.g. `jira_search`) → **timeout** if the target host is only reachable via proxy

### Expected Behavior

The tool call succeeds, proxy is used as configured via `HTTP_PROXY` / `HTTPS_PROXY`.

### Actual Behavior

```shell
Connection timeout — proxy is bypassed.
```

### mcp-atlassian Version

v0.21.0

### Installation Method

### Operating System

Windows

### Python Version

3.10

### Atlassian Instance Type

Jira Server / Data Center, Confluence Server / Data Center

### Client Application

The MCP server runs as a containerized HTTP service (AWS AgentCore Runtime). 

### Additional Context

_No response_

jcbportela:
## Description

Header-based PAT authentication previously hardcoded SSL verification on and disabled all proxy settings. That made operator settings such as `JIRA_SSL_VERIFY=false`, `CONFLUENCE_SSL_VERIFY=false`, `HTTP_PROXY`, `HTTPS_PROXY`, and the service-specific proxy variables ineffective for streamable HTTP requests.

This update now:

- inherits SSL and proxy settings from the global Jira or Confluence configuration when one is available, including CLI-derived settings;
- reads service-specific network variables with shared proxy-variable fallbacks when the service has no global configuration, which is the pure per-request PAT deployment reported in #1159;
- keeps instance-specific custom headers disabled for per-request URLs so a caller-selected public URL cannot receive operator secrets;
- adds regression coverage for both the global-config and environment-only paths.

Also fixes #1159.

## Security Risk Assessment

The URL and PAT still come from request headers, while SSL and proxy behavior remain controlled by the server operator. Existing SSRF validation runs before the fetcher is created, and redirect/rebinding validation remains attached to the session.

Disabling certificate verification retains the operator-selected MITM trade-off already present in the other authentication branches. Proxy credentials stay in proxy configuration and are not emitted as target request headers. Global custom headers are intentionally not copied into the header-PAT configuration because the destination URL is selected per request.

## Verification

- `uv run pre-commit run --all-files` — passed (Ruff, formatting, and mypy)
- `uv run pytest tests/unit/servers/test_dependencies.py -xvs` — 68 passed
- `uv run pytest tests/unit/ -xvs` — 3061 passed, 5 skipped
- `uv run pytest tests/integration/ -xvs` — 23 environment-gated tests skipped
- `uv run pytest tests/e2e/cloud --cloud-e2e` — 82 passed, 15 skipped
- `uv run pytest tests/e2e --ignore=tests/e2e/cloud --dc-e2e` — 94 passed, 1 skipped
- `uv run --isolated --python 3.10 pytest tests/unit/utils/test_http.py::test_configure_rate_limit_throttles_requests -xvs` — passed on Python 3.10.20


sooperset:
Thanks for tracing the header PAT path and fixing the proxy and SSL settings without forwarding instance-specific custom headers to caller-selected URLs. The regression coverage for both configured and environment-only deployments makes the behavior clear.

PR #1212, opened 2026-03-27
fix(confluence): return TextContent for text-based attachments

Manaszz:
## Description

Many MCP clients do not forward `EmbeddedResource` blob bytes into model context. For text-based Confluence attachments, callers may therefore see only the URI and MIME type. This PR returns `TextContent` with a JSON payload containing the decoded file body so agents can read and parse it. Binary attachments continue to use `EmbeddedResource` with a base64 blob.

## Changes

- Detect text attachments by normalized MIME type and known text extensions, including ambiguous `application/octet-stream` attachments.
- Recognize `text/*`, JSON/XML media types, and structured `+json`/`+xml` suffixes.
- Return readable text from both single and batch Confluence attachment download tools.
- Preserve UTF-8 text and fall back to Latin-1 when needed.
- Document text and binary download behavior.
- Add unit coverage for MIME detection and both server tool return paths.

## Verification

- `uv run pytest tests/unit/ -q` — 3244 passed, 5 skipped.
- `uv run pytest tests/integration/ -q` — 23 skipped because this suite is credential-gated.
- `uv run pytest tests/e2e/cloud/test_confluence_cloud_operations.py --cloud-e2e -q` — 17 passed.
- `uv run pytest tests/e2e/test_confluence_dc_operations.py --dc-e2e -q` — 15 passed.
- `uv run pre-commit run --all-files` — passed, including Ruff and mypy.

### DIFF
--- a/src/mcp_atlassian/utils/media.py
+++ b/src/mcp_atlassian/utils/media.py
@@ -28,6 +28,69 @@ _IMAGE_EXTENSIONS = frozenset(
     {".png", ".jpg", ".jpeg", ".gif", ".webp", ".svg", ".bmp"}
 )
 
+_TEXT_MIME_TYPES = frozenset(
+    {
+        "application/xml",
+        "application/json",
+        "application/javascript",
+        "application/xhtml+xml",
+        "application/x-yaml",
+    }
+)
+
+_TEXT_EXTENSIONS = frozenset(
+    {
+        ".bpmn",
+        ".xml",
+        ".drawio",
+        ".json",
+        ".csv",
+        ".txt",
+        ".md",
+        ".yaml",
+        ".yml",
+        ".html",
+        ".htm",
+        ".js",
+        ".ts",
+        ".py",
+        ".java",
+        ".sql",
+    }
+)
+
+
+def is_text_attachment(media_type: str | None, filename: str | None) -> bool:
+    """Detect whether an attachment contains text-based content.
+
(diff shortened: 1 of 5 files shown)


src/mcp_atlassian/utils/media.py
```diff
@@ -28,6 +28,75 @@
     {".png", ".jpg", ".jpeg", ".gif", ".webp", ".svg", ".bmp"}
 )
 
+_TEXT_MIME_TYPES = frozenset(
+    {
+        "application/xml",
+        "application/json",
+        "application/javascript",
+        "application/xhtml+xml",
+        "application/x-yaml",
```
sooperset, line 37: **BLOCKER:** This allowlist recognizes `application/x-yaml`, but not the registered `application/yaml` type, and the suffix check below omits `+yaml`. Because extension fallback only runs for missing or ambiguous MIME values, a normal `config.yaml` reported as `application/yaml` returns `EmbeddedResource` and remains unavailable to the clients this PR is meant to fix. Required for merge: recognize `application/yaml` and structured `+yaml` types, with regression cases for both.

sooperset (changes requested):
Thanks for the contribution @Manaszz! Returning decoded text while preserving the binary resource path is the right shape, and the single/batch regression coverage is useful.

1. Complete YAML MIME detection so registered YAML content doesn't stay on the invisible blob path.
2. Remove the `Made-with: Cursor` line from the commit history to match this repository's commit policy.

See inline comments for the code-level details.

PR #1218, opened 2026-03-28
fix(jira): return newest comments instead of oldest when limiting

### LINKED ISSUE #1214: [Feature]: jira_get_issue returns oldest comments first - should return newest by default
## Problem

`jira_get_issue` with `comment_limit` always returns the **oldest** N comments, not the most recent ones. For issues with many comments (20+), the default `comment_limit=10` returns comments from months/years ago, missing recent activity entirely.

This is because `_get_issue_comments_if_needed` in `src/mcp_atlassian/jira/issues.py`:

1. Calls `self.jira.issue_get_comments(issue_key)` without any `orderBy` parameter (Jira API defaults to oldest-first)
2. Slices `comments[:comment_limit]` — taking the first N (oldest), not the last N (newest)

```python
comments = response["comments"]
if comment_limit is not None:
    comments = comments[:comment_limit]  # always oldest
```

## Suggested Fix (minimal)

Slice from the end to return the **latest** comments by default:

```python
if comment_limit is not None:
    comments = comments[-comment_limit:]  # latest N
```

## Better Fix

Add a `comment_order` parameter to `jira_get_issue`:

```python
def get_issue(
    self,
    issue_key: str,
    comment_limit: int | str | None = 10,
    comment_order: str = "newest",  # "newest" or "oldest"
    ...
```

And pass `orderBy` to the Jira API call:

```python
order_param = "-created" if comment_order == "newest" else "+created"
response = self.jira.issue_get_comments(issue_key, orderBy=order_param)
```

## Use case

AI agents frequently need to check the latest activity on a Jira issue — reviewer feedback, status updates, etc. With the current behavior, the agent has to set `comment_limit=100` and parse through all comments just to find the most recent one.

star7js:
## Summary
- `_get_issue_comments_if_needed` was slicing `comments[:comment_limit]` (oldest N) instead of `comments[-comment_limit:]` (newest N)
- With the default `comment_limit=10`, issues with 20+ comments would return months-old comments while missing all recent activity
- This is especially problematic for AI agents that need to check latest reviewer feedback or status updates

One-line change: `comments[:comment_limit]` → `comments[-comment_limit:]`

Closes #1214

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -337,9 +337,9 @@ class IssuesMixin(
 
                 comments = response["comments"]
 
-                # Limit comments if needed
+                # Return the most recent comments (newest) by slicing from the end
                 if comment_limit is not None:
-                    comments = comments[:comment_limit]
+                    comments = comments[-comment_limit:]
 
                 return comments
             except Exception as e:
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -338,7 +338,7 @@ class IssuesMixin(
 
                 comments = response["comments"]
 
-                # Return the most recent comments (newest) by slicing from the end
+                # Jira returns comments oldest-first; keep the newest comments.
                 if comment_limit is not None:
                     comments = comments[-comment_limit:]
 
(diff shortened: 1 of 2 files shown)


sooperset:
@star7js, this was a nicely scoped fix: `comment_limit` now keeps the recent activity that agents actually need, and the added regression test covers the old oldest-first behavior. Merged after refreshing the branch against current `main`.

PR #1223, opened 2026-03-31
fix(jira): paginate get_worklogs to return all entries, not just first 20

a-savinov:
## Problem

`get_worklogs()` calls `self.jira.issue_get_worklog(issue_key)`, which maps to
`GET /rest/api/2/issue/{key}/worklog` **without any pagination parameters**.
Jira's default `maxResults` is **20**, so issues with more than 20 worklog
entries silently return truncated data with no error or warning.

This means the `jira_get_worklog` MCP tool — despite its description saying it
returns "all" worklogs — was in practice always capped at 20 entries.

## Fix

Replace the single `issue_get_worklog()` call with a pagination loop that uses
explicit `maxResults` and `startAt` parameters (100 entries per page). The loop
continues until `startAt >= total`, guaranteeing all entries are returned.

```python
# Before
result = self.jira.issue_get_worklog(issue_key)  # always ≤20 entries

# After
while True:
    result = self.jira.get(url, params={"maxResults": page_size, "startAt": start_at})
    ...
    start_at += len(page)
    if start_at >= result.get("total", 0) or not page:
        break
```

## Verification

Tested on **Jira Data Center 9.x**:

| True total | Before fix | After fix | Pages fetched |
|-----------|------------|-----------|---------------|
| 68 | 20 | 68 ✅ | 1 |
| 4035 | 20 | 4035 ✅ | 41 |

## Notes

- `page_size = 100` is conservative and works on both Jira Cloud and Server/DC
- The `total` field in the response is used to know when to stop — no over-fetching
- Jira Cloud API v2 and v3 both support `startAt`/`maxResults` on this endpoint

### DIFF
--- a/src/mcp_atlassian/jira/worklog.py
+++ b/src/mcp_atlassian/jira/worklog.py
@@ -229,7 +229,9 @@ class WorklogMixin(JiraClient):
                     url, params={"maxResults": page_size, "startAt": start_at}
                 )
                 if not isinstance(result, dict):
-                    msg = f"Unexpected return value type from worklog API: {type(result)}"
+                    msg = (
+                        f"Unexpected return value type from worklog API: {type(result)}"
+                    )
                     logger.error(msg)
                     raise TypeError(msg)
 
--- a/src/mcp_atlassian/jira/worklog.py
+++ b/src/mcp_atlassian/jira/worklog.py
@@ -215,29 +215,44 @@ class WorklogMixin(JiraClient):
             Exception: If there's an error getting the worklogs
         """
         try:
-            result = self.jira.issue_get_worklog(issue_key)
-            if not isinstance(result, dict):
-                msg = f"Unexpected return value type from `jira.issue_get_worklog`: {type(result)}"
-                logger.error(msg)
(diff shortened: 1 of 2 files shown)


qwazerty:
**Tested on Jira Data Center (on-premise) — works perfectly **

We built this PR locally and tested it against a self-hosted Jira Data Center instance.

**Test case:** `jira_get_worklog` on a ticket with 285 total worklog entries.

| | Before (main) | After (this PR) |
|---|---|---|
| Entries returned | 20 | **285** |

The fix works exactly as described — the pagination loop correctly fetches all pages until `startAt >= total`.

We use this tool in an LLM-driven timesheet audit agent that processes the whole team's weekly worklogs. Before this fix, any ticket with more than 20 entries was silently returning stale/incomplete data with no warning to the model.

Would be great to get this reviewed and merged.

sooperset:
Thanks Alexey, this is a clean fix: the worklog fetch now walks every page while keeping the MCP tool response shape unchanged. The added multi-page test hits the exact truncation case that made this painful.

PR #1228, opened 2026-04-01
feat(jira): add display-name ADF mention syntax

### LINKED ISSUE #1208: [Feature]: Support for ADF mention nodes in jira_add_comment (Markdown → ADF conversion)
## Problem

When adding comments to Jira Cloud issues via `jira_add_comment`, it's currently impossible to properly mention users. 

The tool accepts Markdown, which gets converted to ADF. However, the Markdown-to-ADF converter doesn't handle mention nodes, so:

- `@Display Name` → rendered as plain text (no notification, no link)
- `[~accountid:xxx]` (Jira wiki markup) → also rendered as plain text

Add support for ADF mention nodes in the Markdown-to-ADF conversion used by `jira_add_comment`.

A syntax like `[~accountid:712020:xxx]` or `@[Display Name](accountid:xxx)` in the Markdown body should be converted to a proper ADF mention node:
{
  "type": "mention",
  "attrs": {
    "id": "712020:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
    "text": "@Display Name"
  }
}

I tried using `@Display Name` (plain text) and `[~accountid:xxx]` (Jira wiki markup) — both render as plain text in Jira Cloud. Neither triggers a real mention or notification.

### Use Case

AI agents using mcp-atlassian to reply to Jira tickets cannot properly mention/notify users in comments. This is a core collaboration feature of Jira. Without it, mentions are purely cosmetic and users aren't tagged.

### Additional Context

Tested on Jira Cloud. The `[~accountid:xxx]` syntax renders as literal text instead of being converted to an ADF mention node. The reporter is notified as a participant automatically, but explicit mentions of other users are broken.

christianvogt:
## Description

Add support for Jira Cloud user mentions in Markdown converted by `markdown_to_adf`. Jira Cloud comments and descriptions can now use `@[Display Name](accountid:ACCOUNT_ID)` to produce a real ADF `mention` node with `id` and `text` attrs, so Jira can render the mention instead of storing literal text.

Current `main` already supports Jira's `[~accountid:ACCOUNT_ID]` mention form. This PR keeps that behavior intact and adds the display-name form requested in #1208.

Fixes #1208

## Changes

- Recognize `@[Display Name](accountid:ACCOUNT_ID)` before regular Markdown links are parsed.
- Emit ADF mention nodes with `type: "mention"` and `attrs.id` / `attrs.text` for the display-name syntax.
- Preserve existing `[~accountid:ACCOUNT_ID]` mention behavior.
- Add focused unit coverage for display-name mentions, mixed mention syntaxes, and coexistence with links.
- Extend Cloud e2e mention coverage for both supported mention syntaxes.
- Document the supported Jira Cloud mention formats for comments and issue descriptions.

## Testing

- [x] `uv run pytest tests/unit/models/test_adf.py tests/unit/jira/test_comments.py tests/unit/jira/test_format_conversion.py -q`
- [x] `uv run pytest tests/unit/ -q`
- [x] `uv run pytest tests/integration --integration -q`
- [x] `uv run pytest tests/e2e/cloud --cloud-e2e`
- [x] `uv run pre-commit run --all-files`

## Checklist

- [x] Unit, integration, and Cloud e2e tests passed.
- [x] Documentation updated.

### DIFF
--- a/docs/tools/jira-comments-worklogs.mdx
+++ b/docs/tools/jira-comments-worklogs.mdx
@@ -23,7 +23,10 @@ Add a comment to a Jira issue.
 ```
 
 <Tip>
-Supports Markdown formatting. Use `visibility` parameter for restricted comments (e.g., service desk internal notes).
+Supports Markdown formatting. On Jira Cloud, mention users with
+`[~accountid:ACCOUNT_ID]` or `@[Display Name](accountid:ACCOUNT_ID)`.
+Use `visibility` parameter for restricted comments (e.g., service desk
+internal notes).
 </Tip>
 
 
--- a/docs/tools/jira-issues.mdx
+++ b/docs/tools/jira-issues.mdx
@@ -60,7 +60,10 @@ Create a new Jira issue with optional Epic link or parent for subtasks.
 ```
 
 <Tip>
-Use Markdown in the description — it's automatically converted to ADF (Cloud) or wiki markup (Server/DC). For epics, provide `epic_name` parameter.
+Use Markdown in the description - it's automatically converted to ADF (Cloud)
+or wiki markup (Server/DC). On Jira Cloud, mention users with
+`[~accountid:ACCOUNT_ID]` or `@[Display Name](accountid:ACCOUNT_ID)`.
+For epics, provide `epic_name` parameter.
 </Tip>
 
 
--- a/src/mcp_atlassian/models/jira/adf.py
+++ b/src/mcp_atlassian/models/jira/adf.py
@@ -14,7 +14,7 @@ def _parse_inline_formatting(text: str) -> list[dict[str, Any]]:
     """Parse inline Markdown formatting into ADF inline nodes.
 
     Handles: bold (**), italic (*), inline code (`), links ([text](url)),
-    and strikethrough (~~).
+    mentions (@[name](accountid:xxx)), and strikethrough (~~).
 
     Args:
         text: Raw text potentially containing inline Markdown formatting.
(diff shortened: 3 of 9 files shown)


christianvogt:
Still relevant. Would like this feature to be supported.

1kevgriff:
Independent confirmation this is still needed on current Jira **Cloud** (tested today via the official `mcp/atlassian` Docker image).

**Repro:** `jira_add_comment` cannot create a real mention. Every markdown form — `@Display Name`, `[~accountid:UUID]`, and `@[accountid:UUID]` — is stored as plain text; Jira renders the literal string and no notification fires.

**Root cause is exactly where this PR patches** — `models/jira/adf.py` → `_parse_inline_formatting`. Its inline regex only matches code / bold / strike / link / italic, so `markdown_to_adf` can never emit a `{"type":"mention"}` node from any input. One thing that masks the gap during testing: reading a comment back via `adf_to_text` reverse-renders a literal `[~accountid:…]` to `@user`-ish text, so the API response *looks* like a mention succeeded when it didn't — only the Jira UI render reveals it's plain text.

This PR is mergeable, has tests, and `Fixes #1208`. There are currently three open PRs for the same gap (#1228, #1307, #1406) — consolidating on one and getting a review would unblock a real collaboration feature for everyone driving Jira through this server. Happy to validate a chosen candidate against a live Cloud instance if that helps move it along.

sooperset:
The display-name mention syntax is nicely contained, and the Cloud e2e coverage for both mention formats is exactly the right proof here. Thanks Christian, this should make Jira Cloud comments and descriptions easier to write without breaking the existing account-id form.

PR #1231, opened 2026-04-02
fix: Jira Server/DC string response handling + xhtml content format

andersthorsen:
## Description

Two improvements for Server/DC environments:

1. **Jira Server/DC:** `atlassian-python-api` can return a raw string from `get_issue` when `response.json()` fails. This causes a `TypeError` downstream. Added fallback logic that tries JSON parsing the string, then re-fetches via direct GET.

2. **Confluence:** Added `xhtml` as an accepted `content_format` for `create_page` and `update_page`. Maps to the existing `storage` format but provides a clearer alias — AI assistants (Claude, etc.) sometimes convert XHTML to markdown unnecessarily when only `storage` is listed as an option, since they don't realize `storage` means Confluence XHTML.

## Changes

- Handle string return from `jira.get_issue()` with JSON parse + direct GET fallback
- Add `xhtml` content format option to `create_page` and `update_page` tools
- Update tool descriptions and validation to include `xhtml`
- Add unit tests for both changes

## Testing

- [x] Manual checks performed: Verified on Jira Server/DC instance where `get_issue` returns string responses

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -1,5 +1,6 @@
 """Module for Jira issue operations."""
 
+import json
 import logging
 from collections import defaultdict
 from typing import Any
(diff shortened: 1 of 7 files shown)


sooperset:
Merged this. The Server/DC string-response fallback is tightly scoped, and the `xhtml` alias makes the Confluence tool contract clearer for callers that already have storage-format HTML.

PR #1236, opened 2026-04-03
fix: route Jira issue timestamps through format_timestamp for local TZ

timwinders:
## Problem

`JiraIssue.to_simplified_dict()` bypasses `format_timestamp` for the `created` and `updated` fields, returning raw ISO strings (e.g. `2026-04-02T13:53:47.011-0700`) instead of localized timestamps.

`format_timestamp` in `base.py` already does the right thing — it converts to the user's OS local timezone via `dt.astimezone()` and includes the timezone abbreviation. But `to_simplified_dict()` was assigning `self.created` and `self.updated` directly, skipping that logic entirely.

## Fix

Route `created` and `updated` through `self.format_timestamp()` in `to_simplified_dict()`, consistent with how other timestamps are handled.

## Before / After

**Before:**
```json
"created": "2026-02-05T19:02:14.954-0800",
"updated": "2026-04-02T13:53:47.011-0700"
```

**After:**
```json
"created": "2026-02-05 19:02:14 PST",
"updated": "2026-04-02 13:53:47 PDT"
```

## Notes

- No new dependencies — uses Python stdlib `datetime.astimezone()`
- Cross-platform (Mac, Linux, Windows)
- Tested on macOS with a self-hosted Jira instance (v0.21.0)

### DIFF
--- a/src/mcp_atlassian/models/base.py
+++ b/src/mcp_atlassian/models/base.py
@@ -88,7 +88,7 @@ class TimestampMixin:
                     ts = ts[: tz_pos + 3] + ":" + ts[tz_pos + 3 :]
 
             dt = datetime.fromisoformat(ts)
-            return dt.strftime("%Y-%m-%d %H:%M:%S")
+            return dt.astimezone().strftime("%Y-%m-%d %H:%M:%S %Z")
         except (ValueError, TypeError):
             return timestamp or EMPTY_STRING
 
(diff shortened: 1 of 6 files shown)


sooperset:
Routing both Jira fields through the shared formatter and covering PST/PDT behavior gives this change a clear regression guard. Thanks for tightening this up, @timwinders.

Issue #824, opened 2026-01-05
[Bug]: MCPAtlassianAuthenticationError: Failed to configure OAuth session when using IntelliJ + GitHub Copilot


Nikhils09:
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/sooperset/mcp-atlassian/issues) to make sure this bug has not already been reported.
- [x] I have checked the [README](https://github.com/sooperset/mcp-atlassian/blob/main/README.md) for relevant information.

### Bug Description

Using MCP Atlassian with Jira Data Center (Jira DC) in IntelliJ + GitHub Copilot, calls to the MCP tools fail with an authentication error. This happens even when using a fresh, valid Personal Access Token (PAT).

Configuration used:
```
{
  "servers": {
    "jira-mcp": {
      "url": "https://mcp.jira.test/mcp/",
      "headers": {
        "Authorization": "Token MY_PAT"
      },
      "debug": true
    }
  }
}

```

### Steps to Reproduce

1. Configure MCP Atlassian with the server settings above.
2. In IntelliJ + GitHub Copilot, call the search tool.
3. Observe the MCPAtlassianAuthenticationError in logs.

### Expected Behavior

The MCP tools should run successfully with a fresh PAT.

### Actual Behavior

```shell
Fails with: mcp_atlassian.exceptions.MCPAtlassianAuthenticationError: Failed to configure OAuth session

configure_oauth_session: ensure_valid_token returned False. Token was expired: True, Refresh token present for attempt: False
ERROR Error calling tool 'search': tool_manager.py:129
Failed to configure OAuth session
MCPAtlassianAuthenticationError: Failed to configure OAuth session
Traceback (most recent call last):
  File "/app/.venv/lib/python3.10/site-packages/fastmcp/tools/tool_manager.py", line 120, in call_tool
    return await tool.run(arguments)
  File "/app/.venv/lib/python3.10/site-packages/fastmcp/tools/tool.py", line 144, in run
    result = await result
  File "/app/.venv/lib/python3.10/site-packages/mcp_atlassian/servers/jira.py", line 236, in search
    jira = await get_jira_fetcher(ctx)
  File "/app/.venv/lib/python3.10/site-packages/mcp_atlassian/servers/dependencies.py", line 256, in get_jira_fetcher
    return JiraFetcher(config=app_lifespan_ctx_global.full_jira_config)
[...]

PR #1243, opened 2026-04-07
feat(auth): add *_COMMAND env vars for lazy credential resolution

fepape-ionos:
## Description

MCP clients can eagerly start every configured server. Users who keep Atlassian secrets in tools such as `gopass` or 1Password CLI then have to unlock secret storage at startup, even when they don't use Atlassian tools in that session.

This adds `*_COMMAND` environment variables that run an external command lazily on the first tool call for that service and reuse its trimmed standard output for the process lifetime.

### Example configuration

```json
{
  "type": "stdio",
  "command": "uvx",
  "args": ["mcp-atlassian"],
  "env": {
    "JIRA_URL": "https://mycompany.atlassian.net",
    "JIRA_USERNAME": "user@company.com",
    "JIRA_API_TOKEN_COMMAND": "gopass show -o my/jira/token"
  }
}
```

## Changes

- Supports `JIRA_API_TOKEN_COMMAND`, `JIRA_PERSONAL_TOKEN_COMMAND`, `CONFLUENCE_API_TOKEN_COMMAND`, and `CONFLUENCE_PERSONAL_TOKEN_COMMAND`.
- Resolves credentials only on first use, outside the async event loop.
- Runs commands without an implicit shell. Users can invoke `sh -c` explicitly when they need shell syntax.
- Runs commands with standard input closed, so a helper cannot consume the MCP stdio transport.
- Uses each platform's own argument rules: `shlex` on POSIX, and the raw command string on Windows so `CreateProcess` parses it natively.
- Keeps the Server/Data Center precedence of personal tokens over OAuth and Basic auth for deferred credentials, and falls back to the eagerly loaded configuration if the command fails.
- Applies command results atomically, runs concurrent first calls once per service, and allows retries after failures.
- Uses a positive `CREDENTIAL_COMMAND_TIMEOUT` value, defaulting to 30 seconds.
- Keeps plain token variables higher priority than their matching command variables.
- Advertises tools only when the remaining URL, username, and edition-specific configuration is viable.
- Preserves the HTTP security boundary for operator credentials. Unauthenticated HTTP clients can't trigger deferred commands unless `ALLOW_GLOBAL_CRED_FALLBACK=true` is set.
- Documents the new variables in `.env.example` and the configuration guide.
- Adds resolver, service detection, lifespan, tool listing, dependency, concurrency, and HTTP regression tests.

## Review follow-up verification (commit `71e00ed`)

- Full unit suite: 3115 passed, 5 skipped.
- Pre-commit (Ruff, formatting, and mypy): passed.
- Stdin regression check: with the `stdin=subprocess.DEVNULL` line removed, the new test fails with the resolved token equal to `{"jsonrpc":"2.0","id":1,"method":"tools/list"}`, matching the reported reproduction.
- Live stdio probe with `JIRA_API_TOKEN_COMMAND='head -n 1'`: `initialize` and `tools/list` both answered, 58 tools listed, exit code 0, and the tool call failed with a credential error rather than a protocol desync.
- Server/DC precedence probe (static Basic plus `JIRA_PERSONAL_TOKEN_COMMAND`): startup loads `basic`, the first tool call resolves the command, and the fetcher is built with `auth_type=pat`. With a failing command it degrades to the loaded `basic` configuration.
- Cloud, Data Center, and real-data E2E suites were not re-run for this commit; the numbers below are from the previous revision.

## Earlier verification

- Focused auth and server tests: 165 passed.
- Full unit suite: 3092 passed, 5 skipped.
- Pre-commit (Ruff, formatting, and mypy): passed.
- Cloud E2E: 82 passed, 15 optional OAuth variants skipped.
- Data Center E2E: 94 passed, 1 skipped, 97 non-DC tests deselected.
- Live deferred-command smoke checks authenticated to Jira and Confluence on both Cloud and Data Center.
- Real-data integration suite: 21 passed. Its 2 legacy attachment tests are incompatible with current workspace path confinement and fail before upload, unrelated to this PR.

## Test plan

- [x] Deferred commands stay untouched during startup and tool listing.
- [x] Cloud and Data Center command-based credentials resolve on first use.
- [x] Failure, timeout, invalid command, incomplete config, concurrency, and retry paths are covered.
- [x] Unauthenticated HTTP requests can't trigger operator credential commands.
- [x] A stdin-reading helper can't consume protocol input, proven by a test that fails without the fix.
- [x] Windows argument handling covered by a deterministic test on the injected platform flag (there is no Windows CI).
- [x] Server/DC deferred personal tokens outrank Basic and OAuth, and per-user HTTP auth still inherits URL and SSL from the retained configuration.
- [x] Full unit, lint, typing, Cloud E2E, and Data Center E2E verification completed.


src/mcp_atlassian/utils/credential_command.py
```diff
@@ -0,0 +1,187 @@
...
+                except ValueError as exc:
+                    message = f"Credential command in {command_var} has invalid quoting"
+                    raise ValueError(message) from exc
+                if not arguments:
+                    message = f"Credential command in {command_var} is empty"
+                    raise ValueError(message)
+
+                logger.debug("Resolving %s via %s", target_var, command_var)
+                try:
+                    result = subprocess.run(  # noqa: S603
```
sooperset, line 191: **BLOCKER:** Don't let the helper inherit the MCP server's stdin. `subprocess.run` currently lets a helper such as `head -n 1` consume the next JSON-RPC frame and use it as the credential, which corrupts the stdio transport. Required for merge: isolate stdin, for example with `stdin=subprocess.DEVNULL`, and add a regression test proving protocol input can't be consumed.

src/mcp_atlassian/utils/credential_command.py
```diff
@@ -0,0 +1,187 @@
...
+            resolved_credentials: dict[str, str] = {}
+
+            for command_var in _SERVICE_COMMANDS[service]:
+                target_var = COMMAND_ENV_MAP[command_var]
+                command = os.getenv(command_var)
+                if not command or os.getenv(target_var):
+                    continue
+
+                try:
+                    arguments = shlex.split(command, posix=os.name != "nt")
```
sooperset: **BLOCKER:** `shlex.split(..., posix=False)` leaves surrounding quotes in each token on Windows. The documented `op read "op://Vault/Jira PAT/credential"` example becomes an argument containing literal quote characters, and quoted executable paths are quoted again by `subprocess`. Required for merge: use a parsing representation that yields quote-free argv on Windows and cover a spaced executable or path in a deterministic test or Windows CI.

sooperset (changes requested):
Thanks for the contribution @fepape-ionos! Moving resolution into `asyncio.to_thread` and serializing it per service handles concurrent first calls cleanly.

1. Keep credential helpers away from the MCP transport's stdin.
2. Parse quoted command arguments correctly on Windows.

src/mcp_atlassian/utils/credential_command.py
sooperset, line 191: **BLOCKER:** This child inherits the server's stdin because `stdin` isn't set. In stdio mode, a credential command that reads stdin races the MCP transport. I reproduced a JSON-RPC `tools/list` message becoming `JIRA_API_TOKEN`. Required for merge: isolate the child from protocol input, for example with `stdin=subprocess.DEVNULL`, and add a regression proving that it can't consume MCP stdin.

src/mcp_atlassian/utils/credential_command.py
sooperset: **IMPORTANT:** `shlex.split(..., posix=False)` keeps surrounding quotes. For the documented `op read "op://Vault/Jira PAT/credential"` form, the Windows argv contains the quote characters, so 1Password receives the wrong reference. Required for merge: use a Windows-safe argv representation or parser and cover quoted paths on Windows.

src/mcp_atlassian/servers/main.py
```diff
@@ -157,12 +166,25 @@ async def main_lifespan(app: FastMCP[MainAppContext]) -> AsyncIterator[dict[str,
...
+                logger.error(
+                    f"Failed to load Confluence configuration: {e}", exc_info=True
+                )
         except Exception as e:
             logger.error(f"Failed to load Confluence configuration: {e}", exc_info=True)
 
     app_context = MainAppContext(
         full_jira_config=loaded_jira_config,
         full_confluence_config=loaded_confluence_config,
+        has_deferred_jira_auth=(loaded_jira_config is None and has_deferred_jira_auth),
```
sooperset: **IMPORTANT:** Tying deferred mode to `loaded_jira_config is None` silently drops a higher-priority deferred PAT when Server/DC Basic auth or OAuth can already load. I reproduced `JIRA_PERSONAL_TOKEN_COMMAND` plus valid Basic auth selecting `basic` and never running the PAT command, while a direct `JIRA_PERSONAL_TOKEN` correctly wins in `JiraConfig.from_env`. Required for merge: preserve the existing PAT > OAuth > Basic precedence for deferred Jira and Confluence credentials, with regression tests.

sooperset (changes requested):
Thanks for the contribution @fepape-ionos! The first-use design is useful, and the atomic retry plus HTTP fallback coverage is a strong foundation.

1. Isolate credential commands from MCP stdio.
2. Make quoted command arguments work on Windows.
3. Preserve Server/DC PAT precedence for deferred credentials.

src/mcp_atlassian/utils/credential_command.py
fepape-ionos: Fixed in 71e00ed. `subprocess.run` now passes `stdin=subprocess.DEVNULL`.

Regression test: `TestResolve::test_command_cannot_consume_stdin` puts a real readable pipe holding a JSON-RPC frame on file descriptor 0 (with `os.dup`/`os.dup2` restore) and runs a child that echoes `sys.stdin.read() or 'ISOLATED'`. It asserts the resolved token is `ISOLATED` **and** that the frame is still readable afterwards. The pipe matters: under pytest's default capture fd 0 is already `/dev/null`, so a test without it would pass even with the bug. Removing the `stdin` line makes it fail with the token equal to `{"jsonrpc":"2.0","id":1,"method":"tools/list"}` — your reproduction. `test_custom_timeout` also pins the kwarg.

Confirmed against a live stdio server with `JIRA_API_TOKEN_COMMAND='head -n 1'`: `initialize` and `tools/list` are both answered, 58 tools listed, exit code 0, and the tool call fails with `returned empty output` instead of desyncing the transport.

src/mcp_atlassian/utils/credential_command.py
fepape-ionos: Fixed in 71e00ed. Parsing moved into `_parse_command`, which no longer uses `shlex` on Windows at all — it returns the raw command string, and `subprocess.run` hands a `str` straight to `CreateProcess` as `lpCommandLine` with `lpApplicationName=NULL`, so native argument rules apply. `op read "op://Vault/Jira PAT/credential"` reaches 1Password verbatim, and a quoted executable path is no longer re-quoted by `list2cmdline`. POSIX keeps `shlex.split(..., posix=True)`. `shell=False` in both cases.

I chose this over `posix=False` plus stripping surrounding quotes because `shlex` does not implement MSVCRT rules — embedded and adjacent quotes, `--flag="a b"`, and backslashes all differ — so stripping would be a guess that breaks on any token legitimately containing a quote.

Since all six workflows are `ubuntu-latest`, the platform is an injected keyword (`is_windows`, defaulting to `os.name == "nt"`) rather than something tests patch: patching `os.name` mutates the shared `os` module process-wide. `TestParseCommand` covers the 1Password reference, a quoted `C:\Program Files\...\op.exe` path with spaces, the POSIX split, empty commands on both branches, and unbalanced quotes.

Two pre-existing Windows caveats are now documented: `CreateProcess` with `lpApplicationName=NULL` searches the current directory before `PATH`, and `.bat`/`.cmd`/PowerShell helpers need an explicit `cmd /c` or `powershell -NoProfile -Command` prefix.

src/mcp_atlassian/servers/main.py
fepape-ionos: Fixed in 71e00ed, though not by the most direct route — discarding `loaded_jira_config` would have introduced a worse regression.

`_get_global_config` raises when `full_jira_config` is `None`, and HTTP Branch 2 (user Basic) and Branch 3 (user bearer) both call it purely to inherit URL, SSL, proxy, and custom headers. Dropping the config would turn every authenticated multi-user request into a hard failure — and no existing test would have caught it, since they all set `full_jira_config` explicitly.

So the config is retained and the *meaning* of the flag changed to "a pending command matches or outranks what loaded":

```python
deferred_jira_wins = has_deferred_jira_auth and (
    loaded_jira_config is None
    or deferred_pat_outranks(
        "jira",
        is_cloud=loaded_jira_config.is_cloud,
        auth_type=loaded_jira_config.auth_type,
    )
)
```

`deferred_pat_outranks` returns `True` only for Server/DC when the loaded `auth_type` is not `pat` and `*_PERSONAL_TOKEN_COMMAND` is set without a static `*_PERSONAL_TOKEN` — mirroring the PAT > OAuth > Basic order in `from_env` (#824). In `_get_fetcher` the deferred block now runs before the global-config block; if the command fails and a config did load, it logs a warning and uses that config, so a locked secret store can't take down a deployment that has working static credentials.

The `ALLOW_GLOBAL_CRED_FALLBACK` guard is untouched and still precedes both, so unauthenticated HTTP callers still cannot trigger the command.

Behaviour verified end to end: Server/DC with static Basic plus `JIRA_PERSONAL_TOKEN_COMMAND` loads `basic` at startup, resolves on first tool call, and builds the fetcher with `auth_type=pat`. Cloud with OAuth plus `*_API_TOKEN_COMMAND` keeps OAuth and never runs the command — matching static behaviour, and worth preserving since these commands can trigger biometric prompts.

New tests: two lifespan cases in `test_mcp_protocol.py`, a `deferred_pat_outranks` table, and four in `test_dependencies.py` including one that pins per-user HTTP auth working alongside a pending command.

src/mcp_atlassian/utils/credential_command.py
fepape-ionos: Same fix as the thread above: `stdin=subprocess.DEVNULL` in 71e00ed, with `TestResolve::test_command_cannot_consume_stdin` proving a helper cannot read protocol input.

src/mcp_atlassian/utils/credential_command.py
fepape-ionos: Same fix as the thread above: Windows no longer goes through `shlex` in 71e00ed — the raw command string is passed to `CreateProcess`, so the `op read "op://..."` form arrives quote-free. Covered by `TestParseCommand` via an injected platform flag.

fepape-ionos:
Rebased onto `main` (`ec54351`) to clear the conflicts. Two files needed manual resolution.

**`servers/dependencies.py`** — `main` had since added an `auth_type == "external"` exemption to the unauthenticated-HTTP fallback guard, while this branch had hoisted that guard above the `global_config_fallback` branch so it also covers deferred credentials. Merged so both properties hold:

```python
operator_credentials_at_risk = has_deferred_fallback or (
    global_config_fallback is not None
    and global_config_fallback.auth_type != "external"
)
```

External auth mode is exempt because it holds no operator credentials — the caller supplies them per request. A pending `*_COMMAND` is never exempt, since running it produces exactly the credentials the guard protects. A naive resolution either way would have regressed one of the two; `test_external_auth_exemption_does_not_cover_deferred_commands` now pins the interaction, which neither side covered alone. The existing `TestGetFetcherExternalAuth` cases still pass unchanged, which is the check that the exemption survived.

**`tests/unit/servers/test_dependencies.py`** — both sides added tests at the same location; kept both.

Re-verified after the rebase: full unit suite 3773 passed / 5 skipped, pre-commit (Ruff, formatting, mypy) clean, the `head -n 1` stdio probe still answers `initialize` and `tools/list` and exits 0 (63 tools now), and the Server/DC probe still resolves `basic` → `pat` on first use.

fepape-ionos:
@sooperset: I worked on the requested changes, updated the branch again, fixed the one non green action and resolved conflicts which would occur on merging.
Any further change requests?

PR #1247, opened 2026-04-09
feat(jira): add jira_search_projects tool using projects/picker

### LINKED ISSUE #1246: [Feature]: Add jira_search_projects tool using /rest/api/2/projects/picker
## Problem

The only way to find a project today is `get_all_projects`, which returns every visible project on the instance. On large Jira deployments this means pulling back hundreds of projects when an agent just needs to resolve a name or partial key.

There's a dedicated Jira endpoint for exactly this: [`GET /rest/api/2/projects/picker`](https://developer.atlassian.com/server/jira/platform/rest/v10002/api-group-projects/#api-api-2-projects-picker-get). It accepts a `query` parameter (name or key prefix) and returns a small ranked list of matches.

## Proposed: `jira_search_projects` tool

| Parameter | Type | Default | Description |
|---|---|---|---|
| `query` | string (required) | — | Name or key prefix to search for |
| `max_results` | int | `20` | Max number of results to return |
| `current_project_ids` | list[str] | `None` | Project IDs to exclude from results (optional) |

## Implementation notes

- The upstream `jira` library doesn't wrap this endpoint, so it'll need a direct HTTP call via `self.jira._session`
- Should respect `JIRA_PROJECTS_FILTER` config

LukePeltier:
## Summary
- Adds `jira_search_projects` tool that wraps `GET /rest/api/2/projects/picker`, allowing agents to find projects by name or key prefix without fetching every visible project
- Accepts `query` (required), `max_results`, and `current_project_ids` parameters
- Respects `JIRA_PROJECTS_FILTER` configuration

Closes #1246

## Test plan
- [x] Unit tests for `search_projects()` mixin method (7 tests)
- [x] Server-level tool tests for `jira_search_projects` (5 tests)
- [x] Updated tool count assertion in toolset completeness test
- [x] Full unit suite passes (2595 passed)
- [x] Pre-commit checks pass (ruff-format, ruff, mypy)

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,11 +1,5 @@
 {
   "permissions": {
-    "allow": [
-      "Edit:*",
-      "Write:*",
-      "MultiEdit:*",
-      "Bash:*"
-    ],
-    "defaultMode": "acceptEdits"
+    "defaultMode": "auto"
   }
 }
--- a/.env.example
+++ b/.env.example
@@ -183,6 +183,17 @@ MCP_VERY_VERBOSE=true   # Enables DEBUG level logging (equivalent to 'mcp-atlass
 # Optional: Comma-separated list of Jira project keys to limit searches and other operations to.
 #JIRA_PROJECTS_FILTER=PROJ,DEVOPS
 
+# --- Confluence Cloud attachment download workaround ---
+# Confluence Cloud removed the legacy /download/attachments/... endpoint that an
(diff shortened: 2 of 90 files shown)


LukePeltier:
@sooperset any input on this?

sooperset:
Luke, the Cloud/DC endpoint split is the key bit here, and the focused e2e coverage gives me enough confidence to land it. Merged.

PR #1261, opened 2026-04-13
feat(jira): add jira_unlink_from_epic tool

### LINKED ISSUE #1221: [Feature]: Support unlinking an issue from an epic (jira_unlink_from_epic or allow null epic_key)
## Description

There is currently no way to remove/unlink an issue from an epic using the MCP tools. `jira_link_to_epic` is strictly link-only, and all alternative approaches via `jira_update_issue` silently fail.

## Tested Approaches (all failed)

| Approach | Result |
|---|---|
| `jira_link_to_epic(issue_key="EN-5339", epic_key="")` | Pydantic validation error -- `epic_key` is required and must match `ISSUE_KEY_PATTERN` |
| `jira_link_to_epic(issue_key="EN-5339", epic_key=None)` | Same -- cannot pass null for a required `str` parameter |
| `jira_update_issue(additional_fields={"parent": null})` | **Silent fail** -- returns "Issue updated successfully" but parent unchanged |
| `jira_update_issue(additional_fields={"epicKey": ""})` | **Silent fail** -- same |
| `jira_update_issue(additional_fields={"customfield_10006": null})` | **Silent fail** -- same |

The silent failures are particularly problematic -- the tool reports success, giving no indication that the field was not actually cleared.

## Expected Behavior

One of the following should work:

1. **Preferred:** A new `jira_unlink_from_epic(issue_key)` tool, or
2. Allow `jira_link_to_epic` to accept an optional/nullable `epic_key` that clears the link when null/empty, or
3. `jira_update_issue` with `parent: null` or `customfield_10006: null` should actually clear the field (or return an error if it can't)

## Jira REST API Reference

The Jira Cloud REST API supports unlinking from an epic by setting the parent field to null:

```
PUT /rest/api/3/issue/{issueIdOrKey}
{"fields": {"parent": null}}
```

So the underlying API supports this -- it's the MCP tool layer that doesn't expose it.

## Current Workaround

The only option is to remove the epic link manually in the Jira web UI.

## Environment

- mcp-atlassian (via Docker MCP Toolkit)
- Jira Cloud
- Jira project using next-gen/team-managed hierarchy (parent field + Epic Link custom field `customfield_10006`)

## Related Issues

- #623 -- `jira_create_issue` won't take an epic key (inverse problem -- setting epic at create time; same root cause of incomplete epic/parent field handling)
- #771 -- `update_issue`/`create_issue` silently failing to set parent field (same silent-fail pattern, closed as stale)
- #454 -- Feature request to dynamically detect Epic Link custom field
- #324 -- `link_to_epic` attribute error (fixed, but relevant prior art)

MaxRyabov:
## Summary

There was no reliable MCP operation for removing an issue from an Epic. Passing `null` through `jira_update_issue` could be dropped before reaching Jira, while guessing common custom-field IDs could clear an unrelated field.

This PR adds `jira_unlink_from_epic`, a dedicated write tool that safely removes the relationship and returns the updated issue.

Closes #1221

## Implementation

- Handles Cloud parent relationships, including localized Epic names through Jira's hierarchy metadata.
- Uses Jira Cloud's v3 issue endpoint and the Jira Software removal endpoint when needed.
- Uses a custom Epic Link field only after Jira field metadata verifies its purpose.
- Uses the native Jira v2 issue endpoint for Server/Data Center custom fields.
- Refuses to clear non-Epic parents, so sub-tasks are not detached from Story/Task parents.
- Confirms custom-field removal before reporting success and accepts Cloud's omitted `fields` object when an unset parent is the only requested field.
- Registers and documents the new MCP tool under `toolset:jira_links` with the existing write-access guard.

## Verification

- `pre-commit run --all-files` — passed (Ruff formatting, Ruff, mypy)
- `uv run pytest tests/unit/jira/test_epics.py tests/unit/utils/test_toolsets.py -xq` — 67 passed
- `uv run pytest tests/integration/ -xq` — 23 existing real-API tests skipped by their opt-in guards
- Cloud unlink E2E — passed against the configured Jira Cloud instance
- Server/DC unlink E2E — passed against a temporary Scrum project in the DC harness
- GitHub CI — lint and tests passed on Python 3.10, 3.11, 3.12, and 3.13


src/mcp_atlassian/jira/epics.py
```diff
@@ -428,6 +430,198 @@ def link_issue_to_epic(self, issue_key: str, epic_key: str) -> JiraIssue:
...
+
+        field = self.get_field_by_id(field_id)
+        if not field:
+            return None
+
+        field_name = str(field.get("name", "")).strip().lower()
+        schema = field.get("schema")
+        custom_type = schema.get("custom", "") if isinstance(schema, dict) else ""
+        if (
+            field_name == "epic link"
```
sooperset: **BLOCKER:** This treats the mutable display name `Epic Link` as proof that a custom field owns the Epic relationship. Jira allows duplicate and renamed custom fields, so a normal field with that name can be selected by `get_field_ids_to_epic()` and then cleared, losing unrelated data. The helper can also receive a guessed `customfield_10014` instead of the real schema-backed field, which later reports success without clearing the relationship. Required for merge: scan the available fields for `schema.custom == "com.pyxis.greenhopper.jira:gh-epic-link"`, or an equally stable identifier, and add regression cases for both collisions.

sooperset (changes requested):
Thanks for the contribution @MaxRyabov! The Cloud/DC endpoint split and real E2E coverage are exactly the right shape here.

1. Identify the Epic Link field from its stable schema type, not the current heuristic candidate or display name, so the tool can't clear an unrelated same-named field or miss the real field.

src/mcp_atlassian/jira/epics.py
MaxRyabov: Fixed in d8cffef — thanks for catching this.

`_get_verified_epic_link_field_id()` no longer consults the heuristic candidate from `get_field_ids_to_epic()` or the display name at all. It now scans the instance's field metadata and returns the custom field whose `schema.custom == "com.pyxis.greenhopper.jira:gh-epic-link"`, which is stable on both Cloud and Server/DC. A renamed or duplicated field called "Epic Link" can no longer be selected, and a guessed `customfield_10014` is never treated as the Epic relationship.

Regression tests cover both collisions:
- `test_get_verified_epic_link_field_id_rejects_same_named_field` — a duplicate field named "Epic Link" without the GreenHopper schema is rejected;
- `test_get_verified_epic_link_field_id_rejects_guessed_id` — `customfield_10014` with an unrelated schema type is rejected;
- `test_unlink_issue_from_epic_does_not_clear_unrelated_fields` — end-to-end: no update is issued against a same-named field, so its data is never cleared;
- `test_get_verified_epic_link_field_id_accepts_renamed_field` — the real Epic Link field is still found when its display name was changed.

Also merged `upstream/main` to resolve the conflicts.

PR #1263, opened 2026-04-14
feat(jira): add saved filter tools (my filters, favourites, filter by ID)

anshu-makkar:
## Description

Add support for retrieving Jira saved filters via new MCP tools. Users currently have no way to list or inspect their saved Jira filters through the MCP interface — this fills that gap.

## Changes

- Add `JiraFilter` Pydantic model with API parsing and simplified serialization
- Add `FiltersMixin` with methods for owned filters, favourite filters, and filter lookup by ID
- Register three read-only MCP tools: `get_my_filters`, `get_favourite_filters`, and `get_filter_by_id`
- Add the `jira_filters` toolset and wire the mixin into `JiraFetcher`
- Handle null required fields safely and cover MCP tool behavior with regression tests

## Testing

- [x] Unit tests: 3,297 passed, 5 skipped
- [x] Focused saved-filter/server/toolset tests: 185 passed
- [x] Integration suite collected: 23 environment-gated tests skipped
- [x] Jira Cloud e2e: 89 passed, 15 skipped
- [x] Jira Data Center e2e: 99 passed, 105 skipped
- [x] Pre-commit hooks: passed

## Checklist

- [x] Code follows project style guidelines
- [x] Unit and live compatibility verification passed
- [x] Contributor credit preserved

### DIFF
--- a/src/mcp_atlassian/jira/filters.py
+++ b/src/mcp_atlassian/jira/filters.py
@@ -1,7 +1,6 @@
 """Module for Jira filter operations."""
 
 import logging
-from typing import Any
 
 from requests.exceptions import HTTPError
 
@@ -42,8 +41,9 @@ class FiltersMixin(JiraClient):
         except TypeError:
             raise
         except Exception as e:
-            logger.error(f"Error getting my filters: {str(e)}")
-            raise Exception(f"Error getting my filters: {str(e)}") from e
+            error_msg = f"Error getting my filters: {e!s}"
+            logger.error(error_msg)
+            raise Exception(error_msg) from e
 
     @handle_auth_errors("Jira API")
     def get_favourite_filters(self) -> list[JiraFilter]:
@@ -61,7 +61,9 @@ class FiltersMixin(JiraClient):
             response = self.jira.get("rest/api/2/filter/favourite")
 
             if not isinstance(response, list):
-                msg = f"Unexpected response type from filter/favourite: {type(response)}"
+                msg = (
+                    f"Unexpected response type from filter/favourite: {type(response)}"
+                )
                 logger.error(msg)
                 raise TypeError(msg)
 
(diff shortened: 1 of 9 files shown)


src/mcp_atlassian/jira/filters.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Returns:
+            List of JiraFilter model instances.
+
+        Raises:
+            MCPAtlassianAuthenticationError: If authentication fails.
+            Exception: If there is an error retrieving filters.
+        """
+        try:
+            response: Any = self.jira.get("rest/api/2/filter/my")
```
sooperset, line 39: This endpoint is Cloud-only (404 on Server/DC), so guard before the request, matching the `self.config.is_cloud` pattern in `field_options.py`:

```suggestion
        if not self.config.is_cloud:
            raise ValueError(
                "Listing filters owned by the current user is only supported on "
                "Jira Cloud. On Jira Server/Data Center, use get_favourite_filters "
                "or get_filter_by_id instead."
            )

        try:
            response: Any = self.jira.get("rest/api/2/filter/my")
```

The `ValueError` surfaces as a clean `Configuration Error: …` through the server tool. Worth a unit test asserting `jira.get` is never called when `is_cloud` is `False`, since the e2e suite doesn't exercise these tools directly.

sooperset (commented):
Thanks for digging in here, @anshu-makkar. I traced the `filter/my` failure against a live DC 9.x instance — root cause is that `rest/api/2/filter/my` is Cloud-only and 404s on Server/DC. `get_favourite_filters` and `get_filter_by_id` both work there, so only `get_my_filters` needs gating; there's no owner-based DC fallback (`filter/search` 404s too, and `favourite` is starred, not owned). Fix + a test suggestion inline. Verified locally against DC + Cloud with `test_filters.py` green.

PR #1151, opened 2026-03-10
feat(confluence): add get/add inline comment tools

SmallChX:
## Description

Adds two new MCP tools — `confluence_get_inline_comments` and `confluence_add_inline_comment` — and the backing `CommentsMixin` methods to fetch and create inline comments anchored to a text selection on a Confluence page.

The existing `confluence_add_comment` tool only supports footer (page-level) comments. There was no way to create inline comments through the MCP server. This is a commonly requested capability since inline comments are widely used for contextual feedback in Confluence.

The root cause is that `POST /rest/api/content/` (v1) does not support inline comment creation on Confluence Cloud regardless of auth method — the v2 API is required. A dedicated `_inline_v2_adapter` property activates the v2 path based on `is_cloud` (not `auth_type`), so basic auth and OAuth users on Cloud both benefit.

Fixes: #715 

## Changes

- `confluence/comments.py` — `_inline_v2_adapter` property; `get_inline_comments()` and `add_inline_comment()` methods with Cloud (v2) and Server/DC (v1) paths
- `confluence/v2_adapter.py` — `get_inline_comments()`, `create_inline_comment()`, `_convert_v2_inline_comment_to_v1_format()`; timestamp fields (`created`/`updated`) mapped from `version.createdAt` in both v2 comment converters
- `servers/confluence.py` — `confluence_get_inline_comments` and `confluence_add_inline_comment` tools registered with `toolset:confluence_comments` tags
- `tests/fixtures/confluence_mocks.py` — `MOCK_INLINE_COMMENT_V1_RESPONSE`, `MOCK_INLINE_COMMENT_V2_RESPONSE`, `MOCK_INLINE_COMMENTS_V2_LIST_RESPONSE`
- `tests/unit/confluence/test_comments.py` — `comments_mixin_dc` fixture; 10 new tests covering v1/v2 routing, empty response, network error, HTML format, model parsing
- `tests/unit/utils/test_toolsets.py` — updated Confluence tool count (24 → 26)

## Testing

- [x] Unit tests added/updated (10 new tests: v1/v2 routing, empty response, network error, HTML format, model parsing)
- [ ] Integration tests passed
- [x] Manual checks performed: tested `confluence_add_inline_comment` and `confluence_get_inline_comments` end-to-end against a live Confluence Cloud instance using basic auth (API token). Inline comment created successfully; response returned correct `id`, `body`, `created`, and `updated` fields.

## Checklist

- [x] Code follows project style guidelines (linting passes).
[...]

PR #1264, opened 2026-04-14
feat(confluence): add inline comment tools with Server/DC v1 payload fix

artdaal:
## Summary

Adds `confluence_get_inline_comments` and `confluence_add_inline_comment` MCP tools, backed by new `CommentsMixin` methods. Builds on @SmallChX's work in #1151 (preserved as-is via co-authored commit) and adds a second commit that completes the v1/Server-DC payload so the tool actually creates comments against real Confluence Data Center instances.

Opening this as a parallel PR because #1151 has been awaiting review since 2026-03-10 and the Server/DC code path there returns HTTP 400 as written — see detailed findings in [my review comment on #1151](https://github.com/sooperset/mcp-atlassian/pull/1151#issuecomment-4213653405). Happy to close this in favor of #1151 if @SmallChX folds the fix in, or to have it superseded by #1151 — whichever is easier for maintainers. Full credit to @SmallChX for the original feature work.

## What the fix adds on top of #1151

Confluence Data Center (verified on 8.5.30, build 19293) rejects the v1 `inlineProperties` payload unless four additional fields are present:

| Field | Type | Source |
|---|---|---|
| `numMatches` | int | existing `text_selection_match_count` param |
| `matchIndex` | int | existing `text_selection_match_index` param |
| `lastFetchTime` | string (epoch ms) | generated from wall clock |
| `serializedHighlights` | string (JSON-encoded nested array `[[anchor]]`) | derived from `text_selection` |

The public tool API does not change — the two existing int parameters are now forwarded to `numMatches`/`matchIndex`, and the two generated fields are internal.

Empirically verified failure modes this fix avoids:
- `serializedHighlights: null | \"\"` → HTTP 400 validation error
- `serializedHighlights: []` (real JSON array, not stringified) → HTTP 500 (Jackson deserializer expects a `String`)
- `serializedHighlights: \"[]\"` → HTTP 400 \"Can not create inline comment\"
- `serializedHighlights: \"[[<anchor>]]\"` → HTTP 200 ✅

`GET /rest/api/content/{id}/child/comment?expand=extensions.inlineProperties` only returns `originalSelection` + `markerRef`, so these field formats had to be reverse-engineered against the POST endpoint directly.

The alternative `/rest/inlinecomments/1.0/comments` endpoint does not exist on DC 8.5 (404), so `/rest/api/content/` with the fuller payload is the only working v1 path.

## Commits

1. `47db05c` — **feat(confluence): add get/add inline comment tools** (@SmallChX, from #1151, unchanged)
2. `3481b4a` — **fix(confluence): complete v1 inline comment payload for Server/DC** (this PR)

## Test plan

- [x] All 2601 existing unit tests pass
- [x] `ruff check` / `ruff format` clean
- [x] `mypy` clean
- [x] `test_add_inline_comment_v1_success` extended with assertions for `numMatches`, `matchIndex`, `lastFetchTime`, `serializedHighlights`
- [x] New regression test `test_add_inline_comment_v1_forwards_match_count_and_index` verifying parameter forwarding
- [x] Manually validated against Confluence Data Center 8.5.30 (build 19293) — inline comment creates and anchors correctly
- [ ] Cloud / v2 path unchanged (covered by @SmallChX's existing tests)

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,11 +1,5 @@
 {
   "permissions": {
-    "allow": [
-      "Edit:*",
-      "Write:*",
-      "MultiEdit:*",
-      "Bash:*"
-    ],
-    "defaultMode": "acceptEdits"
+    "defaultMode": "auto"
   }
 }
--- a/.env.example
+++ b/.env.example
@@ -183,6 +183,17 @@ MCP_VERY_VERBOSE=true   # Enables DEBUG level logging (equivalent to 'mcp-atlass
 # Optional: Comma-separated list of Jira project keys to limit searches and other operations to.
 #JIRA_PROJECTS_FILTER=PROJ,DEVOPS
 
+# --- Confluence Cloud attachment download workaround ---
+# Confluence Cloud removed the legacy /download/attachments/... endpoint that an
+# attachment's `_links.download` still points to; it now returns 401 for
+# API-token / scoped-token auth (metadata endpoints keep working). Controls
+# whether attachment/image downloads use the v1 REST endpoint
+# (/rest/api/content/{id}/child/attachment/{aid}/download) instead:
+#   unset (default) -> auto: v1 on Cloud, legacy link on Server/DC
+#   true            -> always v1
+#   false           -> always the legacy link
+#CONFLUENCE_ATTACHMENT_DOWNLOAD_USE_V1=true
+
 # --- SLA Metrics Configuration ---
 # Configure how SLA metrics are calculated for Jira issues.
 # Default metrics to calculate (comma-separated).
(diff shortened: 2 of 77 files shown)


artdaal:
Hi @sooperset — friendly bump on this one. The PR is still relevant and actively used against our Confluence DC 8.5.30 instance; the v1 payload fix is the only thing keeping inline comments working there. Could you take a look when you get a chance? Happy to rebase onto current `main` and address any review feedback. Thanks!

sooperset:
Thanks @artdaal, the Server/DC payload notes and regression coverage around `serializedHighlights`, `numMatches`, and `matchIndex` made this much easier to trust. I also appreciate that you preserved credit for the original #1151 work while tightening the DC path.

PR #1265, opened 2026-04-15
ci: add TruffleHog secret scanning workflow

pmoros:
## Summary

Adds automated secret detection to CI, enforcing the `SECURITY.md` guidance that credentials must never be committed to version control.

- Adds a `Security` workflow for pull requests targeting `main` and pushes to `main`
- Checks out the full repository history with persisted credentials disabled
- Pins both the checkout and TruffleHog actions to immutable commit SHAs
- Reports verified and unknown findings so unverified potential leaks are reviewed instead of silently ignored

## Test plan

- [x] Run the full unit test suite
- [x] Run pre-commit checks, including YAML validation, Ruff, and mypy
- [x] Verify the `Security / Secret Scanning` job runs successfully on the PR


sooperset:
This adds a focused secret-scanning gate with read-only permissions and checks both verified and unknown findings. Thanks for getting the TruffleHog integration started, @pmoros.

PR #1273, opened 2026-04-16
feat(jira): add jira_move_issue tool for cross-project moves

maucher:
## Summary

Adds a `jira_move_issue` MCP tool that moves a Jira Cloud issue to a different project using the Jira Cloud bulk move API.

The implementation preserves the source issue type by resolving the destination issue type ID in the target project, submits Jira's documented `targetToSourcesMapping` payload to the bulk move endpoint, polls the Cloud bulk queue until completion, and then returns the moved issue. Jira may assign a new key in the target project.

The tool is Cloud-only; calling it against a Server/DC instance raises a clear `NotImplementedError`. Write access is gated via `@check_write_access` so `READ_ONLY_MODE=true` blocks it at the server level. The tool is annotated with `destructiveHint=True` to signal its irreversible nature.

## Changes

- `src/mcp_atlassian/jira/issues.py` - new `move_issue(issue_key, target_project_key)` method on `IssuesMixin`
- `src/mcp_atlassian/servers/jira.py` - new `jira_move_issue` MCP tool
- `docs/` and toolset metadata - document the new tool and update the public tool count
- Unit and e2e coverage for the Jira mixin, MCP server tool, Cloud move path, DC unsupported guard, and tool count

## Testing

- `uv run pytest tests/unit/jira/test_issues.py::TestMoveIssue tests/unit/jira/test_projects.py tests/unit/servers/test_jira_server.py::test_jira_move_issue_success tests/unit/servers/test_jira_server.py::test_jira_move_issue_failure tests/unit/servers/test_jira_server.py::test_jira_move_issue_not_cloud tests/unit/utils/test_toolsets.py::TestToolsetTagCompleteness::test_jira_tool_count -xvs` - 77 passed
- `uv run pytest tests/integration/test_real_api.py::TestRealJiraAPI -xvs` - 7 skipped; integration suite credentials were not configured
- `uv run pytest tests/e2e/cloud/test_confluence_cloud_operations.py::TestConfluenceCloudComments::test_add_and_get_inline_comments --cloud-e2e -xvs` - 1 passed after a transient Confluence 404 in the first full Cloud run
- `uv run pytest tests/e2e/cloud --cloud-e2e -xvs` - 70 passed, 15 skipped
- `uv run pytest tests/e2e/test_jira_dc_operations.py tests/e2e/test_jira_auth_matrix.py tests/e2e/test_mcp_tools_dc.py --dc-e2e -xvs` - 44 passed, 1 skipped
- `uv run pre-commit run --all-files` - passed

## Checklist

- [x] Code follows project style guidelines (linting passes)
- [x] Unit, Cloud e2e, and DC e2e verification completed

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -1,6 +1,7 @@
 """Module for Jira issue operations."""
 
 import logging
+import time
 from collections import defaultdict
 from typing import Any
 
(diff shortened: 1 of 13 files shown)


maucher:
Hi @sooperset - would you be able to review this PR when you get a chance? This would unblock my team from being able to move Jira issues across projects via the MCP. Thanks!

src/mcp_atlassian/jira/issues.py
```diff
@@ -1445,6 +1446,198 @@ def delete_issue(self, issue_key: str) -> bool:
...
+                        "inferFieldDefaults": True,
+                        "inferStatusDefaults": True,
+                        "inferSubtaskTypeDefault": True,
+                        "issueIdsOrKeys": [issue_key],
+                    }
+                },
+            }
+
+            response = self._post_api3("bulk/issues/move", data)
+
```
sooperset, line 1504: **BLOCKER:** The second segment of this mapping key is the target project ID, but Jira's bulk move schema expects `<project ID or key>,<issueType ID>,<parent ID or key>`. For a normal cross-project move, project ID and issue type ID won't match, so Cloud will reject the move or route it to the wrong target type.

Required for merge: resolve the destination issue type ID before building this key, for example by preserving the source issue type in the target project or adding an explicit target issue type input. Then update the unit test to use different project and issue type IDs so this can't regress.

sooperset (changes requested):
Thanks for the contribution @maucher! The Cloud-only guard, read-only gating, and async polling shape are all good fits for this repo.

1. The bulk move payload is using the target project ID where Jira expects the destination issue type ID.
2. The unit test currently locks in that wrong shape, so it passes while the real Cloud API call would fail.

sooperset:
Merged this, @maucher. Resolving the destination issue type before building `targetToSourcesMapping` is the key fix here, and the Cloud/DC tests cover the paths I was worried about.

PR #1275, opened 2026-04-16
feat(confluence): add content_file to create/update page tools

nlwstein:
## Description

Large Confluence page bodies sent through the MCP stdio transport as a
single JSON argument can stall an MCP client or exceed buffer limits. This
PR adds a `content_file` parameter to `confluence_create_page` and
`confluence_update_page` so callers can hand off the body via the
filesystem instead. This mirrors the `file_path` pattern already used by
`upload_attachment`.

## Changes

- `confluence_create_page` and `confluence_update_page` each gain a new
  `content_file: str | None` parameter. The existing `content` parameter
  becomes optional (`str | None`, defaults to `None`).
- New `_resolve_page_content()` helper validates that exactly one of
  `content` / `content_file` is supplied and reads UTF-8 from disk.
  Violations raise `ValueError` with a clear message.
- Tool descriptions are updated so agents know when to reach for
  `content_file`.
- Existing call sites that pass `content` (positionally or by keyword)
  are unchanged.

## Testing

- [x] Unit tests added (5 new):
  - `test_create_page_with_content_file` — file body flows to the fetcher
  - `test_update_page_with_content_file` — same, for update
  - `test_create_page_rejects_both_content_and_file`
  - `test_create_page_rejects_neither_content_nor_file`
  - `test_update_page_rejects_missing_file`
- [x] Full suite: `PYTHONPATH=. pytest tests/unit/servers/test_confluence_server.py`
  → 31 passed (26 existing + 5 new), no regressions.
- [x] Manual smoke test against a live Confluence Cloud instance:
  create from file → v1, update from file → v2, round-trip read confirmed
  the body on the server matched the file. Error paths (both args, neither
  arg) also surfaced cleanly through the MCP boundary.

## Checklist

- [x] Tests added for changes.
- [x] All tests pass locally.
- [ ] Documentation updated (if needed). *— The tool descriptions are
  updated in-code, which is where the MCP clients read them. No separate
  docs file appears to exist for these tools; happy to add one if a
  maintainer points me at the right place.*

### DIFF
--- a/src/mcp_atlassian/servers/confluence.py
+++ b/src/mcp_atlassian/servers/confluence.py
@@ -51,13 +51,9 @@ def _resolve_page_content(content: str | None, content_file: str | None) -> str:
     has_content = content is not None and content != ""
     has_file = content_file is not None and content_file != ""
     if has_content and has_file:
-        raise ValueError(
-            "Provide either 'content' or 'content_file', not both."
-        )
+        raise ValueError("Provide either 'content' or 'content_file', not both.")
     if not has_content and not has_file:
-        raise ValueError(
-            "One of 'content' or 'content_file' must be provided."
-        )
+        raise ValueError("One of 'content' or 'content_file' must be provided.")
     if has_content:
         return content  # type: ignore[return-value]
     path = Path(content_file).expanduser()  # type: ignore[arg-type]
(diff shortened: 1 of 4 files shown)


izhyvaiev:
I found this very useful and would like it to get merged. Leaving it here to keep PR alive

sooperset:
The `content_file` addition is well shaped: it keeps existing positional calls intact and covers both empty inline content and file-backed create/update. I also verified the new Cloud and DC MCP e2e paths, so this is ready to merge.

PR #1253, opened 2026-04-10
fix(confluence): complete DC/Server attachment upload — header, POST method, versioning fallback

vu2sar:
## Summary

Fixes three independent bugs that together prevent `confluence_upload_attachment` from working on **Confluence Server / Data Center**. All three must be fixed for uploads to succeed end-to-end.

Closes #1207

---

### Bug 1 — Wrong `X-Atlassian-Token` header value

| | Value |
|---|---|
| Before | `"nocheck"` |
| After | `"no-check"` |

Confluence Server requires the hyphenated form to bypass XSRF validation on multipart uploads. Sending `"nocheck"` (no hyphen) causes the server to return **403 Forbidden** and reject the upload entirely.

Reference: [Atlassian REST API examples — Upload an attachment](https://developer.atlassian.com/server/confluence/confluence-rest-api-examples/#upload-an-attachment)

---

### Bug 2 — Wrong HTTP method (`PUT` → `POST`)

| | Method |
|---|---|
| Before | `PUT /rest/api/content/{id}/child/attachment` |
| After | `POST /rest/api/content/{id}/child/attachment` |

Confluence Server/DC does not support `PUT` on the attachment list endpoint — it returns 404 or 405. `POST` is the correct method for creating a new attachment per the REST API docs.

---

### Bug 3 — No versioning fallback for existing attachments (new fix, not in any open PR)

When a file with the same name already exists, Confluence Server/DC returns **HTTP 400** with `"same file name"` in the body. The previous code had no handling for this case, causing re-uploads to always fail.

**Fix:** Detect the 400 response, look up the existing attachment ID via `GET /child/attachment?filename=<name>`, then `POST` to `/child/attachment/{id}/data` to create a new version.

```python
if response.status_code == 400 and "same file name" in response.text:
    # find existing att_id, then:
    response = session.post(f".../child/attachment/{att_id}/data", ...)
```

This is the fix that makes the tool actually usable day-to-day — without it, you can upload a file once but can never update it.

---

## Relation to Other Open PRs

| PR | Fixes |
|---|---|
| #1202 | Bug 2 only (PUT → POST) |
| #1249, #1217, #1210 | Bug 1 only (header value) |
| **This PR** | **All three bugs, including the versioning fallback** |

This PR supersedes those for Server/DC users. If the maintainers prefer to merge the existing PRs separately, only Bug 3 (versioning fallback) would remain.

---

## Files Changed

| File | Change |
|---|---|
[...]

PR #1276, opened 2026-04-16
fix(confluence): use POST instead of PUT for attachment uploads on Server/DC

flitzrrr:
## Summary

Fixes attachment uploads failing on all Confluence Server/Data Center instances.

### Root cause

`_upload_attachment_direct()` in `confluence/attachments.py` uses `PUT` on the collection URL `/rest/api/content/{id}/child/attachment`. This works on **Confluence Cloud** (where PUT acts as create-or-update), but **Confluence Server/DC requires `POST`** for new attachments. PUT on the collection URL returns an error on Server/DC, which gets swallowed by the generic exception handler, resulting in `"Failed to upload attachment"` for every upload attempt.

### Fix

Changed `PUT` to `POST` in `_upload_attachment_direct()`. POST works on **both** platforms:
- Creates a new attachment if the filename does not exist
- Creates a new version if the filename already exists

### Changed files

- `src/mcp_atlassian/confluence/attachments.py` — `session.put(...)` → `session.post(...)`
- `tests/unit/confluence/test_attachments.py` — updated mock assertions from `.put` to `.post`

### Tested

- 52/52 unit tests pass
- Manually verified on Confluence Server (Telefónica on-premise instance) with PNG uploads

### Confluence REST API reference

- [Server/DC: Add attachment](https://developer.atlassian.com/server/confluence/content-resource/#post-content-id-child-attachment) — **POST** to `/rest/api/content/{id}/child/attachment`
- [Cloud: Create or update attachment](https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-content---attachments/#api-wiki-rest-api-content-id-child-attachment-put) — **PUT** to `/wiki/rest/api/content/{id}/child/attachment`

### DIFF
--- a/src/mcp_atlassian/confluence/attachments.py
+++ b/src/mcp_atlassian/confluence/attachments.py
@@ -484,12 +484,13 @@ class AttachmentsMixin(ConfluenceClient, AttachmentsOperationsProto):
             if minor_edit is not None:
                 data["minorEdit"] = str(minor_edit).lower()
 
-            # Use PUT to support creating new versions of existing attachments
-            # PUT will create a new attachment if it doesn't exist, OR create a new
-            # version if an attachment with the same filename already exists
-            response = self.confluence._session.put(
-                url, headers=headers, files=files, data=data
-            )
+            # Confluence Cloud accepts PUT on the collection URL for both
+            # create and update.  Confluence Server/DC requires POST for new
+            # attachments — PUT on the collection URL returns an error.
+            # POST works on *both* platforms (creates a new version when the
+            # filename already exists), so we use POST unconditionally.
+            method = self.confluence._session.post
+            response = method(url, headers=headers, files=files, data=data)
             response.raise_for_status()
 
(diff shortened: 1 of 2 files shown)


sooperset:
Your diagnosis was right: Server/DC needs POST on the attachment collection endpoint.

I rechecked this against current main. #1253 has since landed with the POST fix, the `X-Atlassian-Token: no-check` fix, Cloud `/wiki` URL handling, and the existing-file versioning fallback. This branch now conflicts with that code, and merging it as-is would roll back those later fixes.

To merge this branch now, it would need to:

1. Rebase onto current `main`.
2. Keep the current attachment upload implementation instead of replacing it with the older POST-only block.
3. Narrow the PR to a remaining bug after #1253, with a regression test.

I ran the focused unit, integration, Cloud e2e, and DC e2e checks against current main, and they pass. Unless you can point to a remaining gap, I'd close this one in favor of #1253.

PR #1281, opened 2026-04-17
fix(jira): use correct API URL for create_version on Server/DC

Mrflatt:
## Description

The create_version method in JiraClient hardcoded /rest/api/3/version which only exists on Jira Cloud. On Jira Server/Data Center, the API uses /rest/api/2/version, causing the call to silently return None and fail with `Unexpected response from Jira API: None`.

Use resource_url('version') to resolve the correct API version per platform automatically (api/2 for Server/DC, api/3 for Cloud).

Fixes: #

## Changes

- Replace hardcoded /rest/api/3/version URL with self.jira.resource_url("version") in JiraClient.create_version

## Testing

- [x] Manual checks performed: `Verified create_version works on Jira Server/DC via MCP tool call (jira_create_version), confirmed version is created successfully with correct name and release date.`

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/client.py
+++ b/src/mcp_atlassian/jira/client.py
@@ -382,7 +382,8 @@ class JiraClient:
         if description:
             payload["description"] = description
         logger.info(f"Creating Jira version: {payload}")
-        result = self.jira.post("/rest/api/3/version", json=payload)
+        url = self.jira.resource_url("version")
+        result = self.jira.post(url, json=payload)
         if not isinstance(result, dict):
             error_message = f"Unexpected response from Jira API: {result}"
             raise ValueError(error_message)


erickalmeidaptc:
**Independent reproduction confirming this PR is correct and the bug affects production Server/DC users.**

**Environment:** mcp-atlassian 0.21.1 via `uvx`, JIRA Data Center, `JIRA_PERSONAL_TOKEN` auth.

**Reproduction:** calling `mcp__mcp-atlassian__jira_create_version` against any Server project returns:
```json
{"success": false, "error": "Unexpected response from Jira API: None"}
```
No version is created. `jira_get_project_versions` confirms the absence. The v3 endpoint returns nothing parseable on Server, so `result is None` fails the `isinstance(result, dict)` check at `mcp_atlassian/jira/client.py:387` (the location this PR touches).

**Impact:** any tooling that depends on `jira_create_version` is broken on Server/DC. The failure mode is a soft envelope (`success: false`), **not an exception** — callers that only check for raised exceptions will think the call succeeded and proceed with a missing version id.

**Workaround we shipped while waiting for this PR to merge:** bypass the MCP tool and call `/rest/api/2/version` directly with the same `JIRA_PERSONAL_TOKEN` already configured for mcp-atlassian. Works, but every Server/DC user has to discover the bug and re-implement the same workaround independently.

**Adjacent observation while investigating:** there is no `jira_update_version` (or archive-version) tool exposed at all — the same v2 PUT/`released:true`/`archived:true` workaround is needed for marking releases. Out of scope for this PR but worth tracking separately.

This PR's `resource_url("version")` approach is the correct fix and matches how other functions in `client.py` already branch on `is_cloud`. +1 from a fresh user — would love to see this un-staled and merged.

sooperset:
Your diagnosis was right: `create_version` was calling a Cloud-only v3 endpoint, and the fix needs to cover Server/DC.

Since this PR sat for a while, the same user-visible bug has already landed in `main`, with a focused unit test and both Cloud and DC e2e coverage passing.

To merge this PR now, I’d need it to become a distinct change over that landed fix:
1. Rebase on current `main`.
2. Keep the current REST v2 behavior unless there’s a concrete reason to switch back to `resource_url("version")`.
3. Add or adjust tests for any behavior that’s different from the landed fix.

Next step: I’m leaning toward closing this as superseded, but I’ll keep it open if you want to retarget it to a separate improvement.

PR #1282, opened 2026-04-17
fix(security): harden OAuth token verification and secret handling

rober15:
## Summary

Hardens the OAuth proxy so opaque upstream tokens are no longer accepted solely because they are non-empty. Secret-bearing CLI flags remain compatible but now direct users toward environment variables.

## What changed

- Validate Atlassian Cloud tokens against the accessible-resources endpoint.
- Bind Cloud validation to the configured site or cloud ID and enforce resource scopes.
- Validate Data Center tokens against authenticated Jira or Confluence current-user endpoints.
- Fail closed on upstream errors, malformed responses, missing resources, and missing scopes.
- Cache successful validation for five minutes using SHA-256 token keys so raw tokens are not retained as cache keys.
- Deprecate `--jira-token`, `--confluence-token`, `--oauth-client-secret`, and `--oauth-access-token` with runtime warnings and descriptive CLI help.

## Validation

- `uv run pytest tests/unit/ -q` — 2578 passed, 5 skipped.
- `uv run pytest tests/integration/ --integration -q` — 8 passed, 15 skipped.
- `uv run pytest tests/e2e/cloud --cloud-e2e -q` — 47 passed, 14 skipped.
- `uv run pytest tests/e2e --dc-e2e --ignore=tests/e2e/cloud -q` — 71 passed, 1 skipped.
- Ruff formatting and lint checks passed for all changed files.
- Mypy passed for all changed source files.
- GitHub lint, Python 3.10–3.13 test, and Docker build checks passed.

### DIFF
--- a/src/mcp_atlassian/__init__.py
+++ b/src/mcp_atlassian/__init__.py
@@ -61,6 +61,37 @@ logging_stream = sys.stdout if is_env_truthy("MCP_LOGGING_STDOUT") else sys.stde
 # Set up logging using the utility function
 logger = setup_logging(logging_level, logging_stream)
 
+_DEPRECATED_SECRET_OPTION_ENV_MAP: dict[str, str] = {
+    "confluence_token": "CONFLUENCE_API_TOKEN",
+    "jira_token": "JIRA_API_TOKEN",
+    "oauth_client_secret": "ATLASSIAN_OAUTH_CLIENT_SECRET",
+    "oauth_access_token": "ATLASSIAN_OAUTH_ACCESS_TOKEN",
+}
+
+
+def _warn_deprecated_secret_options(click_ctx: click.Context | None) -> None:
+    """Warn when deprecated secret-bearing CLI flags are used.
+
+    Secret flags remain supported for compatibility but are discouraged because
+    command-line args may be exposed in shell history/process listings.
+    """
+    if not click_ctx:
+        return
+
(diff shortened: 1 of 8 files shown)


sooperset:
The bounded hash-keyed cache and Cloud site/scope checks resolve the biggest problems in the original version. I can't merge this security-sensitive path with the current Data Center evidence yet.

1. Run the DC e2e suite after both Jira and Confluence finish setup. The current run skips all 99 tests, so it doesn't satisfy the required both-edition tier.
2. Add a live OAuth-proxy verifier check for Cloud, Jira DC, and Confluence DC, covering both a valid token and a rejected invalid or revoked token. The current e2e suite never invokes this verifier, and the unit tests replace the HTTP responses with mocks.

Once those checks execute and pass, rerun CI and I'll review the results.

sooperset:
The live Data Center verifier now checks out for both Jira and Confluence, and the hash-keyed cache plus fail-closed unit coverage look good. Cloud is still the missing security evidence.

1. Add a repeatable Cloud e2e check that passes a valid OAuth access token for the configured site through `AtlassianOpaqueTokenVerifier.verify_token` and confirms acceptance.
2. Run the same path with an invalid or revoked token and confirm rejection. Please include the reviewed head and unskipped result, since the current Cloud run skips all OAuth cases.

Once those 2 checks are in, I'll re-review the final head.

PR #1283, opened 2026-04-18
feat(jira): add saved filter and Structure board tools

trevorh:
## Summary

- Add two new toolsets: **Filters** (`jira_filters`, default: on) and **Structures** (`jira_structures`, default: off)
- Filters: get saved filter by ID, search filters by name (Cloud API + Server/DC favourite-filter fallback), list favourite filters
- Structures: get Almworks Structure board metadata, raw forest hierarchy, and fully resolved issue hierarchy
- Structure resolution routes through `search_issues()` so `JIRA_PROJECTS_FILTER` is respected
- Failed resolution batches produce generic placeholders and `partial=True` metadata instead of silently dropping rows
- Server/DC filter search explicitly marks results as `partial=True` since only favourites are searched

## Test plan

- [ ] 27 unit tests covering both mixins
- [ ] Verify `jira_get_filter` returns filter details with JQL
- [ ] Verify `jira_search_filters` works on Cloud and falls back gracefully on Server/DC
- [ ] Verify `jira_get_structure_issues` resolves hierarchy correctly
- [ ] Verify structure resolution respects `JIRA_PROJECTS_FILTER`

> **Note:** Tools reference docs update will be submitted as a separate PR after feature PRs are merged.

### DIFF
--- a/src/mcp_atlassian/jira/__init__.py
+++ b/src/mcp_atlassian/jira/__init__.py
@@ -17,7 +17,9 @@ from .development import DevelopmentMixin
 from .epics import EpicsMixin
 from .field_options import FieldOptionsMixin
 from .fields import FieldsMixin
+from .filters import FiltersMixin
 from .forms_api import FormsApiMixin  # Forms REST API
+from .structures import StructuresMixin
 from .formatting import FormattingMixin
 from .issues import IssuesMixin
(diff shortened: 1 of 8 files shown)


trevorh:
Thanks for including this!

PR #1284, opened 2026-04-18
feat(jira): add compare_issue_sets tool for JQL set diff

trevorh:
## Summary

- Add `jira_compare_issue_sets` for diffing two JQL result sets
- Return issues only in A, only in B, field differences for shared issues, and the unchanged count
- Paginate Cloud searches internally and Server/Data Center searches by offset
- Compare built-in and custom fields selected by field ID or display name
- Add the opt-in `jira_set_analysis` toolset and generated-tool documentation mapping

## Test plan

- [x] Targeted Jira set analysis, Jira server, and toolset unit tests (181 passed)
- [x] Full unit suite (3,253 passed, 5 skipped)
- [x] Integration suite collected (23 credential-gated tests skipped)
- [x] Cloud end-to-end suite (88 passed, 15 optional variants skipped)
- [x] Data Center end-to-end suite (98 passed, 104 non-DC tests skipped)
- [x] Ruff formatting, Ruff lint, and strict mypy checks
- [x] Tool documentation completeness check (97 tools mapped)

### DIFF
--- a/docs/tools/jira-search-fields.mdx
+++ b/docs/tools/jira-search-fields.mdx
@@ -46,6 +46,32 @@ Cloud supports cursor-based pagination via `page_token` for deterministic result
 
 ---
 
+### Compare Issue Sets
+
+Compare two JQL result sets and report issues that were added, removed, changed, or unchanged.
+
+**Parameters:**
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `jql_a` | `string` | Yes | JQL query defining the first issue set |
+| `jql_b` | `string` | Yes | JQL query defining the second issue set |
+| `compare_fields` | `string` | No | Comma-separated field names to compare; defaults to `status,assignee,priority,labels` |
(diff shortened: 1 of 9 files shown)


trevorh:
@sooperset lmk if you want something like this included or not given the tool surface simplifcation.

This came up because I needed to validate the use of labels to identify a set of issues vs. finding descendents/siblings from a target issue.

If it's too specific, totally fine!

sooperset:
Appreciate you asking before building more on it, @trevorh. That's the right instinct.

Let's leave `compare_issue_sets` out. It's a useful operation, but a caller can get the same result from 2 `jira_search` calls and a diff on the keys, so it doesn't earn a permanent slot in the tool surface we're keeping small for v1.0 (#1104). The label-validation use case you hit is real though: if it turns out to need server-side set logic a client can't reasonably do (very large result sets, pagination across both sides), reopen with that constraint spelled out and I'll reconsider.

Thanks for the batch of Jira analysis work. The epic summary is landing as an `include` option (see #1285), and the graph/hierarchy work is heading for the upcoming `jira_discover`. This one I'd rather keep as a recipe than a tool.

PR #1285, opened 2026-04-18
feat(jira): add epic_summary as an include option on get_issue (fka new get_epic_summary tool)

trevorh:
## Summary

- Add jira_get_epic_summary for epic metadata and child aggregation by status, assignee, and issue type.
- Calculate completion from Jira status categories and optionally include individual children.
- Reuse cross-edition child discovery with bounded Cloud and Server/Data Center pagination.
- Preserve authentication and total-search failures instead of reporting an incorrect empty epic.
- Add an opt-in jira_epic_analysis toolset and update generated tool documentation.

## Verification

- [x] Focused Jira, server, and toolset unit tests: 76 passed
- [x] Full unit suite: 3,252 passed, 5 skipped
- [x] Cloud end-to-end suite: 87 passed, 15 skipped
- [x] Feature-specific Cloud epic summary test: passed
- [x] Data Center end-to-end suite: 98 passed, 1 skipped
- [x] Documentation mapping check: all 97 tools mapped
- [x] Full pre-commit gate: passed

## Data Center note

The configured Data Center test project does not expose the Epic issue type, so its feature-specific epic creation case skips. The full Data Center suite passes, and Server/Data Center pagination, authentication propagation, and failure handling have dedicated regression coverage.

### DIFF
--- a/src/mcp_atlassian/jira/epic_analysis.py
+++ b/src/mcp_atlassian/jira/epic_analysis.py
@@ -1,0 +1,176 @@
+"""Module for Jira epic analysis operations."""
+
+import logging
+from collections import Counter
+from typing import Any
+
+from requests.exceptions import HTTPError
+
+from ..utils.decorators import handle_auth_errors
+from .client import JiraClient
+
+logger = logging.getLogger("mcp-jira")
+
+# JQL queries tried in order to find children of an epic.
+_CHILD_JQL_TEMPLATES = [
+    'parent = "{key}"',
+    "'Epic Link' = \"{key}\"",
+]
+
+
+class EpicAnalysisMixin(JiraClient):
+    """Mixin for Jira epic summary and analysis."""
+
+    @handle_auth_errors("Jira API")
+    def get_epic_summary(
+        self,
+        epic_key: str,
+        include_children: bool = True,
+        max_children: int = 200,
+    ) -> dict[str, Any]:
+        """Summarise an epic: metadata, child counts, and aggregations.
+
+        Fetches the epic issue and its children, then groups the children
+        by status, assignee, and issue type with a completion percentage.
+
+        Args:
+            epic_key: The key of the epic issue (e.g. ``PROJ-123``).
+            include_children: When *True* the individual child issue list
+                is included in the response.
+            max_children: Upper bound on the number of children fetched.
+
+        Returns:
+            Dict with ``epic`` metadata, ``summary`` aggregations, and
+            optionally ``children`` list.
+
+        Raises:
+            ValueError: If the issue is not an Epic.
+            MCPAtlassianAuthenticationError: If authentication fails.
+            Exception: On API or query errors.
+        """
+        epic_issue = self.get_issue(epic_key)  # type: ignore[attr-defined]
+
+        type_name = ""
+        if epic_issue.issue_type:
+            type_name = epic_issue.issue_type.name
+        if type_name.lower() != "epic":
+            raise ValueError(
+                f"{epic_key} is a {type_name or 'unknown type'}, not an Epic"
+            )
+
+        epic_info: dict[str, Any] = {
+            "key": epic_issue.key,
+            "summary": epic_issue.summary,
+            "status": (epic_issue.status.name if epic_issue.status else None),
+            "assignee": (
+                epic_issue.assignee.display_name
+                if epic_issue.assignee
+                else "Unassigned"
+            ),
+        }
+
+        children = self._fetch_epic_children(epic_key, max_children)
+
+        by_status: Counter[str] = Counter()
+        by_assignee: Counter[str] = Counter()
+        by_type: Counter[str] = Counter()
+        done_count = 0
+
+        children_output: list[dict[str, Any]] = []
+
+        for child in children:
+            status_name = "Unknown"
+            if child.status:
+                status_name = child.status.name
+                if child.status.category and child.status.category.key == "done":
+                    done_count += 1
+            by_status[status_name] += 1
+
+            assignee_name = "Unassigned"
+            if child.assignee:
+                assignee_name = child.assignee.display_name
+            by_assignee[assignee_name] += 1
+
+            type_label = "Unknown"
+            if child.issue_type:
+                type_label = child.issue_type.name
+            by_type[type_label] += 1
+
+            if include_children:
+                children_output.append(
+                    {
+                        "key": child.key,
+                        "summary": child.summary,
+                        "status": status_name,
+                        "assignee": assignee_name,
+                        "issue_type": type_label,
+                    }
+                )
+
+        total = len(children)
+        completion = round(done_count / total * 100, 1) if total else 0.0
+
+        return {
+            "epic": epic_info,
+            "summary": {
+                "total_children": total,
+                "by_status": dict(by_status),
+                "by_assignee": dict(by_assignee),
+                "by_type": dict(by_type),
+                "completion_percentage": completion,
+            },
+            "children": children_output,
+        }
+
+    # ------------------------------------------------------------------
+    # Helpers
+    # ------------------------------------------------------------------
+
+    def _fetch_epic_children(
+        self,
+        epic_key: str,
+        max_children: int,
+    ) -> list:
+        """Try multiple JQL patterns to find children of an epic.
+
+        Args:
+            epic_key: The epic's issue key.
+            max_children: Maximum children to return.
+
+        Returns:
+            List of JiraIssue objects.
+        """
+        from ..models.jira.issue import JiraIssue
+
+        for template in _CHILD_JQL_TEMPLATES:
+            jql = template.format(key=epic_key)
+            try:
+                result = self.search_issues(  # type: ignore[attr-defined]
+                    jql=jql,
+                    fields=["summary", "status", "issuetype", "assignee"],
+                    limit=max_children,
+                )
+                issues: list[JiraIssue] = list(result.issues)
+                if not issues:
+                    continue
+
+                # Server/DC caps at 50 — page if needed.
+                while len(issues) < max_children and len(result.issues) >= 50:
+                    result = self.search_issues(  # type: ignore[attr-defined]
+                        jql=jql,
+                        fields=["summary", "status", "issuetype", "assignee"],
+                        start=len(issues),
+                        limit=max_children - len(issues),
+                    )
+                    if not result.issues:
+                        break
+                    issues.extend(result.issues)
+
+                return issues[:max_children]
+            except HTTPError:
+                raise
+            except Exception:
+                logger.debug("JQL pattern failed for epic %s: %s", epic_key, jql)
+                continue
+
+        return []
--- a/docs/tools/jira-issues.mdx
+++ b/docs/tools/jira-issues.mdx
@@ -17,7 +17,9 @@ Get details of a specific Jira issue.
 | `comment_limit` | `integer` | No | Maximum number of comments to include (0 or null for no comments) |
 | `properties` | `string` | No | (Optional) A comma-separated list of issue properties to return |
 | `update_history` | `boolean` | No | Whether to update the issue view history for the requesting user |
-| `include` | `string` | No | (Optional) Comma-separated sections to inline in the response, avoiding extra tool calls. Supported: all, remote_links, transitions, watchers, changelog, comments, worklogs |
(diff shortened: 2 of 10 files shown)


src/mcp_atlassian/jira/epic_analysis.py
```diff
@@ -0,0 +1,191 @@
...
+
+from ..models.jira import JiraIssue
+from ..utils.decorators import handle_auth_errors
+from ..utils.pagination import clamp_limit
+from .client import JiraClient
+from .protocols import EpicIssuesOperationsProto, IssueOperationsProto
+
+logger = logging.getLogger("mcp-jira")
+
+_LOCALIZED_EPIC_NAMES: set[str] = {"에픽", "エピック"}
```
sooperset: **BLOCKER:** This list is narrower than the localization support already used by `IssuesMixin._is_epic_issue_type`. `get_epic_summary()` returns `ValueError` for valid Epic names such as `史诗`, `史詩`, `épica`, `épique`, `epik`, `эпик`, and `епік`; I reproduced each mismatch directly. Please use one shared Epic-type predicate for both paths and add regression cases for every supported localized name. `get_epic_issues()` needs the same shared check so the later delegation doesn't reject them again.

tests/e2e/test_jira_dc_operations.py
```diff
@@ -103,6 +103,15 @@ def test_create_epic(
         resource_tracker.add_jira_issue(epic.key)
         assert epic.key.startswith(dc_instance.project_key)
 
+        summary = jira_fetcher.get_epic_summary(
```
sooperset: **BLOCKER:** The recorded DC run skips this test before it reaches `get_epic_summary()` because the configured project can't create Epics. Since this PR also changes the shared `EpicsMixin.get_epic_issues()` path, a green DC suite with this feature test skipped doesn't satisfy the required both-edition evidence. Please exercise this call against an existing DC Epic (or a test project with the Epic type enabled) and record a non-skipped pass on the final commit.

sooperset (changes requested):
Thanks for the contribution, @trevorh. The bounded pagination and explicit auth and outage regression coverage are solid.

1. Epic-type validation drops existing localized names and rejects valid Epics.
2. The Data Center feature test never reaches the new call, so the required cross-edition verification is incomplete.

src/mcp_atlassian/jira/epic_analysis.py
sooperset: **BLOCKER:** This list is narrower than the localization support already used by `IssuesMixin._is_epic_issue_type`. `get_epic_summary()` returns `ValueError` for valid Epic names such as `史诗`, `史詩`, `épica`, `épique`, `epik`, `эпик`, and `епік`; I reproduced each mismatch directly. Please use one shared Epic-type predicate for both paths and add regression cases for every supported localized name. `get_epic_issues()` needs the same shared check so the later delegation doesn't reject them again.

tests/e2e/test_jira_dc_operations.py
sooperset: **BLOCKER:** The recorded DC run skips this test before it reaches `get_epic_summary()` because the configured project can't create Epics. Since this PR also changes the shared `EpicsMixin.get_epic_issues()` path, a green DC suite with this feature test skipped doesn't satisfy the required both-edition evidence. Please exercise this call against an existing DC Epic (or a test project with the Epic type enabled) and record a non-skipped pass on the final commit.

sooperset (changes requested):
Thanks for the contribution, @trevorh. The bounded pagination and explicit auth and outage regression coverage are solid.

1. Epic-type validation drops existing localized names and rejects valid Epics.
2. The Data Center feature test never reaches the new call, so the required cross-edition verification is incomplete.

sooperset:
Thanks for pushing the follow-up, @trevorh. The shared Epic-type predicate and the non-skipped DC run resolve both points from the last review, and the localized-name coverage is exactly right.

The one thing left isn't your code. It's a fit question with where the tool surface is going: we're consolidating from ~95 tools toward ~30-35 outcome-oriented ones for v1.0 (#1104), so new capability is meant to land as a parameter or orchestration on an existing tool rather than a new standalone one. The `get_epic_summary` tool is net-new, so it runs into that direction.

That's on me, not you. I should have raised it in the first review instead of after you'd already done 2 rounds of fixes. Sorry for the wasted cycles.

Here's how I'd like to land it without more work on your side: fold the aggregation (children grouped by status and assignee, plus completion %) into the existing epic-reading path as an opt-in `include=summary` option, reusing the pagination and outage handling you built. I'm happy to take that adaptation on and keep you credited on the merge (Co-authored-by plus a link back to this PR). If you'd rather do it yourself, say the word and I'll send the exact parameter shape.

Either way your work lands. Just let me know which you prefer.

trevorh:
> Here's how I'd like to land it without more work on your side: fold the aggregation (children grouped by status and assignee, plus completion %) into the existing epic-reading path as an opt-in `include=summary` option, reusing the pagination and outage handling you built. I'm happy to take that adaptation on and keep you credited on the merge (Co-authored-by plus a link back to this PR). If you'd rather do it yourself, say the word and I'll send the exact parameter shape.

Hey @sooperset, no problem at all! Your project has been a great help to me (the DC support was key) so happy to take on the work.

Send me over what you're thinking, and I'll make the changes to align with where you want to go. Lmk if you want me to combine this one and #1287 together in a single PR or keep them separate.

sooperset:
Great, let's do it. One course-correct from my side first: keep them as separate PRs, and they're actually heading to 2 different homes, not the same one.

**#1285, epic summary.** This fits as an `include=epic_summary` section on `jira_get_issue`: when the issue is an Epic (your shared `_is_epic_issue_type` predicate), fetch children via `get_epic_issues` and inline the aggregation. A few guardrails I want baked in, since this turns a single-issue read into something that fans out:

1. Bound the output, not just the pagination: cap child count and response size, and set `truncated: true` when you hit the cap.
2. Never report an authoritative completion % on partial data. If pagination or an outage stops you fetching all children, return `partial: true` and distinguish "no children" from "couldn't fetch children".
3. Compute completion from the stable status category (the `done` category), not localized status names.
4. Aggregate assignees by account id where you have it, with an explicit `unassigned` bucket.
5. On a non-Epic, don't silently drop the section: return `applicable: false, reason: "not_epic"` and make zero child calls.
6. Keep `epic_summary` to the aggregate. If you also expose a child list, keep it to compact rows (key, status, assignee), not full issue objects, or the payload gets away from us.

Drop the standalone `get_epic_summary` tool and its toolset. That's a clean, self-contained PR.

**#1287, link graph + hierarchy.** I changed my mind here, and I owe you the honest version rather than the convenient one. These aren't issue-detail enrichments, they're multi-issue graph/discovery operations: cost grows with branching factor, not just depth, and they'd make an ordinary `get_issue` call unexpectedly large. Folding them into `get_issue` would satisfy the letter of the consolidation but not the point of it.

Their real home is the `jira_discover` meta-tool in Phase B (v0.23), which is exactly the discovery/traversal surface. So I don't want you reworking #1287 onto `get_issue`. Two options: hold #1287 as-is until I open the `jira_discover` design, or, if you're up for it, help shape that tool's traversal contract (depth, node/edge budgets, truncation, direction, cycle handling) and land the graph there. Your level-batched BFS, cross-link dedup, and native-parent work carries over either way.

No rush on #1287, #1285 is the one to land now. Happy to review a draft whenever.

trevorh:
Hey @sooperset, apologies for the delay on this, I've been traveling. Converted the standalone tool to an include parameter on `jira_get_issue` — no new tool, no new toolset.

On an Epic, the response gains child-issue counts by status, assignee, and type, plus a completion %. Aggregate-only by default; `epic_summary_children=true` adds a compact child list.

**Guardrails** (since this turns a single-issue read into a fan-out):
- Bounded: `max_epic_children` (default 100, hard cap 500), with `truncated: true` when there are more.
- Honest on partial data: truncated or unfetchable children report `partial: true` with no completion %, and "no children" is distinguished from "couldn't fetch".
- Completion is from the stable *done* status category, not localized status names.
- Non-Epic → `applicable: false` with zero child calls.
- Works on Cloud and Server/DC.

trevorh:
Hey @sooperset, notes below:

**1. Child-search failure now distinguishable from an empty epic.**
`get_epic_issues` tracked every failed JQL strategy but returned `[]` when all of them failed, so a child-search outage was indistinguishable from a childless epic. It now tracks whether any strategy's search *completed* (`had_clean_completion`). If every attempted strategy errored, it raises (carrying the last error as context) instead of returning `[]`; if at least one strategy ran cleanly and simply matched nothing, it still returns `[]` (a genuinely empty epic). `get_epic_summary` maps that fetch failure to `partial: true, reason: "fetch_failed"` with `total_children: null`, so it never reports an authoritative `total_children: 0` over a failed fetch. The regression drives this through failing `search_issues` calls (`test_partial_when_underlying_searches_fail`, plus `test_get_epic_issues_all_strategies_error_raises` at the method level), not by mocking `get_epic_issues` to raise.

**2. Authentication failures propagate.**
`MCPAtlassianAuthenticationError` subclasses `Exception`, so it was being swallowed both by each strategy's broad `except` inside `get_epic_issues` and by `get_epic_summary`'s own catch. Both now re-raise it explicitly before any broad handler (every strategy, both nested loops, the top-level handler, and the summary), so it reaches `ErrorPreservingFastMCP` untouched rather than becoming a partial payload. Regressions: `test_get_epic_issues_auth_error_propagates` and `test_auth_error_propagates_through_summary`.

**3. Pagination-ceiling truncation is now detected via a real `has_more` signal.**
You're right that `search_issues()` re-applies `clamp_limit()`, so the old `max_children + 1` sentinel was clamped back to the ceiling whenever `ATLASSIAN_MAX_PAGINATION_LIMIT <= max_children`, and the `len(fetched) > max_children` check could never fire. Rather than patch the sentinel, I took your option (b): the fetch now derives a real "more children exist" signal from the winning search's own metadata. `get_epic_issues` is split so the strategy chain returns the winning `JiraSearchResult` (`_get_epic_issues_result`); `get_epic_issues` stays a thin `.issues` wrapper (same list contract). `_fetch_epic_children` returns `(children, has_more)`, where `has_more` is the Cloud `next_page_token` (present iff more results exist beyond the page) or the Server/DC authoritative `total`. The `+ 1` sentinel is gone.

The summary then classifies the reason by which limit is binding: `reason: "pagination_ceiling"` when the operator ceiling capped the fetch (the caller cannot override it), and `reason: "truncated"` when the caller's own `max_children` is the cap (raise it to see more). Because the signal is now the server's real page boundary, an epic with *exactly* the ceiling number of children is correctly reported as complete rather than a false partial. Regressions cover Cloud and Server/DC with a low ceiling (`test_pagination_ceiling_limits_cloud`, `test_pagination_ceiling_limits_server_dc`), the exactly-at-ceiling complete case on both editions (`test_exactly_ceiling_children_is_authoritative` and its `_server_dc` sibling), and a small epic under the ceiling staying authoritative (`test_below_pagination_ceiling_is_authoritative`).

One subtlety the refactor also handles: on Server/DC the child list is paged, and paging now re-queries the *winning* strategy's JQL directly rather than re-running the whole discovery chain. Re-running the chain would let an empty later page fall through (via the same [#1509](https://github.com/sooperset/mcp-atlassian/pull/1509) empty-result behavior) to a different strategy and mix result sets, so `_get_epic_issues_result` now carries the winning JQL out for the pager to reuse.

**On [#1509](https://github.com/sooperset/mcp-atlassian/pull/1509)** (now merged). I've rebased this branch onto current `main`, so that `if search_result.issues:` truthy fix is already in place, and the BLOCKER 1 work here builds directly on it: the `had_clean_completion` failure signal wraps the same discovery gate #1509 corrected. It's the same class of "false authoritative zero" (an empty-but-truthy first result would have made the summary report `total_children: 0, partial: false`), so the two changes are complementary rather than duplicated.

**A design note for the record.** For BLOCKER 1 the failure signal originates in the strategy chain (that's where each strategy is caught): if every attempted strategy errors it raises, and `get_epic_summary` maps that to `fetch_failed` exactly as before. I kept `get_epic_issues`'s public `list[JiraIssue]` contract intact by moving the chain into a private `_get_epic_issues_result` that preserves the winning `JiraSearchResult`; the public method is a thin wrapper. That private result-returning helper is also what makes BLOCKER 3's real `has_more` possible, so both fixes share one seam rather than two. If you'd prefer a different split (for example keeping everything inside `get_epic_issues`), say the word.

All unit tests pass, ruff/mypy are clean, and the tool-docs check is green.

PR #1286, opened 2026-04-18
feat(jira): add project epic hierarchy and cross-project dependency tools

trevorh:
## Summary

- Add `jira_get_project_epic_hierarchy` and `jira_get_cross_project_dependencies` tools
- Epic hierarchy groups a project's epics under their cross-project parent issues, with direction-aware link classification
- Cross-project dependencies scans all issues and groups links by target project and link type
- Hierarchy phrase sets (parent-of / child-of) shared from `constants.py`, with custom link types configurable via `HIERARCHY_LINK_PHRASES`
- Cloud search paginates internally, while Server/Data Center uses offset pagination
- Propagate parent-summary lookup failures instead of returning incomplete hierarchy data
- Add `HIERARCHY_LINK_PHRASES` to `.env.example`

## Test plan

- [x] 20 project-analysis unit tests, including direction-aware parent detection, hierarchy inversion, custom phrase parsing, lookup error propagation, and Cloud no-repaging
- [x] Full unit and integration suite: 3260 passed, 28 skipped
- [x] Jira Cloud E2E suite: 12 passed, including live project analysis
- [x] Jira Data Center E2E suite: 11 passed, 1 environment-dependent skip, including live project analysis
- [x] Ruff, formatting, and strict mypy checks

> **Note:** Tools reference docs update will be submitted as a separate PR after feature PRs are merged.

PR #1287, opened 2026-04-18
feat(jira): add link graph traversal and issue tree tools

trevorh:
## Summary

- Add `jira_trace_issue_links` for breadth-first issue-link graph traversal with type, direction, depth, and issue-count controls
- Add `jira_get_issue_tree` for direction-aware containment trees with lateral links reported separately
- Reuse the shared hierarchy phrase configuration added by #1286, including `HIERARCHY_LINK_PHRASES`
- Register both read-only tools in the opt-in `jira_link_analysis` toolset

## Maintainer updates

- Rebased onto current `main` and integrated the overlapping project-analysis hierarchy constants and toolset registry
- Propagate all Jira lookup failures instead of returning partial or empty graphs silently
- Exclude mirrored child-to-parent hierarchy links from lateral `cross_links`
- Strengthen depth-batch, error-propagation, and mirrored-link regression coverage

## Verification

- [x] Focused unit tests: 52 passed
- [x] Full unit suite: 3292 passed, 5 skipped
- [x] Integration suite: 23 credential-gated tests skipped cleanly
- [x] Pre-commit hooks: passed (Ruff, formatting, mypy)
- [x] Jira Cloud end-to-end suite: 89 passed, 15 skipped
- [x] Jira Server/Data Center end-to-end suite: 99 passed, 105 skipped

Tools reference documentation can follow separately after the feature merges.

### DIFF
--- a/src/mcp_atlassian/jira/link_analysis.py
+++ b/src/mcp_atlassian/jira/link_analysis.py
@@ -128,7 +128,51 @@ class LinkAnalysisMixin(JiraClient):
             )
             for issue in search.issues:
                 result[issue.key] = issue.to_simplified_dict()
+
+        missing_keys = sorted(set(keys) - set(result))
+        if missing_keys:
+            missing = ", ".join(missing_keys)
+            message = f"Jira did not return requested issue(s): {missing}"
+            raise ValueError(message)
+
         return result
+
+    def _fetch_native_child_keys(
+        self,
+        parent_key: str,
+        max_issues: int,
+    ) -> list[str]:
+        """Fetch issue keys whose native Jira parent is *parent_key*.
+
+        The ``parent`` JQL field is available in both Jira Cloud and
+        Server/Data Center.  Cloud search handles its own continuation
+        tokens; Server/Data Center requires explicit paging.
+        """
+        if max_issues <= 0:
+            return []
+
+        jql = f'parent = "{parent_key}"'
+        child_keys: list[str] = []
+        start = 0
+
+        while len(child_keys) < max_issues:
+            result: JiraSearchResult = self.search_issues(  # type: ignore[attr-defined]
+                jql=jql,
+                fields=_LINK_FIELDS,
+                start=start,
+                limit=max_issues - len(child_keys),
+            )
+
+            page_keys = [issue.key for issue in result.issues if issue.key]
+            for child_key in page_keys:
+                if child_key not in child_keys:
+                    child_keys.append(child_key)
+
+            if self.config.is_cloud or len(result.issues) < _SERVER_DC_PAGE_SIZE:
+                break
+            start += len(result.issues)
+
+        return child_keys[:max_issues]
 
     @staticmethod
     def _extract_links(
@@ -204,6 +248,7 @@ class LinkAnalysisMixin(JiraClient):
 
         Raises:
             ValueError: If direction_filter is invalid.
+            ValueError: If Jira does not return a requested issue.
             MCPAtlassianAuthenticationError: If authentication fails.
             Exception: On API errors.
         """
@@ -326,6 +371,7 @@ class LinkAnalysisMixin(JiraClient):
             list, and ``total_nodes``.
 
         Raises:
+            ValueError: If Jira does not return a requested issue.
             MCPAtlassianAuthenticationError: If authentication fails.
             Exception: On API errors.
         """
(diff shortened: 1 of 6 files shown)


src/mcp_atlassian/jira/link_analysis.py
```diff
@@ -0,0 +1,393 @@
...
+                    remainder.append((nk, nd))
+            for item in remainder:
+                queue.append(item)
+
+            fetched = self._batch_fetch_issues(batch_keys)
+
+            for bk in batch_keys:
+                if bk in visited or len(visited) >= max_issues:
+                    continue
+                issue_dict = fetched.get(bk)
```
sooperset, line 295: **BLOCKER:** `_batch_fetch_issues` can return fewer issues than requested when the root doesn't exist, a linked issue isn't visible, or the project allowlist filters it out. Continuing here makes `trace_issue_links` return a successful empty or partial graph, while `get_issue_tree` later fabricates an `Unknown` root. Required for merge: fail an unresolved root lookup and expose or fail unresolved linked keys so callers can't mistake incomplete data for a complete graph. Please cover both cases with regression tests.

src/mcp_atlassian/jira/link_analysis.py
```diff
@@ -0,0 +1,393 @@
...
+                if st_key and st_key not in visited:
+                    child_keys.add(st_key)
+
+            for link in self._extract_links(issue_dict):
+                target = link["target_key"]
+                if link["is_child"]:
+                    if target not in visited:
+                        child_keys.add(target)
+                elif not link["is_hierarchy"]:
+                    cross_links.append(
```
sooperset: **IMPORTANT:** Jira exposes the same issue link from both endpoint payloads. When both endpoints are hierarchy nodes, this appends the same lateral relationship twice with reversed source and target. Please deduplicate `cross_links` by a canonical issue pair and link type, matching the approach in `trace_issue_links`, and add a mirrored non-hierarchy link test.

src/mcp_atlassian/jira/link_analysis.py
```diff
@@ -0,0 +1,393 @@
...
+            issue_dict = fetched.get(key)
+            if issue_dict is None:
+                return None
+
+            visited.add(key)
+            node = _node_from_issue(issue_dict, depth)
+            children: list[dict[str, Any]] = []
+            child_keys: set[str] = set()
+
+            for st in issue_dict.get("subtasks", []):
```
sooperset, line 395: **IMPORTANT:** The public tool description promises initiative-to-epic-to-story navigation and `_LINK_FIELDS` requests `parent`, but the implementation never reads that field. Jira's native parent relationship lives on the child, and `subtasks` only provides sub-task children, so an initiative root won't discover native epic or story children. Required for merge: either query direct children through the native parent relationship in a Cloud and Server/DC-safe way, or narrow the tool name and documentation to an explicit issue-link tree and remove the unused field. Please add a contract test for the chosen behavior.

sooperset (changes requested):
Thanks for the contribution @trevorh! The level-batched BFS and direction-aware hierarchy tests give this feature a solid foundation.

1. Don't return successful graphs or fabricated tree roots when Jira doesn't return a requested issue.
2. Deduplicate mirrored lateral links in `cross_links`.
3. Either support Jira's native `parent` hierarchy or narrow the tool contract to explicit issue links.

sooperset:
Thanks for pushing the fixes, @trevorh. The missing-issue handling, the cross-link dedup, and the native `parent` coverage close out the last round cleanly, and the level-batched BFS is holding up well.

Same situation as your epic PR: the one open item isn't your code, it's tool-surface fit. We're consolidating from ~95 tools toward ~30-35 outcome-oriented ones for v1.0 (#1104), so new capability should land as a parameter or orchestration on an existing tool rather than new standalone ones. This PR adds 2 net-new tools (the link-graph traversal and the issue tree), so it runs into that direction.

That's on me. I should have raised it in the first review, not after you'd already turned the code around. Sorry for the extra loop.

Here's how I'd like to land it without more work on your side: the traversal and tree both fit as an orchestration or `include` on the existing issue-reading path (or as a mode under the planned `jira_discover` consolidation), reusing the depth limit, filtering, and dedup you already built. I'm happy to take that adaptation on and keep you credited (Co-authored-by plus a link back to this PR). If you'd rather shape it yourself, tell me and I'll send the exact parameter surface.

Either way the work lands. Just let me know which you prefer.

sooperset:
Replied on #1285 with the plan for both PRs. Short version for this one: after another look, the graph traversal and hierarchy tree are multi-issue discovery operations, so folding them into `jira_get_issue` isn't the right fit. Their home is the `jira_discover` meta-tool coming in Phase B (v0.23). I'd rather hold this PR for that than reshape it onto `get_issue` now. Your BFS, dedup, and native-parent handling carries straight over. Details and options on #1285.

trevorh:
Thanks @sooperset definitely like that idea of collapsing these to a single tool method.

Are you thinking of something like this? Varying params per discovery operation?

```
jira_discover(kind, params)
  kind = projects | agile_boards | fields | link_types | … | issue_link_graph | issue_hierarchy_tree

  e.g. { kind: "projects",         params: { include_archived, start_at, limit } }
       { kind: "issue_link_graph", params: { issue_key, direction, link_types, max_depth, node_budget, edge_budget } }
```

And keeping a uniform response?

```
{ kind, payload, meta: { edition, partial, truncated, reason } }
```

Definitely want to keep cost and effort in mind so BFS, edge dedup, and the other strategies from this PR could carry over unless you've got something else in mind? (Was thinking of bounding the graph/tree kinds by node/edge budgets rather than depth alone, since cost grows with branching factor.)

If you'd like me to propose something more detailed, lmk. Just wanted to help jumpstart things. A couple of other things came to mind:

- Cloud JSM discovery: support now, or return an "unsupported edition" for the DC-only bits?
- Tree mode: descendant-only, or also walk ancestors?
- Truncation: a resumable token, or just a "partial + reason" flag?

PR #1288, opened 2026-04-20
feat(auth): add multi-user mode for URL-only server deployments

kaylovh:
## Motivation

Teams hosting one shared remote MCP service should not need to store a service
account or restart the server for every user. This PR adds a URL-only mode in
which the operator pins the Jira and Confluence instances while each HTTP client
supplies its own Atlassian credentials per request.

## Usage

```bash
JIRA_URL=https://your-company.atlassian.net \
CONFLUENCE_URL=https://your-company.atlassian.net/wiki \
uvx mcp-atlassian --transport streamable-http --multi-user --port 9000
```

Clients send one of:

- `Authorization: Basic <base64(email:api_token)>` for Cloud
- `Authorization: Bearer <oauth_token>` for Cloud OAuth; the server resolves
  and pins the Cloud ID from its configured URL
- `Authorization: Token <pat>` for Server/Data Center

## What changed

- Add `MCP_ATLASSIAN_MULTI_USER_MODE` and the `--multi-user` CLI flag.
- Allow Jira and Confluence configs to represent a pinned URL without
  server-side credentials.
- Resolve and validate a user-specific fetcher from each request's credentials.
- Build Cloud OAuth clients from a request token while resolving the tenant from
  the operator-configured URL.
- Reject mismatching request Cloud IDs, including when legacy or complete global
  OAuth configuration is also present.
- Treat a Cloud bearer token without `X-Atlassian-Cloud-Id` as OAuth in strict
  mode while retaining bearer-PAT compatibility for Server/Data Center.
- Process per-request credentials for both Streamable HTTP and SSE message
  endpoints.
- Reject requests that omit credentials instead of falling back to an
  auth-less global config.
- Require an HTTP transport and reject `stdio` in URL-only multi-user mode.
- Ignore per-request Jira and Confluence URL overrides in this mode so clients
  cannot redirect the server to another host.
- Give strict URL-only mode precedence when the legacy OAuth flag is also set.
- Document the deployment and client configuration in the README, auth docs,
  HTTP transport docs, Dockerfile, and `.env.example`.

## Compatibility

- Existing deployments with complete global credentials are unchanged; those
  credentials continue to take precedence.
- Existing `ATLASSIAN_OAUTH_ENABLE=true` deployments remain supported and keep
  their legacy per-request URL-header behavior.
- The stricter URL and tenant pinning applies only to the new URL-only mode.

## Verification

- `uv run pre-commit run --all-files` — passed (Ruff, formatting, strict mypy)
- `uv run pytest tests/unit/ -q` — 3,267 passed, 5 skipped
- `uv run pytest tests/integration/ -q` — 23 optional real-data tests skipped
- `uv run pytest tests/e2e --dc-e2e -q` — 98 passed, 103 Cloud-only skipped
- `uv run pytest tests/e2e/cloud --cloud-e2e -q`: 87 passed, 15 skipped

### DIFF
--- a/src/mcp_atlassian/servers/dependencies.py
+++ b/src/mcp_atlassian/servers/dependencies.py
@@ -439,16 +439,23 @@ def _create_user_config_for_fetcher(
             raise ValueError(
                 "OAuth access token missing in credentials for user auth_type 'oauth'"
             )
-        if (
-            not base_config
-            or not hasattr(base_config, "oauth_config")
-            or not getattr(base_config, "oauth_config", None)
-        ):
+        global_oauth_cfg = getattr(base_config, "oauth_config", None)
+        if global_oauth_cfg is None and cloud_id:
+            # URL-only multi-user configs intentionally have no server-side
+            # OAuth configuration. The request's access token and Cloud ID are
+            # sufficient to construct the Atlassian gateway client.
+            global_oauth_cfg = OAuthConfig(
+                client_id="",
+                client_secret="",
+                redirect_uri="",
+                scope="",
+                cloud_id=cloud_id,
+            )
+        if global_oauth_cfg is None:
             raise ValueError(
                 f"Global OAuth config for {type(base_config).__name__} is missing, "
-                "but user auth_type is 'oauth'."
+                "but user auth_type is 'oauth' and no Cloud ID was supplied."
             )
-        global_oauth_cfg = base_config.oauth_config
 
         # Determine if this is DC OAuth (base_url set, no cloud_id)
         is_dc_oauth = getattr(global_oauth_cfg, "is_data_center", False) is True
--- a/src/mcp_atlassian/servers/dependencies.py
+++ b/src/mcp_atlassian/servers/dependencies.py
@@ -18,7 +18,11 @@ from starlette.requests import Request
 from mcp_atlassian.confluence import ConfluenceConfig, ConfluenceFetcher
 from mcp_atlassian.jira import JiraConfig, JiraFetcher
 from mcp_atlassian.servers.context import MainAppContext
-from mcp_atlassian.utils.env import is_env_ssl_verify, is_env_truthy
+from mcp_atlassian.utils.env import (
+    is_env_ssl_verify,
+    is_env_truthy,
+    is_multi_user_mode,
+)
 from mcp_atlassian.utils.oauth import OAuthConfig
 from mcp_atlassian.utils.urls import validate_url_for_ssrf
 
@@ -699,12 +703,13 @@ async def _get_fetcher(ctx: Context, spec: _ServiceSpec) -> Any:
             f"Attempting global {spec.name}Fetcher for non-HTTP."
         )
 
-    # Fallback to global fetcher
+    # Fallback to global fetcher. Skip configs with no server-side auth_type:
+    # those exist only in multi-user mode and require per-request credentials.
     app_ctx = _get_app_lifespan_ctx(ctx)
     global_config_fallback = (
         getattr(app_ctx, spec.config_attr, None) if app_ctx else None
     )
-    if global_config_fallback:
+    if global_config_fallback and global_config_fallback.auth_type is not None:
         # An HTTP request that reaches here has no per-user identity: serving it
         # with the operator's global credentials lets any unauthenticated caller
         # transact as the operator. Refuse unless explicitly opted in. Non-HTTP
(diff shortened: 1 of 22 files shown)


src/mcp_atlassian/servers/dependencies.py
```diff
@@ -435,16 +439,23 @@ def _create_user_config_for_fetcher(
...
+        if global_oauth_cfg is None and cloud_id:
+            # URL-only multi-user configs intentionally have no server-side
+            # OAuth configuration. The request's access token and Cloud ID are
+            # sufficient to construct the Atlassian gateway client.
+            global_oauth_cfg = OAuthConfig(
+                client_id="",
+                client_secret="",
+                redirect_uri="",
+                scope="",
+                cloud_id=cloud_id,
```
sooperset: **BLOCKER:** In strict URL-only mode, this copies the request's Cloud ID into an OAuth config without checking it against `base_config.url`. Both `JiraClient` and `ConfluenceClient` then replace the pinned URL with `https://api.atlassian.com/ex/{service}/{cloud_id}`. A client can therefore select a different Atlassian tenant even though the docs promise configured URLs are always used. Required for merge: pin the Cloud ID server-side or validate that the requested Cloud ID resolves to the configured site, then add Jira and Confluence regression tests for mismatched tenants.

sooperset (changes requested):
Thanks for the contribution @kaylovh! The strict/legacy mode split and the added SSE message-path handling close real gaps.

1. Enforce the configured tenant for Cloud OAuth instead of trusting an arbitrary request Cloud ID.

kaylovh:
Thanks for the review @sooperset! Fixed the tenant blocker in 8ca0ff5.

In strict URL-only mode the server now pins the Cloud ID to the configured URL instead of trusting the request. A new resolve_cloud_id_for_url() helper resolves the configured site's Cloud ID server-side — from ATLASSIAN_OAUTH_CLOUD_ID, or the site's unauthenticated /_edge/tenant_info endpoint (cached). A request carrying a different X-Atlassian-Cloud-Id is now rejected, and clients no longer need to send that header at all: the tenant always follows the configured URL.

Added regression tests for mismatched Jira and Confluence tenants (test_oauth_url_only_rejects_mismatched_tenant) plus resolver unit tests, and updated the README, auth/HTTP-transport docs, and env.example. pre-commit (Ruff + strict mypy) and the unit suite (3,234 passed) are green.

One design note: the Cloud ID is resolved lazily at request time (then cached; skipped entirely when ATLASSIAN_OAUTH_CLOUD_ID is set). Happy to move resolution to startup in the lifespan instead if you'd prefer to keep the request path network-free.

src/mcp_atlassian/utils/oauth.py
```diff
@@ -48,6 +48,71 @@
...
+    Args:
+        url: The configured Atlassian site base URL.
+
+    Returns:
+        The Cloud ID string, or ``None`` if the URL is not an Atlassian Cloud
+        site or the Cloud ID could not be resolved.
+    """
+    env_cloud_id = os.getenv("ATLASSIAN_OAUTH_CLOUD_ID")
+    if env_cloud_id:
+        return env_cloud_id
```
sooperset, line 85: **BLOCKER:** `ATLASSIAN_OAUTH_CLOUD_ID` returns before `url` is examined. This means a stale value, or a deployment with Jira and Confluence on different Cloud sites, routes one service through the wrong tenant even though this mode promises that configured URLs stay authoritative. Required for merge: validate the ID against each configured URL, use service-specific pins, or reject configurations that can't prove a 1:1 match. Add a regression test with the 2 services on different Cloud origins.

src/mcp_atlassian/utils/oauth.py
```diff
@@ -48,6 +48,71 @@
...
+        return None
+
+    parsed = urllib.parse.urlparse(url)
+    origin = f"{parsed.scheme}://{parsed.netloc}"
+    if origin in _CLOUD_ID_CACHE:
+        return _CLOUD_ID_CACHE[origin]
+
+    tenant_info_url = f"{origin}{TENANT_INFO_PATH}"
+    try:
+        response = requests.get(tenant_info_url, timeout=HTTP_TIMEOUT)
```
sooperset, line 97: **BLOCKER:** This synchronous `requests.get` runs while `_get_fetcher()` is servicing an async request. A slow lookup can block the shared event loop for the 5-second connect plus 20-second read timeout, stalling unrelated Basic/PAT users. Only successful lookups are cached, so every failed Bearer request repeats the stall. Required for merge: resolve at startup or offload the call, cache failures for a bounded period, and add a concurrency regression test.

sooperset (changes requested):
The latest strict-mode precedence update closes the request-header bypass, and the Jira/Confluence coverage for that path is solid.

1. Validate every server-side Cloud ID against its configured service URL.
2. Keep tenant discovery off the async request path and bound retries after failures.

PR #1291, opened 2026-04-20
fix(jira): simplify get_all_projects output

balcsida:
## Summary

`jira_get_all_projects` currently returns the raw Jira REST payload untouched — so every call ships back `avatarUrls`, `self`, `expand`, `projectTypeKey`, `simplified`, `style`, `isPrivate`, `properties`, `uuid`, `entityId`, etc. for every project. On instances with hundreds of projects this dominates the tool response and eats agent context for no benefit.

The `JiraProject` model already has `from_api_response()` + `to_simplified_dict()` that emit only the useful keys (`key`, `name`, `description`, `category`, `lead`, `avatar_url`), but `ProjectsMixin.get_all_projects()` was never wired up to use them.

This PR routes the raw list through `JiraProject` so the tool returns the same simplified shape every other Jira model already uses.

## Changes

- `src/mcp_atlassian/jira/projects.py` — `get_all_projects()` parses each entry via `JiraProject.from_api_response(...).to_simplified_dict()` before returning.
- `tests/unit/jira/test_projects.py`:
  - New `test_get_all_projects_strips_bloat_fields` asserting `avatarUrls`, `self`, `expand`, `projectTypeKey`, `properties`, `uuid`, `entityId`, etc. are stripped and whitelisted keys are present (including nested `lead`).
  - Updated existing `test_get_all_projects` to assert on the simplified shape instead of raw equality.

## Compatibility

Internal callers continue to work with the simplified shape:

- `get_project_keys()` reads `project["key"]` → simplified dict exposes `key`.
- `get_project_leads()` reads `lead.get("name") or lead.get("displayName")` → `JiraUser.to_simplified_dict()` exposes `name` (= `username or display_name`), satisfying the primary lookup.
- `get_user_accessible_projects()` reads `project["key"]` → exposed.
- `servers/jira.py::get_all_projects` uppercases `project["key"]` and filters by `JIRA_PROJECTS_FILTER` using the same key — unchanged.

## Test plan

- [x] New unit test passes (RED → GREEN confirmed)
- [x] `uv run pytest tests/unit/jira/test_projects.py` — 48 passed
- [x] `uv run pytest tests/unit/` — 2579 passed, 5 skipped
- [x] `uv run pre-commit run --files src/mcp_atlassian/jira/projects.py tests/unit/jira/test_projects.py` — ruff-format, ruff, mypy all clean

### DIFF
--- a/src/mcp_atlassian/jira/projects.py
+++ b/src/mcp_atlassian/jira/projects.py
@@ -23,19 +23,25 @@ class ProjectsMixin(JiraClient, SearchOperationsProto):
         """
         Get all projects visible to the current user.
 
+        Returns simplified project dicts (key, name, description, category,
+        lead, avatar_url) with raw Jira fields like ``avatarUrls``, ``self``,
+        ``expand``, ``projectTypeKey`` stripped.
+
         Args:
             include_archived: Whether to include archived projects
 
         Returns:
-            List of project data dictionaries
+            List of simplified project data dictionaries
         """
         try:
-            params = {}
-            if include_archived:
-                params["includeArchived"] = "true"
-
             projects = self.jira.projects(included_archived=include_archived)
-            return projects if isinstance(projects, list) else []
+            if not isinstance(projects, list):
+                return []
+            return [
+                JiraProject.from_api_response(p).to_simplified_dict()
+                for p in projects
+                if isinstance(p, dict)
+            ]
 
         except Exception as e:
             logger.error(f"Error getting all projects: {str(e)}")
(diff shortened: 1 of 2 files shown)


balcsida:
still relevant

balcsida:
Still relevant

sooperset:
Nice cleanup here: routing the project list through `JiraProject` keeps the useful fields while cutting the noisy raw REST payload. I also appreciate the coverage around stripped fields, it made the merge review easy.

PR #1302, opened 2026-04-27
feat(jira): add JSM customer request read tools

any-victor:
## Description

Adds Jira Service Management customer-request read support for requests visible through the customer portal API but not through the standard Jira issue API.

This complements the request discovery and creation tools already on `main`.

## Changes

- Add `jira_get_customer_request` and `jira_search_customer_requests`.
- Add read tools for request comments, status history, and customer-visible transitions.
- Keep the REST methods in the shared `CustomerRequestsMixin` used by Cloud and Server/Data Center.
- Update the Service Desk toolset documentation and generated-doc category mapping.
- Add fetcher and MCP server regression coverage.
- Merge current `main` while preserving the contributor commits.

## Testing

- `uv run pytest tests/unit/ -q` — 3232 passed, 5 skipped
- Focused Jira customer-request, queue, server, and toolset tests — 199 passed
- `uv run pytest tests/e2e/cloud/test_jira_*.py tests/e2e/cloud/test_mcp_tools_cloud.py --cloud-e2e -q -rs` — 41 passed, 8 unavailable OAuth-variant skips
- `uv run pytest tests/e2e --dc-e2e -q -rs` — 98 passed, 102 Cloud exclusions; one project-specific Epic skip
- Full Cloud e2e — 85 passed, 15 skipped, with one pre-existing Confluence search failure reproduced unchanged on `origin/main`
- `uv run pre-commit run --files ...` — passed, including Ruff and mypy
- `uv run python scripts/generate_tool_docs.py --check` — all 101 tools mapped

The shared Cloud account does not have Jira Service Management request permission, so a direct live request-list probe returns 403. The normal Jira Cloud e2e tier is green, and the endpoint paths and parameters were cross-checked against the official Cloud and Server/Data Center REST references.

## Checklist

- [x] Unit and server regression tests pass.
- [x] Cloud and Server/Data Center e2e tiers were re-run.


src/mcp_atlassian/servers/jira.py
```diff
@@ -3392,6 +3392,265 @@ async def get_queue_issues(
...
+    annotations={"title": "Search Customer Requests", "readOnlyHint": True},
+)
+async def search_customer_requests(
+    ctx: Context,
+    request_ownership: Annotated[
+        str,
+        Field(
+            description=(
+                "Request ownership filter. Common values: OWNED_REQUESTS, "
+                "PARTICIPATED_REQUESTS, ALL_REQUESTS."
```
sooperset, line 3451: **BLOCKER:** This still exposes Cloud and DC through one scalar contract. Cloud defines `requestOwnership` as repeated values and supports `APPROVER`, `ORGANIZATION`, and `ALL_ORGANIZATIONS`, while DC accepts one value from a smaller set. `ALL_REQUESTS` is deprecated on Cloud, and `organizationId` is Cloud-only and valid only with `ORGANIZATION`. Required for merge: make the parameters and validation edition-aware and add regression coverage for both contracts.

src/mcp_atlassian/jira/customer_requests.py
```diff
@@ -377,6 +377,224 @@ def _should_retry_without_on_behalf(error_message: str) -> bool:
...
+        """
+        if not issue_id_or_key or not issue_id_or_key.strip():
+            raise ValueError("Issue ID or key is required")
+        if start_at < 0:
+            raise ValueError("start_at must be >= 0")
+        if limit < 1:
+            raise ValueError("limit must be >= 1")
+
+        response = self.jira.get(
+            f"rest/servicedeskapi/request/{issue_id_or_key.strip()}/transition",
```
sooperset, line 586: **BLOCKER:** Older supported DC releases mark this operation experimental and require the `X-ExperimentalApi` opt-in header. This call omits it, so the tool can fail on part of the documented Jira 8.14+ support range. Required for merge: send the header where needed and assert it in a regression test.

docs/tools-reference.mdx
```diff
@@ -78,7 +78,7 @@ Toolsets group related tools for easier management via `TOOLSETS` env var or `--
 | `jira_attachments` | No | `jira_download_attachments`, `jira_get_issue_images` |
 | `jira_users` | No | `jira_get_user_profile` |
 | `jira_watchers` | No | `jira_get_issue_watchers`, `jira_add_watcher`, `jira_remove_watcher` |
-| `jira_service_desk` | No | `jira_get_service_desk_for_project`, `jira_get_service_desk_queues`, `jira_get_queue_issues`, `jira_get_request_types`, `jira_get_request_type_fields`, `jira_create_customer_request` |
+| `jira_service_desk` | No | `jira_get_service_desk_for_project`, `jira_get_service_desk_queues`, `jira_get_queue_issues`, `jira_get_request_types`, `jira_get_request_type_fields`, `jira_create_customer_request`, `jira_get_customer_request`, `jira_search_customer_requests`, `jira_get_customer_request_comments`, `jira_get_customer_request_statuses`, `jira_get_customer_request_transitions` |
```
sooperset, line 81: **IMPORTANT:** This row adds 5 tools, raising the total from 96 to 101, but the page metadata, introduction, and `TOOLSETS=all` example still say 96. Required for merge: update those public totals.

sooperset (changes requested):
Thanks for the contribution @any-victor! Updating the ownership descriptions in the fetcher, tool schema, and docs keeps those 3 surfaces aligned.

1. The shared search and transition wrappers still don't match the supported Cloud and DC contracts.
2. None of the Cloud or DC e2e tests calls the 5 new endpoints, and the direct Cloud probe returned 403.
3. The generated catalog still reports 96 tools after adding 5.

src/mcp_atlassian/servers/jira.py
sooperset, line 3451: **BLOCKER:** This still exposes Cloud and DC through one scalar contract. Cloud defines `requestOwnership` as repeated values and supports `APPROVER`, `ORGANIZATION`, and `ALL_ORGANIZATIONS`, while DC accepts one value from a smaller set. `ALL_REQUESTS` is deprecated on Cloud, and `organizationId` is Cloud-only and valid only with `ORGANIZATION`. Required for merge: make the parameters and validation edition-aware and add regression coverage for both contracts.

src/mcp_atlassian/jira/customer_requests.py
sooperset, line 586: **BLOCKER:** Older supported DC releases mark this operation experimental and require the `X-ExperimentalApi` opt-in header. This call omits it, so the tool can fail on part of the documented Jira 8.14+ support range. Required for merge: send the header where needed and assert it in a regression test.

docs/tools-reference.mdx
sooperset, line 81: **IMPORTANT:** This row adds 5 tools, raising the total from 96 to 101, but the page metadata, introduction, and `TOOLSETS=all` example still say 96. Required for merge: update those public totals.

sooperset (changes requested):
Thanks for the contribution @any-victor! Updating the ownership descriptions in the fetcher, tool schema, and docs keeps those 3 surfaces aligned.

1. The shared search and transition wrappers still don't match the supported Cloud and DC contracts.
2. None of the Cloud or DC e2e tests calls the 5 new endpoints, and the direct Cloud probe returned 403.
3. The generated catalog still reports 96 tools after adding 5.

PR #1306, opened 2026-04-28
fix(confluence): fix attachment upload for Confluence Server/Data Center

maxsur:
Confluence Server and Data Center require a different approach for attachment uploads compared to Confluence Cloud:

1. Use POST instead of PUT The original implementation used PUT for the attachment endpoint. PUT works in Confluence Cloud (creates or updates) but is not supported for attachment uploads in Confluence Server/Data Center. The correct method for Server/DC is POST.

2. Use 'no-check' (with hyphen) for X-Atlassian-Token header Confluence Server requires 'X-Atlassian-Token: no-check' (with a hyphen) to bypass XSRF token validation. The value 'nocheck' (without hyphen) is not accepted.

3. Handle existing attachments via the /data sub-endpoint Confluence Server returns HTTP 400 when POSTing a new attachment whose filename already exists on the page, rather than silently creating a new version. To update an existing attachment, the request must be sent to: POST /rest/api/content/{pageId}/child/attachment/{attachmentId}/data The fix checks for an existing attachment with the same filename before uploading and selects the correct endpoint accordingly.

Tested on Confluence Server 6.x+ with Personal Access Token (PAT) authentication.

## Description

Fixes: #

## Changes

-
-
-

## Testing

## Checklist

- [ ] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/confluence/attachments.py
+++ b/src/mcp_atlassian/confluence/attachments.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+import sys
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


sooperset:
Your diagnosis was right: Confluence Server/DC needs `POST`, the `no-check` token value, and the per-attachment `/data` endpoint when updating an existing filename.

Those fixes have since landed in [#1253](https://github.com/sooperset/mcp-atlassian/pull/1253), along with regression tests and live Cloud and Data Center coverage. That implementation also preserves the upload comment and `minorEdit` metadata.

If you find a Server/DC case that #1253 still doesn't cover, please open a narrowed follow-up PR with a regression test for that behavior. I'm closing this one in favor of the merged change.

PR #1307, opened 2026-04-28
feat(jira): support [~accountid:UUID] mention syntax in markdown_to_adf

### LINKED ISSUE #1208: [Feature]: Support for ADF mention nodes in jira_add_comment (Markdown → ADF conversion)

merely04:
## Description

Adds support for Jira's wiki-style mention syntax `[~accountid:ACCOUNT_ID]` to the Markdown-to-ADF converter. Previously, that syntax was preserved verbatim as plain text, so `jira_add_comment` could not actually `@`-mention a user — comments rendered the literal `[~accountid:...]` text in Jira UI with no notification.

After this change the converter emits a proper ADF `mention` node, which Jira Cloud renders as a clickable user pill and which triggers a real `@`-notification.

The syntax mirrors what the **read** path already returns when a real ADF mention is read back through Jira's v2 API: `preprocessing/jira.py` rewrites incoming mentions into `[~accountid:ID]` text. So now read and write are symmetric — the same string round-trips cleanly, no display-name lookup needed on the LLM side.

Fixes: #1208

## Relation to #1228

PR #1228 (open, no review yet) solves the same issue with a **different** input syntax: `@[Display Name](accountid:xxx)`. That syntax requires the caller to know the user's display name; this PR's `[~accountid:UUID]` only needs the account id, which is what `get_user_profile` and other read paths already surface. The two PRs are complementary — they touch the same regex but add non-conflicting alternatives — and can be merged together. Happy to rebase on top of #1228 if you'd prefer to land that one first.

## Changes

- Add `[~accountid:(?P<mention_id>[^\]]+)]` alternative to the inline regex in `_parse_inline_formatting()`. Placed first so it cannot be mis-matched as a regular `[text](url)` link.
- Emit `{type: \"mention\", attrs: {id: ...}}` ADF node when matched. This shape matches what `adf_to_text` already understands on the read path (`adf.py:313-316`).
- Both modern (`712020:UUID-with-dashes`) and legacy (`24-hex-bare`) account-id formats are accepted — the regex doesn't constrain the account-id contents.

## Testing

- [x] Unit tests added: 5 new cases in `TestMarkdownToAdf` covering modern format, legacy format, surrounding text, no false-positive on regular links, and multiple mentions in one paragraph.
- [x] All tests pass: \`uv run pytest\` → **2583 passed, 5 skipped, 0 failures**.
- [x] Pre-commit clean: ruff-format, ruff, mypy all pass on changed files.
- [ ] Live-tested on Jira Cloud — pending; will verify after merge or on a draft branch.

## Checklist

- [x] Code follows project style guidelines (ruff/mypy).
- [x] Tests added for the new feature.
- [x] All tests pass locally.
- [x] No documentation update needed (existing docstring updated to mention the new syntax).

### DIFF
--- a/src/mcp_atlassian/models/jira/adf.py
+++ b/src/mcp_atlassian/models/jira/adf.py
@@ -14,21 +14,28 @@ def _parse_inline_formatting(text: str) -> list[dict[str, Any]]:
     """Parse inline Markdown formatting into ADF inline nodes.
 
     Handles: bold (**), italic (*), inline code (`), links ([text](url)),
-    and strikethrough (~~).
+    strikethrough (~~), and Jira-flavored user mentions
+    ([~accountid:ACCOUNT_ID]).
+
+    The mention syntax mirrors what Jira's v2 wiki text returns when a real
+    ADF mention is read back, so the read and write paths are symmetric.
 
     Args:
         text: Raw text potentially containing inline Markdown formatting.
 
     Returns:
-        List of ADF inline nodes (text nodes with optional marks).
+        List of ADF inline nodes (text nodes with optional marks, plus
+        mention nodes when [~accountid:...] is present).
     """
     if not text:
         return []
 
     nodes: list[dict[str, Any]] = []
-    # Pattern order matters: bold before italic, code before others
+    # Pattern order matters: mention before link (both start with `[`),
+    # bold before italic, code before others
     inline_re = re.compile(
-        r"`(?P<code_inner>[^`]+)`"
+        r"\[~accountid:(?P<mention_id>[^\]]+)\]"
+        r"|`(?P<code_inner>[^`]+)`"
         r"|\*\*(?P<bold_inner>.+?)\*\*"
         r"|~~(?P<strike_inner>.+?)~~"
         r"|\[(?P<link_text>[^\]]+)\]\((?P<link_href>[^)]+)\)"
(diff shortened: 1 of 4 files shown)


sooperset:
Merged. The `[~accountid:...]` syntax now maps cleanly to a Jira mention node, and the parser plus Cloud write coverage made this easy to verify.

PR #1309, opened 2026-04-28
feat(jira): add ADF expand block support

slinkardbrandon:
## Description

Adds native collapsible sections to Jira Cloud issue descriptions converted from Markdown.

## Changes

- Recognize `{expand:Title}...{expand}` and untitled `{expand}...{expand}` blocks in `markdown_to_adf()`.
- Recursively convert supported Markdown inside expand blocks, including headings, lists, and code blocks.
- Document the Cloud-only syntax in the create/update tool schemas and Jira issue guide.
- Add unit coverage and a live Jira Cloud write/readback regression test.

The raw ADF update guard from the original PR is no longer part of this diff because current `main` already supports explicit ADF descriptions and includes regression coverage.

## Verification

- `uv run pytest tests/unit/models/test_adf.py tests/unit/servers/test_jira_server.py -xvs` — 232 passed
- `uv run pytest tests/integration/ -xvs` — 23 collected and skipped by the existing integration opt-in guards
- `uv run pytest tests/e2e/cloud/test_adf_write.py --cloud-e2e -xvs` — 13 passed
- `uv run pre-commit run --all-files` — passed


sooperset:
The live Cloud regression test and matching tool guidance make the expand syntax easy to trust and use. Thanks for carrying both through, Brandon!

PR #1310, opened 2026-04-29
feat: opt-in HTTP hardening (retry, rate limit, concurrency cap, circuit breaker, pagination clamp)

strvmarv:
## Description

Adds opt-in HTTP hardening to protect Atlassian instances — especially self-hosted Server / Data Center — from runaway agent loops. All features gate behind environment variables and default to disabled / off, so behavior is unchanged unless an operator opts in.

This is contributed back from a private fork that ran for ~24h against a self-hosted Jira/Confluence DC instance.

Fixes: # 

## Changes

- Add urllib3 Retry on Sessions; respect Retry-After on 429/502/503/504
- Surface Retry-After value in 429 error messages
- Optional process-wide concurrency cap (`ATLASSIAN_MAX_CONCURRENT_REQUESTS`)
- Optional outbound RPS limiter (`ATLASSIAN_REQUESTS_PER_SECOND`)
- Optional pagination ceiling (`ATLASSIAN_MAX_PAGINATION_LIMIT`)
- Optional circuit breaker for repeated 429/503 (`ATLASSIAN_CIRCUIT_BREAKER_*`)
- Windows test fix: explicit utf-8 subprocess capture in `test_stdio_lifecycle`

## Testing

- [x] Unit tests added/updated (`tests/unit/utils/test_http.py`, `tests/unit/utils/test_pagination.py`)
- [x] Integration tests passed (full unit suite: 2600 passed locally on Windows)
- [x] Manual checks performed: ran the fork against a real self-hosted Jira/Confluence DC instance for ~24 hours, verified retry, concurrency cap, and pagination clamp behavior end-to-end through MCP tool calls

Two pre-existing platform-specific Windows failures unrelated to this PR (Python 3.13 timestamp boundary in `test_date`, `.zip` MIME detection in `test_media`) are unchanged.

## Checklist

- [x] Code follows project style guidelines (linting passes — `pre-commit run --all-files` green).
- [x] All tests pass locally.

## Notes for the reviewer

- **All knobs default to off.** I started with conservative defaults active by default in my fork, but realized changing observable behavior (request concurrency, response sizes) without explicit operator consent isn't appropriate for an upstream merge. Cloud users see no change.
- **Commits are atomic and reviewable.** Six feature commits + one `chore:` cleanup commit applying ruff-format / mypy / FBT003 fixes that I'm happy to fold into the relevant feature commits via interactive rebase if you'd prefer cleaner history.
- **Happy to split into smaller PRs.** If a single PR is too large, I can submit commits 1–2 (retry + Retry-After) as a first PR and follow up with the rest. Let me know what fits your review style best.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -141,6 +141,20 @@ CONFLUENCE_API_TOKEN=your_confluence_api_token_here
 #JIRA_TIMEOUT=75
 #CONFLUENCE_TIMEOUT=75
 
+# --- HTTP Hardening (Advanced) ---
+# Optional safeguards for overloaded Atlassian instances. All are disabled by
+# default unless set here.
+# Retry only read requests by default. Set ATLASSIAN_RETRY_INCLUDE_WRITES=true
+# only if your write endpoints are known to be idempotent.
+#ATLASSIAN_RETRY_TOTAL=5
+#ATLASSIAN_RETRY_BACKOFF=1.0
+#ATLASSIAN_RETRY_INCLUDE_WRITES=false
+#ATLASSIAN_MAX_CONCURRENT_REQUESTS=4
+#ATLASSIAN_REQUESTS_PER_SECOND=5
+#ATLASSIAN_MAX_PAGINATION_LIMIT=100
+#ATLASSIAN_CIRCUIT_BREAKER_THRESHOLD=5
+#ATLASSIAN_CIRCUIT_BREAKER_COOLDOWN=30
+
 # --- Read-Only Mode ---
 # Disables all write operations (create, update, delete). Default is false.
 #READ_ONLY_MODE=false
--- a/docs/configuration.mdx
+++ b/docs/configuration.mdx
@@ -160,6 +160,22 @@ This guide covers IDE integration, environment variables, and advanced configura
 
 See [.env.example](https://github.com/sooperset/mcp-atlassian/blob/main/.env.example) for all available options.
 
+### HTTP Hardening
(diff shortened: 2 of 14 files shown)


strvmarv:
Thanks for the review! All 5 points addressed in `af6fa17`:

1. **Circuit breaker half-open** — fixed. `before_send` no longer resets `failures` on cooldown elapse, so a single 429/503 in `on_response` re-opens immediately (matching the docstring); a 2xx still drops it to closed.
2. **`Retry-After` HTTP-date** — handled. Try `int()` parse first; on failure surface the raw header value without claiming "seconds". Added a regression test.
3. **`status_forcelist` ordering** — switched to set comparison.
4. **Env helper duplication** — promoted `get_int_env` and `get_float_env` into `utils/env.py` (alongside the existing truthy helpers) and replaced `_bool_env` with `is_env_extended_truthy`. The local helpers in `utils/http.py` are gone.
5. **`confluence.search_user`** — added `clamp_limit` to `search_user`; the Server/DC fallback inherits via the already-clamped limit.

Pre-commit (`ruff-format`, `ruff`, `mypy`) green; full unit suite passes.

strvmarv:
Not stale

sooperset:
The opt-in defaults and the half-open circuit-breaker regression test were the right shape here. I especially appreciate that the hardening stays off unless operators choose the knobs, which keeps existing Cloud and DC installs compatible.

PR #1316, opened 2026-05-05
Preserve MCP tool error details

### LINKED ISSUE #1315: Preserve MCP tool error details
## Problem
When a Jira or Confluence MCP tool fails, clients can receive a generic error like `Error calling tool 'search'` instead of the underlying exception details. Some Confluence search API failures are also converted into empty successful result lists, which hides real API/network/processing failures.

## Expected behavior
Tool failures should include actionable details from the underlying exception, and search helpers should propagate failures instead of returning success-shaped empty lists.

## Proposed fix
- register Atlassian FastMCP tools through an error-preserving wrapper
- convert unhandled tool exceptions into detailed `ToolError`s
- propagate detailed Confluence API/search failures
- add regression tests for Jira/Confluence search and validation/config errors

WZHBEDZ:
## Summary
- register Jira, Confluence, and main Atlassian MCP tools through an error-preserving FastMCP wrapper
- convert unexpected tool failures into detailed `ToolError`s instead of generic masked errors
- stop Confluence search helpers from returning empty success results for API/network/processing failures
- add regression coverage for Jira/Confluence search and validation/config error details
- update the branch to the current `main`

Closes #1315

## Tests
- `uv run pytest tests/unit/utils/test_decorators.py tests/unit/servers/test_confluence_server.py tests/unit/servers/test_jira_server.py tests/unit/confluence/test_search.py -q` (258 passed)
- `uv run pytest tests/unit/ -q` (3,086 passed, 5 skipped)
- `uv run pytest tests/integration/ --integration -q` (8 passed, 15 skipped)
- `uv run pytest tests/e2e/cloud --cloud-e2e -q` (82 passed, 15 skipped)
- `uv run pytest tests/e2e --dc-e2e -q` (94 passed, 98 skipped)

### DIFF
--- a/src/mcp_atlassian/confluence/search.py
+++ b/src/mcp_atlassian/confluence/search.py
@@ -67,6 +67,16 @@ class SearchMixin(ConfluenceClient):
         # Execute the CQL search query
         results = self.confluence.cql(cql=cql, limit=limit)
 
+        # A well-formed response always carries a "results" field (which may be
+        # an empty list). A response missing it entirely indicates an API,
+        # network, or processing failure that must surface as an error rather
+        # than be silently treated as an empty result set.
+        if not isinstance(results, dict) or "results" not in results:
+            raise ValueError(
+                "Confluence search returned a malformed response "
+                "missing the 'results' field"
+            )
+
         # Convert the response to a search result model
         search_result = ConfluenceSearchResult.from_api_response(
             results,
--- a/src/mcp_atlassian/confluence/search.py
+++ b/src/mcp_atlassian/confluence/search.py
@@ -48,13 +48,13 @@ class SearchMixin(ConfluenceClient):
 
     @staticmethod
     def _validate_search_response(response: Any, operation: str) -> dict[str, Any]:
-        """Ensure a search API response carries the required 'results' field.
+        """Ensure a search API response has a list of results.
 
         A well-formed Confluence search response always includes a "results"
-        field (which may legitimately be an empty list). A response missing it
-        entirely indicates an API, network, or processing failure that must
-        surface as an error rather than be silently treated as an empty result
-        set.
+        field containing a list (which may legitimately be empty). A response
+        missing it or containing a different type indicates an API, network, or
+        processing failure that must surface as an error rather than be
+        silently treated as an empty result set.
 
         Args:
             response: The raw response returned by the Confluence API.
@@ -65,13 +65,21 @@ class SearchMixin(ConfluenceClient):
             The validated response dictionary.
 
         Raises:
-            ValueError: If the response is not a dict or is missing "results".
+            ValueError: If the response is not a dict or "results" is not a
+                list.
         """
         if not isinstance(response, dict) or "results" not in response:
-            raise ValueError(
+            error = (
                 f"Confluence {operation} returned a malformed response "
                 "missing the 'results' field"
             )
+            raise ValueError(error)
+        if not isinstance(response["results"], list):
+            error = (
+                f"Confluence {operation} returned a malformed response "
+                "where 'results' is not a list"
+            )
+            raise ValueError(error)
         return response
 
     @handle_atlassian_api_errors("Confluence API")
--- a/src/mcp_atlassian/confluence/search.py
+++ b/src/mcp_atlassian/confluence/search.py
@@ -46,6 +46,34 @@ class SearchMixin(ConfluenceClient):
         logger.info(f"Applied spaces filter to query: {cql}")
         return cql
 
+    @staticmethod
+    def _validate_search_response(response: Any, operation: str) -> dict[str, Any]:
+        """Ensure a search API response carries the required 'results' field.
+
+        A well-formed Confluence search response always includes a "results"
+        field (which may legitimately be an empty list). A response missing it
+        entirely indicates an API, network, or processing failure that must
+        surface as an error rather than be silently treated as an empty result
+        set.
+
+        Args:
+            response: The raw response returned by the Confluence API.
+            operation: Human-readable operation name for the error message
+                (e.g. "search", "user search").
+
+        Returns:
+            The validated response dictionary.
+
+        Raises:
+            ValueError: If the response is not a dict or is missing "results".
+        """
+        if not isinstance(response, dict) or "results" not in response:
+            raise ValueError(
+                f"Confluence {operation} returned a malformed response "
+                "missing the 'results' field"
+            )
+        return response
+
     @handle_atlassian_api_errors("Confluence API")
     def search(
         self, cql: str, limit: int = 10, spaces_filter: str | None = None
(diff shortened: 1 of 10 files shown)


WZHBEDZ:
Additional context: preserving the underlying Atlassian error message is important for MCP clients because it gives them actionable signal for recovery decisions. For example, if Jira returns `Field 'customfield_10014' cannot be set. It is not on the appropriate screen, or unknown.`, the client can stop retrying the same payload and instead remove that field, ask for a valid field, or inspect project/issue-type metadata before trying again. With only a generic `Error calling tool '<name>'` response, the client cannot distinguish validation/configuration problems from transient failures.

tests/unit/confluence/test_search.py
```diff
@@ -107,10 +107,8 @@ def test_search_key_error(self, search_mixin):
         # Mock a response missing required keys
         search_mixin.confluence.cql.return_value = {"incomplete": "data"}
 
-        # Act
         results = search_mixin.search("invalid query")
 
-        # Assert
         assert isinstance(results, list)
         assert len(results) == 0
```
sooperset: **BLOCKER:** This test still accepts a malformed Confluence search response as a successful empty result. The issue is specifically about API/search failures being hidden as `[]`, and this response shape never reaches the new `handle_atlassian_api_errors` error path because the parser treats a missing `results` key as an empty list. Required for merge: make this case raise a detailed error, and keep `[]` only for a real `"results": []` response.

sooperset (changes requested):
Thanks for the contribution @WZHBEDZ! The error-preserving FastMCP wrapper is a good direction, and the new Jira/Confluence tool tests cover the client-visible `ToolError` path well.

1. Malformed Confluence search responses still return `[]`, so one of the failure modes from the issue remains hidden as success.
2. The branch is behind current `main`, so please rerun the required verification after the fix lands on the latest base.

tests/unit/confluence/test_search.py
WZHBEDZ: Addressed in bbde8bf. `SearchMixin.search` now validates the API response shape: a response missing the `results` field is treated as a malformed API/network/processing failure and raises a detailed `ValueError` ("Confluence search returned a malformed response missing the 'results' field"), which surfaces through `handle_atlassian_api_errors` as `Error processing search results: ...`. A genuine `{"results": []}` response still returns `[]` (covered by `test_search_with_empty_results`). The former `test_search_key_error` is replaced by `test_search_malformed_response_missing_results`, which now asserts the error is raised.

WZHBEDZ:
Thanks for the review @sooperset! Both points addressed in bbde8bf:

1. **Malformed Confluence search now raises instead of returning `[]`** — `SearchMixin.search` validates the response shape and raises a detailed error when the `results` field is absent (malformed API/network/processing failure), which surfaces via `handle_atlassian_api_errors` as `Error processing search results: Confluence search returned a malformed response missing the 'results' field`. A genuine `{"results": []}` response still returns `[]`. Replaced `test_search_key_error` with `test_search_malformed_response_missing_results` asserting the error path, and kept `test_search_with_empty_results` for the valid-empty case.

2. **Rebased onto latest `main`** — the branch now includes current base commits.

Verification:
```
UV_NATIVE_TLS=true uv run pytest tests/unit/confluence/test_search.py -q  # 52 passed
UV_NATIVE_TLS=true uv run ruff check src/mcp_atlassian/confluence/search.py  # clean
```

WZHBEDZ:
Update: merged latest `main` (ed9a902) — the branch is now even with the current base (0 commits behind). Resolved a conflict in `servers/confluence.py` by keeping the new `_resolve_page_content` helper from `main` alongside the `ErrorPreservingFastMCP` wrapper.

Re-ran the required verification on the updated base:
```
UV_NATIVE_TLS=true uv run pytest \
  tests/unit/utils/test_decorators.py \
  tests/unit/servers/test_confluence_server.py \
  tests/unit/servers/test_jira_server.py \
  tests/unit/confluence/test_search.py -q
# 222 passed

UV_NATIVE_TLS=true uv run ruff check \
  src/mcp_atlassian/servers/confluence.py \
  src/mcp_atlassian/confluence/search.py
# clean (repo ruff ignore set)
```

src/mcp_atlassian/confluence/search.py
```diff
@@ -80,6 +80,16 @@ def search(
         # Execute the CQL search query
         results = self.confluence.cql(cql=cql, limit=limit)
 
+        # A well-formed response always carries a "results" field (which may be
+        # an empty list). A response missing it entirely indicates an API,
+        # network, or processing failure that must surface as an error rather
+        # than be silently treated as an empty result set.
+        if not isinstance(results, dict) or "results" not in results:
```
sooperset: **BLOCKER:** This validation only covers content search. `search_user` still passes `results or {}` to `ConfluenceUserSearchResults.from_api_response`, and the existing `test_search_user_edge_cases` expects both `None` and `{"incomplete": "data"}` to return `[]`. That silently reports a malformed Cloud response as no users found; the DC helper has the same missing-`results` behavior. Required for merge: validate both user-search response shapes, keep only a real `{"results": []}` as an empty success, and add regression coverage for Cloud and DC malformed responses.

sooperset (changes requested):
Thanks for the contribution @WZHBEDZ! The content-search response guard cleanly separates malformed replies from genuine empty results, and the Cloud/DC verification is strong.

1. User search still converts malformed responses into a successful empty list in both Cloud and DC paths.

src/mcp_atlassian/confluence/search.py
WZHBEDZ: Addressed in a3b1e9f. Extracted a shared `_validate_search_response` helper and applied it to **both** user-search paths:

- **Cloud** (`search_user`): validates the `rest/api/search/user` response and drops the old `results or {}` fallback, so a response missing `results` now raises `Error processing search_user results: Confluence user search returned a malformed response missing the 'results' field` instead of returning `[]`.
- **Server/DC** (`_search_user_server_dc`): validates each group-member page response the same way.

A genuine `{"results": []}` still returns an empty list on both paths. Replaced the old `test_search_user_edge_cases` (which asserted malformed → `[]`) with parametrized regression tests covering `{"incomplete": "data"}` and `None` for Cloud and Server/DC, plus genuine-empty success cases (`test_server_dc_empty_results_returns_empty`, existing `test_search_user_with_empty_results`). The same helper also backs content search.

WZHBEDZ:
Addressed the user-search blocker in a3b1e9f (rebased onto the latest `main`, which now includes the security-hardening changes).

**Change:** a shared `_validate_search_response` helper now guards content search **and** both user-search paths (Cloud CQL endpoint + Server/DC group-member fallback). Malformed responses missing the `results` field raise a detailed error instead of being reported as "no users found"; a genuine `{"results": []}` still returns `[]`. Added Cloud + Server/DC malformed regression tests and genuine-empty success coverage; removed the old `test_search_user_edge_cases` that asserted malformed → `[]`.

Verification:
```
UV_NATIVE_TLS=true uv run pytest \
  tests/unit/confluence/test_search.py \
  tests/unit/utils/test_decorators.py \
  tests/unit/servers/test_confluence_server.py \
  tests/unit/servers/test_jira_server.py -q
# 262 passed

UV_NATIVE_TLS=true uv run ruff check \
  src/mcp_atlassian/confluence/search.py \
  tests/unit/confluence/test_search.py
# clean (repo ruff ignore set)
```

PR #1326, opened 2026-05-11
fix(auth): set explicit User-Agent on Jira/Confluence sessions

StefanWinter-ViREQ:
## Summary

- Some WAFs in front of Jira/Confluence Server/DC reject the default `python-requests/X.Y` User-Agent and return **403**, even when the bearer token is valid. The error surfaces in mcp-atlassian as a misleading `Authentication failed for Jira API (403). Token may be expired or invalid` — although the same PAT works via `curl`.
- `JiraClient` and `ConfluenceClient` now set `User-Agent: mcp-atlassian/<version>` on the session right after the underlying `atlassian` client is created, **before** `_apply_custom_headers()`, so user-supplied `JIRA_CUSTOM_HEADERS=User-Agent=...` / `CONFLUENCE_CUSTOM_HEADERS=User-Agent=...` still wins.
- Adds `utils/user_agent.py::get_default_user_agent()` (uses `importlib.metadata.version("mcp-atlassian")`, falls back to `0.0.0` when not installed — same pattern as the package `__version__`).

## Reproduction (Jira DC 9.12.4 behind a WAF)

```python
from atlassian import Jira
jira = Jira(url="https://portal.example.com", token=PAT, cloud=False)
jira.get_issue("PROJ-1")
# requests.exceptions.HTTPError: 403 Client Error  (default UA: python-requests/2.x)

# Workaround (now built-in by this PR):
jira._session.headers["User-Agent"] = "mcp-atlassian/0.21.1"
jira.get_issue("PROJ-1")   # 200 OK
```

Direct `curl` with the same URL + Bearer token consistently returns `200`. The WAF blocks only the default `python-requests/X.Y` UA — any identifiable UA (`curl/*`, `Mozilla/*`, `mcp-atlassian/*`, …) is allowed.

## Test plan

- [x] `uv run pytest tests/unit/jira/test_client.py tests/unit/confluence/test_client.py tests/unit/jira/test_custom_headers.py tests/unit/confluence/test_custom_headers.py` — all green
- [x] `pre-commit run` (ruff, ruff-format, mypy) — green
- [x] Pre-existing failures on `main` (`test_stdio_lifecycle`, `test_date::…boundary_max_valid`, `test_media::…[zip]`) verified unrelated to this change
- [x] New unit tests:
  - `test_jira_client_sets_default_user_agent` / `test_confluence_client_sets_default_user_agent` — default UA is set on the session
  - `test_jira_client_custom_user_agent_overrides_default` / equivalent for Confluence — `custom_headers` still wins
- [x] Updated `test_no_custom_headers_applied` (Jira & Confluence) to expect the default UA being present

### DIFF
--- a/src/mcp_atlassian/confluence/client.py
+++ b/src/mcp_atlassian/confluence/client.py
@@ -143,7 +143,7 @@ class ConfluenceClient:
             log_config_param(logger, "Confluence", "NO_PROXY", self.config.no_proxy)
 
         # Set an explicit User-Agent so requests aren't blocked by WAFs that
-        # reject the default ``python-requests/X.Y`` header (#TBD). User-supplied
+        # reject the default ``python-requests/X.Y`` header. User-supplied
         # custom headers below can still override this.
         self.confluence._session.headers["User-Agent"] = get_default_user_agent()
 
--- a/src/mcp_atlassian/confluence/client.py
+++ b/src/mcp_atlassian/confluence/client.py
@@ -11,6 +11,7 @@ from ..exceptions import MCPAtlassianAuthenticationError
 from ..utils.logging import get_masked_session_headers, log_config_param, mask_sensitive
 from ..utils.oauth import configure_oauth_session
 from ..utils.ssl import configure_ssl_verification
+from ..utils.user_agent import get_default_user_agent
 from .config import ConfluenceConfig
 
 # Configure logging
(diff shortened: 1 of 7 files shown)


sooperset:
The User-Agent default is well scoped, and I like that the tests cover both the built-in header and the custom override paths for Jira and Confluence. Merged.

PR #1335, opened 2026-05-18
feat(jira): expose required transition fields and worklog support

0w1afk:
## Description

Jira transitions can require values from a transition screen, but agents previously could not discover those requirements or send transition-time update operations such as worklogs.

This PR now:

- requests `expand=transitions.fields` and returns `has_screen` plus compact `required_fields` metadata from `jira_get_transitions`
- adds `update_data` to `jira_transition_issue` and forwards it to the Jira REST `update` object
- combines explicit comments with caller-supplied update operations without dropping either
- sends fields and update operations atomically when falling back to a direct transition ID
- exposes `has_screen` and `is_conditional` from `JiraTransition`
- documents the discovery and transition workflow

Example workflow:

```text
jira_get_transitions
  -> required_fields includes resolution or timetracking
jira_transition_issue(
  fields={"resolution": {"name": "Fixed"}},
  update_data={"worklog": [{"add": {"timeSpent": "1h"}}]}
)
```

Jira still validates that each submitted field or update operation is permitted by the selected transition and its screen configuration.

## Verification

- `uv run pre-commit run --all-files` — passed, including Ruff and strict mypy
- `uv run pytest tests/unit/ -q` — 3072 passed, 5 skipped
- focused transition, model, and server tests — 157 passed
- `uv run pytest tests/integration/ -q` — 23 skipped by environment markers
- `uv run pytest tests/e2e/cloud --cloud-e2e -q` — 82 passed, 15 skipped
- `uv run pytest tests/e2e --dc-e2e -q` — 94 passed, 98 skipped

## Checklist

- [x] Current `main` merged into the contributor branch
- [x] Regression coverage added for discovery, MCP parsing, comment merging, and atomic direct-ID updates
- [x] Cloud and Data Center behavior rechecked against live test instances
- [x] Contributor commits and authorship preserved

### DIFF
--- a/src/mcp_atlassian/jira/transitions.py
+++ b/src/mcp_atlassian/jira/transitions.py
@@ -148,14 +148,38 @@ class TransitionsMixin(JiraClient, IssueOperationsProto, UsersOperationsProto):
             allowed_values = field_data.get("allowedValues")
             if isinstance(allowed_values, list):
                 field_info["allowed_values"] = [
-                    {"id": str(v.get("id", "")), "name": v.get("name", "")}
+                    TransitionsMixin._normalize_allowed_value(v)
                     for v in allowed_values
-                    if isinstance(v, dict)
                 ]
 
             required_fields.append(field_info)
 
         return required_fields
+
+    @staticmethod
+    def _normalize_allowed_value(value: Any) -> dict[str, str]:
+        """
+        Normalize a single allowed value to {id, name, value} dict.
+
+        Jira returns allowed values in multiple shapes:
+        - Scalar (str, int): ``"Fixed"``
+        - ``{id, name}``: ``{"id": "1", "name": "Fixed"}``
+        - ``{id, value}`` (custom field options): ``{"id": "1", "value": "Fixed"}``
+
+        Args:
+            value: A raw allowed value from the Jira API.
+
+        Returns:
+            A dict with ``id``, ``name``, and ``value`` keys.
+        """
+        if not isinstance(value, dict):
+            scalar = str(value)
+            return {"id": scalar, "name": scalar, "value": scalar}
+
+        vid = str(value.get("id", value.get("optionId", "")))
+        vname = str(value.get("name", value.get("value", "")))
+        vvalue = str(value.get("value", value.get("name", vname)))
+        return {"id": vid, "name": vname, "value": vvalue}
 
     def get_transitions(
         self, issue_key: str, expand: str | None = None
--- a/src/mcp_atlassian/jira/transitions.py
+++ b/src/mcp_atlassian/jira/transitions.py
@@ -148,8 +148,7 @@ class TransitionsMixin(JiraClient, IssueOperationsProto, UsersOperationsProto):
             allowed_values = field_data.get("allowedValues")
             if isinstance(allowed_values, list):
                 field_info["allowed_values"] = [
-                    TransitionsMixin._normalize_allowed_value(v)
-                    for v in allowed_values
+                    TransitionsMixin._normalize_allowed_value(v) for v in allowed_values
                 ]
 
             required_fields.append(field_info)
--- a/src/mcp_atlassian/jira/transitions.py
+++ b/src/mcp_atlassian/jira/transitions.py
@@ -20,7 +20,7 @@ class TransitionsMixin(JiraClient, IssueOperationsProto, UsersOperationsProto):
     def get_available_transitions(
         self,
         issue_key: str,
-        expand: str | None = None,
+        expand_fields: bool = False,
     ) -> list[dict[str, Any]]:
         """
         Get the available status transitions for an issue.
@@ -41,10 +41,9 @@ class TransitionsMixin(JiraClient, IssueOperationsProto, UsersOperationsProto):
 
         Args:
             issue_key: The issue key (e.g. 'PROJ-123')
-            expand: Optional expand parameter. Pass
-                ``"transitions.fields"`` to include required
-                field metadata (allowed values, schema) with
-                each transition. Defaults to None (lightweight).
+            expand_fields: Whether to request required field metadata
+                (allowed values, schema) for each transition. Defaults
+                to False, which keeps the response lightweight.
 
         Returns:
             List of available transitions with id, name,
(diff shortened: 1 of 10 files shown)


src/mcp_atlassian/jira/transitions.py
```diff
@@ -21,20 +21,26 @@ def get_available_transitions(self, issue_key: str) -> list[dict[str, Any]]:
...
 
         Raises:
             MCPAtlassianAuthenticationError: If authentication fails
                 with the Jira API (401/403)
             Exception: If there is an error getting transitions
         """
         try:
-            transitions_data = self.jira.get_issue_transitions(issue_key)
+            transitions_data = self.get_transitions(
+                issue_key, expand="transitions.fields"
```
sooperset: **IMPORTANT:** This leaves the existing needs-work concern unresolved. Every call now requests full transition field metadata, including callers that only need IDs and names. Workflows with large option sets pay the extra response-size and latency cost on every lookup. Required for merge: make expansion opt-in at the library/tool boundary, or otherwise keep a lightweight default, and cover both modes.

src/mcp_atlassian/jira/transitions.py
```diff
@@ -74,7 +88,51 @@ def get_available_transitions(self, issue_key: str) -> list[dict[str, Any]]:
...
+            schema = field_data.get("schema")
+            if isinstance(schema, dict):
+                field_info["schema"] = schema
+
+            allowed_values = field_data.get("allowedValues")
+            if isinstance(allowed_values, list):
+                field_info["allowed_values"] = [
+                    {"id": str(v.get("id", "")), "name": v.get("name", "")}
+                    for v in allowed_values
+                    if isinstance(v, dict)
```
sooperset: **BLOCKER:** Jira doesn't guarantee `{id, name}` objects here. The Cloud schema permits scalar choices, and this repository already models custom options as `{id, value}`. This comprehension drops scalars and turns `{id, value}` into an empty `name`, so agents can't discover what to submit for a required select field. Required for merge: preserve these response shapes and add regression cases for scalar and `value`-labelled options.

sooperset (changes requested):
Thanks for the contribution @0w1afk! The atomic direct-ID request and the comment/update merge coverage are solid.

1. Allowed-value normalization drops valid Jira metadata shapes.
2. The lightweight transition-list path is still lost because field expansion is unconditional.
3. Neither edition's e2e transition calls `update_data`, so the new REST payload still lacks cross-edition evidence.

alessandro-redi:
Hi, thanks for the work on this PR! I'd like to raise a related issue that affects the same jira_get_transitions tool.

Currently, the tool calls get_available_transitions() internally, which strips the Jira API's to object down to a single string field (to_status), discarding to.id and the full to.statusCategory object. This means agents consuming the tool cannot determine the destination status category (e.g. done, indeterminate, new) without a separate API call.

The TransitionsMixin class already has a get_transitions() method that calls get_issue_transitions_full() and returns the complete raw payload — it just isn't wired up to the MCP tool.

Would it be possible to extend this PR (or a follow-up) so that jira_get_transitions returns the full to object alongside the new has_screen / required_fields fields you're already adding? 
Something like:

```json
{
  "id": "41",
  "name": "Send to some team",
  "to": {
    "id": "3",
    "name": "In Progress",
    "statusCategory": {
      "id": 4,
      "key": "indeterminate",
      "name": "In Progress"
    }
  },
  "has_screen": false,
  "is_conditional": false
}
```

This would allow agents to make routing decisions based on destination status category without needing extra round-trips to the Jira API.

Thanks!

Clarkkkk:
Need this too, can it be merged now?

erichll:
**Server/DC real-world case: a transition screen with a mandatory "Time Spent" (worklog) field - the `update_data` passthrough in this PR (#1335) is exactly what is missing**

Environment: mcp-atlassian **0.23.1** (current PyPI release) against **Jira Server/DC** (REST v2, `is_cloud == false`), instance localized to zh_CN. One transition in our workflow has a screen that requires the "Time Spent" worklog field. Logging the worklog *before* the transition with `jira_add_worklog` does not satisfy that screen: Jira expects the worklog inside the transition request (`update.worklog[].add`).

`jira_transition_issue` with `fields` only is rejected by Jira with its required-field error for the worklog field ("please fill in the time spent on resolving the task"). Every shape expressible through the current `fields` parameter is rejected:

| `fields` payload | Jira error |
| --- | --- |
| `{"worklog": {"started": "...", "timeSpentSeconds": 60}}` | `Field does not support update 'worklog'` |
| `{"worklog": [{"add": {...}}]}` | `Field does not support update 'worklog'` |
| `{"worklog": "1m"}` | `Field does not support update 'worklog'` |
| `{"update": {"worklog": [{"add": {...}}]}}` | `Field 'update' cannot be set` |
| `{"timetracking": {"originalEstimate": "1h", "remainingEstimate": "0m"}}` | `Field 'timetracking' cannot be set. It is not on the appropriate screen, or unknown.` |

`jira_update_issue` cannot work around it either: `fields={"status": "<target status name>"}` returns `Could not find transition to status '<name>'. Available transitions: <transition name>` (same family as #455 - `atlassian-python-api`'s `get_issue_transitions` returns `to` as a *string*, while `TransitionsMixin.get_available_transitions` only handles `to` as a dict / `to_status` / `status`), and even when a numeric transition id is used, `set_issue_status_by_transition_id` posts only `{"transition": {"id": ...}}`.

The shape that would unblock this workflow:

```json
jira_transition_issue(
  issue_key="PROJ-123",
  transition_id="21",
  fields={"resolution": {"name": "Fixed"}},
  update_data={"worklog": [{"add": {"started": "2026-01-01T09:00:00.000+0800", "timeSpentSeconds": 1200}}]}
)
```

Two things I would like to confirm for this PR:

1. On Server/DC the payload goes out through `self.jira.post(url, data=payload)` against `/rest/api/2/issue/{key}/transitions` - does `update_data` get forwarded verbatim (that is, the full `{"update": {...}}` object), the same way `update.comment` already is?
2. Is Server/DC explicitly covered by the verification for this PR? The description mentions DC e2e, but I could not tell whether a transition screen with a *required worklog* is part of it.

Until this lands, the only way to complete such a transition is a human doing it in the Jira UI - which is exactly the step I would like to automate. Thanks for the work on this PR.

HarrievG:
Hi All, 

I also just hit this , everything else is working, and other transitions that do not have a mandatory "Time-Spent" filed work great. 

There seems to be a missing tooloption that is able to do Jira API call that supports transition ` update.worklog `.

The offending transition window that cannot be automated is this:

image

From my perspective, if this one is fixed, it seems to be _fully_ functional for day to day use. Please consider merging / fixing upstream.

PR #1339, opened 2026-05-21
fix(jira): use correct API version for version creation on Server/DC

### LINKED ISSUE #1338: [Bug]: jira_create_version fails on Server/DC — hardcoded API v3 endpoint
### Prerequisites

### Bug Description

The create_version method in client.py hardcodes /rest/api/3/version, which only exists on Jira Cloud. When used against a Jira Server/Data Center instance, the API call fails because Server/DC requires /rest/api/2/version.

### Steps to Reproduce

Configure mcp-atlassian against a Jira Server / Data Center instance
Call jira_create_version with valid project key and version name
Observe the API call fails

### Expected Behavior

jira_create_version should work on both Cloud and Server/Data Center deployments, using the appropriate API version for each.

### Actual Behavior

```shell
The call fails on Server/DC because the endpoint /rest/api/3/version does not exist. Only /rest/api/2/version is available on Server/DC.
The API behavior is otherwise the same so we just need to replace the number to make it work.
```

### mcp-atlassian Version

0.21.1

### Installation Method

Docker

### Operating System

### Python Version

3.13

### Atlassian Instance Type

Jira Server / Data Center

### Client Application

Github Copilot (VSCode)

### Additional Context

The relevant code in client.py is l385:
`result = self.jira.post("/rest/api/3/version", json=payload)`

A fix I can propose, to dynamically selects the API version based on config.is_cloud:
```python
api_version = "3" if self.config.is_cloud else "2"
url = self.jira.resource_url("version", api_version=api_version)
result = self.jira.post(url, json=payload)
```

bahumbert:
## Description

The create_version method hardcoded /rest/api/3/version which only works on Cloud. Server/Data Center requires API v2, the behavior is the same after that. 
This dynamically selects the API version based on config.is_cloud.

Fixes: #1338

## Changes

Use `self.config.is_cloud` to select API v3 (Cloud) or v2 (Server/DC) for version creation
Use `self.jira.resource_url()` to build the endpoint dynamically instead of a hardcoded path
Add parametrized unit test covering both Cloud and Server/DC paths

## Testing

- [x] Manual checks performed: `this patch-fix is deployed on our production for a Jira DC`

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/client.py
+++ b/src/mcp_atlassian/jira/client.py
@@ -382,7 +382,9 @@ class JiraClient:
         if description:
             payload["description"] = description
         logger.info(f"Creating Jira version: {payload}")
-        result = self.jira.post("/rest/api/3/version", json=payload)
+        api_version = "3" if self.config.is_cloud else "2"
+        url = self.jira.resource_url("version", api_version=api_version)
+        result = self.jira.post(url, json=payload)
         if not isinstance(result, dict):
             error_message = f"Unexpected response from Jira API: {result}"
             raise ValueError(error_message)
--- a/tests/e2e/cloud/test_jira_cloud_operations.py
+++ b/tests/e2e/cloud/test_jira_cloud_operations.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import uuid
 
 import pytest
+import requests
 from requests.exceptions import HTTPError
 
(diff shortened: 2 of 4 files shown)


sooperset:
The API-version split for version creation is nicely scoped, and the Cloud/DC coverage made the behavior easy to verify. Merged, thanks Bastien.

PR #1349, opened 2026-05-22
feat(jira): add update_version tool

prochazkal:
## Summary

Adds a `jira_update_version` MCP tool plus the underlying `update_version` client method and `update_project_version` mixin wrapper, backed by `PUT /rest/api/2/version/{id}`.

The tool sends only the fields the caller provides, so a single attribute (e.g. `archived`) can be toggled without overwriting the rest of the version. It mirrors the shape and tags of the existing `create_version` / `batch_create_versions` tools.

## Motivation

The version-related tools today only cover read + create:

- `jira_get_project_versions`
- `jira_create_version`
- `jira_batch_create_versions`

There's no way to update, archive, unarchive, release, rename, or shift the dates on a version — those operations all require falling back to the Jira UI or raw REST. This PR fills that gap with the smallest viable change.

Real-world driver: archiving a large batch of historical build-candidate versions in one project — a task that's trivial via `PUT /rest/api/2/version/{id}` but currently unreachable from the MCP.

## What's in the PR

- **Client (`src/mcp_atlassian/jira/client.py`)**: new `update_version(version_id, name=None, description=None, start_date=None, release_date=None, archived=None, released=None)`. Builds the payload from non-`None` arguments only; raises `ValueError` if no field is supplied.
- **Mixin (`src/mcp_atlassian/jira/projects.py`)**: thin `update_project_version` wrapper mirroring `create_project_version`.
- **Server (`src/mcp_atlassian/servers/jira.py`)**: `@jira_mcp.tool` named `update_version`, tagged `jira / write / toolset:jira_projects`, `destructiveHint=True`, `@check_write_access`.
- **Tests (`tests/unit/jira/test_projects.py`)**: three new tests covering single-field update, all-fields update, and error propagation — same pattern as the existing `create_project_version` tests.
- **Toolset count assertion (`tests/unit/utils/test_toolsets.py`)**: bumped from 49 → 50.

## Endpoint choice: v2 vs v3

This PR uses `/rest/api/2/version/{id}` rather than v3 (which the existing `create_version` uses). Reason, per Atlassian's own documentation:

- **Jira Cloud** publishes both v2 and v3. Neither is deprecated; the only documented difference is rich-text handling (v3 uses ADF for body fields like description/comments). The project-versions endpoint has no rich-text fields, so v2 and v3 payloads are byte-identical for this surface.
- **Jira Server / Data Center** documents v2 only. The DC platform reference states verbatim *"The current API version is `2`"* (https://developer.atlassian.com/server/jira/platform/about-the-jira-server-rest-apis/, https://docs.atlassian.com/software/jira/docs/api/REST/9.17.0/). There is no v3 namespace on DC — not "partial," literally no v3 paths are documented. Atlassian has made no public commitment to bring v3 to DC; the open feature request is [JRASERVER-70688](https://jira.atlassian.com/browse/JRASERVER-70688), in *Gathering Interest* with 105 votes.

Empirically: `PUT /rest/api/3/version/{id}` returns **HTTP 405 Method Not Allowed** on a current DC instance, and `GET /rest/api/3/version/{id}` 302-redirects to v2.

Picking v2 keeps the tool working on both Cloud and DC by Atlassian's documented contract, with no `is_cloud` branch needed. (`create_version`'s use of v3 may have a similar latent issue on DC — out of scope for this PR, but a candidate for a small follow-up.)

## Test plan

- [x] `uv run pytest tests/unit/` — 2569 passed (3 pre-existing Windows-only flakes around `mimetypes`, `stdio_lifecycle`, and a timestamp boundary are deselected; verified red on `main` before this change, so unrelated).
- [x] `pre-commit` Ruff lint clean; `ruff-format` applied.
- [x] `mypy` against the touched files reports zero new errors (same 21 pre-existing errors on `main`, all from missing-stub noise unrelated to this change).
- [x] **End-to-end verified against a real Jira Server / Data Center instance**: archived a version, called `update_version(version_id, archived=false)`, confirmed `archived: false` in the response and on re-listing project versions. Re-archived afterwards to leave state untouched.

## Notes

- This PR is intentionally scoped to update only. Batch-update and delete are useful follow-ups but kept out to keep the diff small and reviewable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Documentation is also available in [llms.txt format](https://llmstxt.org/), whic
 | `jira_update_issue` - Update issues | `confluence_update_page` - Update pages |
 | `jira_transition_issue` - Change status | `confluence_add_comment` - Add comments |
 
-**78 tools total** — See [Tools Reference](https://mcp-atlassian.soomiles.com/docs/tools-reference) for the complete list.
+**80 tools total** — See [Tools Reference](https://mcp-atlassian.soomiles.com/docs/tools-reference) for the complete list.
 
 ## Security
 
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Documentation is also available in [llms.txt format](https://llmstxt.org/), whic
 | `jira_update_issue` - Update issues | `confluence_update_page` - Update pages |
 | `jira_transition_issue` - Change status | `confluence_add_comment` - Add comments |
 
-**77 tools total** — See [Tools Reference](https://mcp-atlassian.soomiles.com/docs/tools-reference) for the complete list.
+**78 tools total** — See [Tools Reference](https://mcp-atlassian.soomiles.com/docs/tools-reference) for the complete list.
 
 ## Security
 
--- a/docs/tools-reference.mdx
+++ b/docs/tools-reference.mdx
@@ -1,9 +1,9 @@
 ---
 title: "Tools Reference"
-description: "Overview of all 78 MCP tools for Jira and Confluence — organized by category with quick links"
+description: "Overview of all 80 MCP tools for Jira and Confluence — organized by category with quick links"
 ---
 
-MCP Atlassian provides **78 tools** for interacting with Jira and Confluence. Tools are organized by category below.
+MCP Atlassian provides **80 tools** for interacting with Jira and Confluence. Tools are organized by category below.
 
 ## Jira Tools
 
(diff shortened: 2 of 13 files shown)


prochazkal:
Not stale, waiting for review

sooperset:
`jira_update_version` landed cleanly because the partial-update shape only sends fields callers provide, including explicit `False` booleans. Thanks for filling the Jira version-management gap and documenting the v2 endpoint choice clearly.

PR #1355, opened 2026-05-26
fix(jira): route transition comments through the correct API version

### LINKED ISSUE #1262: [Bug]: jira_transition_issue fails when transition screen requires a comment on Jira Cloud — comment sent as plain text instead of ADF
## Bug Description

`jira_transition_issue` fails on Jira Cloud when the target transition screen has a required Comment field. The tool sends the `comment` parameter as a plain text string, but the Jira Cloud v3 REST API requires the comment in Atlassian Document Format (ADF). This makes it impossible to complete any transition that has Comment as a required screen field.

## Steps to Reproduce

1. Have a Jira Cloud project with a workflow transition (e.g., "Discarded") that requires both a Resolution and a Comment field on its transition screen.
2. Call `jira_transition_issue` with the `comment` parameter set and the resolution in `fields`:
   - `transition_id`: the target transition
   - `fields`: `{"resolution": {"name": "Duplicate"}}`
   - `comment`: `"Closing as duplicate of PROJ-98."`
3. The call fails with: **"Operation value must be an Atlassian Document (see the Atlassian Document Format)"**
4. Omitting the `comment` parameter entirely fails with: **"Field Comment is required."**
5. Attempting to pass `comment` inside `fields` fails with: **"Field does not support update 'comment'"**

There is no combination of parameters that allows the transition to succeed.

## Expected Behavior

The tool should convert the Markdown/plain text `comment` parameter to ADF before sending it to the Jira Cloud transitions API, similar to how `jira_add_comment` successfully handles Markdown input. The transition should complete with both the resolution and comment set.

## Actual Behavior

The `comment` parameter is sent as a raw string to the v3 API, which rejects it because it expects an ADF document. The transition cannot be completed through the tool.

## Atlassian Instance Type

- [ ] Jira Server / Data Center
- [x] Jira Cloud

## Additional Context

- `jira_add_comment` handles Markdown-to-ADF conversion correctly, so the conversion logic exists in the codebase. It just isn't applied to transition comments.
- `jira_update_issue` can set the resolution field independently, so a partial workaround exists (set resolution + add comment separately), but the status itself cannot be changed without completing the transition.
- An alternative fix would be to use the v2 API endpoint for transitions, which accepts plain text comments.

tony-davis:
## Summary

Closes #1262.

Fixes `jira_transition_issue` on Jira Cloud when a transition includes a comment, including workflows whose transition screen requires Comment. Jira Cloud comments are converted to Atlassian Document Format (ADF), so the atomic transition request now uses REST API v3. Jira Server and Data Center continue to send wiki-markup comments through REST API v2.

## Changes

- Send the transition ID, fields, and update operations in one atomic request.
- Route Cloud transition payloads through the existing REST v3 helper so ADF comments are accepted.
- Keep Server/Data Center transition payloads on the dependency-provided REST v2 URL.
- Add unit regressions for Cloud ADF and Server/Data Center wiki-markup payloads.
- Exercise comment-bearing transitions in both Cloud and Data Center E2E suites.
- Retain defensive transition-response handling and make the touched module clean under strict mypy and Ruff checks.

## Why the endpoint differs by deployment

The original proposal forced REST v2 everywhere. That is no longer correct because the current Cloud formatter produces ADF, which REST v2 rejects. Cloud must use v3 for the ADF body, while Server/Data Center must retain v2 for wiki markup.

## Verification

- `uv run pre-commit run --all-files` — passed
- `uv run pytest tests/unit/jira/ -q` — 883 passed
- `uv run pytest tests/integration/ -q` — 23 skipped by environment markers
- `uv run pytest tests/e2e/cloud/test_adf_write.py tests/e2e/cloud/test_jira_*.py --cloud-e2e -q` — 41 passed, 8 skipped
- `uv run pytest tests/e2e --dc-e2e -q` — 98 passed, 103 skipped
- Full Cloud suite — the Jira transition regression passed; 86 passed and 15 skipped overall, with one unrelated existing Confluence search normalization failure

## Contributor credit

The contributor commits remain in the branch unchanged. The maintainer follow-up updates the endpoint selection for the current ADF behavior and adds live cross-edition coverage.

### DIFF
--- a/tests/e2e/cloud/test_jira_cloud_operations.py
+++ b/tests/e2e/cloud/test_jira_cloud_operations.py
@@ -290,7 +290,11 @@ class TestJiraCloudTransitions:
         if target_id is None:
             target_id = transitions[0]["id"]
 
-        jira_fetcher.transition_issue(issue.key, target_id)
+        jira_fetcher.transition_issue(
+            issue.key,
+            target_id,
+            comment=f"Cloud transition comment {uid}",
+        )
 
         updated = jira_fetcher.get_issue(issue.key)
         assert updated.status is not None
--- a/tests/e2e/test_jira_dc_operations.py
+++ b/tests/e2e/test_jira_dc_operations.py
@@ -271,7 +271,11 @@ class TestJiraDCTransitions:
         if target_id is None:
             target_id = transitions[0]["id"]
 
-        jira_fetcher.transition_issue(issue.key, target_id)
+        jira_fetcher.transition_issue(
+            issue.key,
+            target_id,
+            comment=f"Data Center transition comment {uid}",
+        )
 
         updated = jira_fetcher.get_issue(issue.key)
         assert updated.status is not None
--- a/src/mcp_atlassian/jira/transitions.py
+++ b/src/mcp_atlassian/jira/transitions.py
@@ -34,11 +34,12 @@ class TransitionsMixin(JiraClient, IssueOperationsProto, UsersOperationsProto):
             Exception: If there is an error getting transitions
         """
         try:
-            transitions_data = self.jira.get_issue_transitions(issue_key)
+            transitions_data: object = self.jira.get_issue_transitions(issue_key)
+            if not isinstance(transitions_data, list):
+                return []
             result: list[dict[str, Any]] = []
 
             for transition in transitions_data:
-                # Skip non-dict transitions
                 if not isinstance(transition, dict):
                     continue
 
@@ -72,7 +73,8 @@ class TransitionsMixin(JiraClient, IssueOperationsProto, UsersOperationsProto):
         except Exception as e:
             error_msg = f"Error getting transitions for {issue_key}: {str(e)}"
             logger.error(error_msg)
-            raise Exception(f"Error getting transitions: {str(e)}") from e
+            public_error = f"Error getting transitions: {str(e)}"
+            raise Exception(public_error) from e
 
     def get_transitions(self, issue_key: str) -> list[dict[str, Any]]:
         """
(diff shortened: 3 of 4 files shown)


sooperset:
Thanks @tony-davis, unifying the transition, fields, and comment into one atomic request was the right foundation for fixing required transition screens. The focused regression tests made the Cloud API mismatch clear and helped carry the fix across both deployment types.

tests/e2e/cloud/test_jira_cloud_operations.py
```diff
@@ -290,7 +290,11 @@ def test_transition_lifecycle(
         if target_id is None:
             target_id = transitions[0]["id"]
 
-        jira_fetcher.transition_issue(issue.key, target_id)
+        jira_fetcher.transition_issue(
+            issue.key,
+            target_id,
+            comment=f"Cloud transition comment {uid}",
```
sooperset, line 296: **BLOCKER:** This only proves the transition endpoint returns success. The configured Cloud workflow has no required transition fields, and Jira silently drops this optional comment; the earlier read-back assertion returned an empty list. Required for merge: exercise a workflow where Comment is required, then assert the transition succeeds and the comment is present afterward. Without that, the exact #1262 failure remains unverified.

tests/e2e/test_jira_dc_operations.py
```diff
@@ -271,7 +271,11 @@ def test_transition_lifecycle(
         if target_id is None:
             target_id = transitions[0]["id"]
 
-        jira_fetcher.transition_issue(issue.key, target_id)
+        jira_fetcher.transition_issue(
+            issue.key,
+            target_id,
+            comment=f"Data Center transition comment {uid}",
```
sooperset, line 277: **BLOCKER:** The Data Center run has the same blind spot: an earlier read-back assertion found no matching comment, but removing that assertion lets the test pass on the status change alone. Because this shared endpoint wrapper changes both deployment paths, use a transition screen that accepts Comment and assert the stored wiki-markup comment after the transition.

sooperset (changes requested):
Thanks for the contribution @tony-davis! The Cloud v3 and DC v2 split matches each deployment's comment format, and the payload unit tests cover that branch clearly.

1. The live tests currently pass even when Jira silently discards the transition comment, so they don't verify the reported failure.

PR #1361, opened 2026-05-29
fix(jira): don't italicize intraword underscores in markdown_to_jira

wownepiece:
## Problem

`JiraPreprocessor.markdown_to_jira` runs a single emphasis pass that treats `_` exactly like `*`:

```python
re.sub(r"([*_]+)(.*?)\1", ...)   # preprocessing/jira.py
```

This pairs underscores anywhere on a line, which violates CommonMark — **underscores between two word characters are literal, not emphasis** (only `*` does intraword emphasis). Because the Jira wiki renderer italicizes any `_word_` span, the result is that identifiers get corrupted with spurious italics:

| Markdown input | Renders in Jira as |
|---|---|
| `baseline_samples_5m` | baseline*samples*5m (italic "samples") |
| `customfield_10101` | customfield*10101 |
| `netflow_ap and sre_analytics` | netflow*ap and sre*analytics (cross-token italic span) |

This is a frequent foot-gun for technical content (code identifiers, custom-field IDs, table/MV names) posted via `jira_create_issue` / `jira_update_issue` / `jira_add_comment`. It's in the same family as the existing underscore/code-span reports.

## Fix

Escape intraword underscores (`\_`) before the emphasis conversion so the Jira renderer treats them literally:

```python
line = re.sub(r"(?<=\w)_(?=\w)", r"\\_", line)
```

- Genuine word-boundary `_emphasis_` still converts to Jira italic.
- All `*`/`**` handling is unchanged.
- Inline-code spans are already extracted to placeholders before this pass, so `` `find_provider_by_url` `` is untouched.
- Placeholders are `\x00PREFIX<n>\x00` (null-flanked, no `_`), so the rule can't disturb them.

## Tests

Adds three regression tests in `tests/unit/preprocessing/test_preprocessing.py`:

- `test_markdown_to_jira_preserves_intraword_underscores` — snake_case / customfield IDs / cross-token pairing
- `test_markdown_to_jira_word_boundary_underscore_still_italicizes` — genuine `_emphasis_` still works
- `test_markdown_to_jira_underscore_in_inline_code_untouched` — `` `ident_with_underscores` `` stays monospace

Full preprocessing suite: **116 passed** (3 new, 0 regressions).

## Scope / follow-up

This targets the unambiguous **intraword** case (the common identifier breakage). Leading-underscore tokens at a word boundary (`_v2`, `_local`) are genuinely ambiguous in Markdown and best expressed as inline code; a full CommonMark left/right-flanking implementation for `_` would be a larger, separate change.

### DIFF
--- a/src/mcp_atlassian/preprocessing/jira.py
+++ b/src/mcp_atlassian/preprocessing/jira.py
@@ -480,6 +480,34 @@ class JiraPreprocessor(BasePreprocessor):
             flags=re.MULTILINE,
         )
 
+        markdown_url_targets: list[str] = []
+
+        def store_markdown_url_target(target: str) -> str:
+            placeholder = f"\x00MARKDOWNURL{len(markdown_url_targets)}\x00"
+            markdown_url_targets.append(target)
+            return placeholder
+
+        def protect_markdown_link_target(match: re.Match[str]) -> str:
+            return (
+                match.group(1)
+                + store_markdown_url_target(match.group(2))
+                + match.group(3)
+            )
+
+        def protect_markdown_autolink_target(match: re.Match[str]) -> str:
+            return "<" + store_markdown_url_target(match.group(1)) + ">"
+
+        output = re.sub(
+            r"(!?\[[^\]\n]*\]\()([^)]+)(\))",
+            protect_markdown_link_target,
+            output,
+        )
+        output = re.sub(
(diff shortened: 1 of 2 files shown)


sooperset:
The URL-target follow-up is the key detail here: it keeps snake_case text safe without changing link, image, or autolink destinations. Thanks for tightening the regression tests around that.

PR #1365, opened 2026-06-03
ci: pin types-cachetools in the mypy pre-commit hook

endofcake:
## Description

The mypy pre-commit hook lists types-cachetools unpinned in its additional_dependencies. pre-commit builds the hook's isolated env from the latest matching release, and a newer types-cachetools stub types TTLCache with three generic parameters. The existing two-arg annotation in src/mcp_atlassian/servers/main.py

```
token_validation_cache: TTLCache[
    int, tuple[bool, str | None, JiraFetcher | None, ConfluenceFetcher | None]
] = TTLCache(maxsize=100, ttl=300)
```

then fails type-checking with error: "TTLCache" expects 3 type arguments, but 2 given [type-arg]. Because the dependency is unpinned, this breaks the mypy hook (and CI) on main for anyone with a freshly built hook environment, independent of their own changes.

This pins types-cachetools to 6.2.0.20251022 — the version uv.lock already resolves, whose stub keeps TTLCache two-arg — so the pre-commit gate is reproducible and matches the project's own mypy environment (uv run mypy). No source code changes; the existing annotation is already correct for the pinned stub.

Fixes: N/A — no existing issue; pre-commit/CI reproducibility fix (happy to open one if you'd prefer it tracked).

## Changes

- Pin types-cachetools==6.2.0.20251022 in .pre-commit-config.yaml (the mypy hook's additional_dependencies), with an inline comment noting newer stubs make TTLCache 3-arg.
- One-line change; no production code touched.

## Testing

Reproduced: on a fresh hook env, pre-commit run mypy --all-files fails with servers/main.py: "TTLCache" expects 3 type arguments, but 2 given.
After the pin: pre-commit run mypy --all-files → Passed.
No code changed, so uv run pytest is unaffected.

- [x] Manual checks performed: `pre-commit run mypy --all-files fails before / passes after the pin`

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,7 @@ repos:
             'types-requests',
             'types-setuptools',
             'types-urllib3',
-            'types-cachetools',
+            'types-cachetools==6.2.0.20251022',  # pin: newer stubs make TTLCache 3-arg
             'types-PyYAML',
             'atlassian-python-api',
             'beautifulsoup4',


endofcake:
Waiting for a review.

PR #1366, opened 2026-06-03
feat(confluence): support base64 content in upload_attachment

Arbuzov:
## Summary

`confluence_upload_attachment` currently requires a server-readable `file_path`. This does not work when the MCP server runs **remotely or in a container** and has no access to the calling client's filesystem — the path the client knows simply does not exist on the server.

This PR adds an alternative, **filesystem-free** upload path while keeping the existing `file_path` behaviour fully intact.

## Changes

- **`AttachmentsMixin.upload_attachment_from_content(content_id, filename, content, comment, minor_edit)`** — uploads raw `bytes` directly. The multipart REST request is extracted into a shared `_send_attachment_request` helper reused by both the file-path and in-memory paths, so the existing upload behaviour is unchanged (`requests` accepts raw bytes as multipart content, so no temp file is needed).
- **`confluence_upload_attachment` tool** gains two parameters:
  - `content_base64` — base64-encoded file content
  - `filename` — required with `content_base64` (determines title/type)

  `file_path` is now optional. Callers must provide **exactly one** of `file_path` / `content_base64`. Invalid base64 and a missing `filename` are rejected with clear `ValueError`s.
- New abstract method added to `AttachmentsOperationsProto`.

## Backwards compatibility

Fully backwards compatible — existing `file_path` calls are unchanged. The only signature change is `file_path` becoming optional.

## Tests

- Mixin: success / no-content-id / no-filename / API-error for the in-memory path.
- Tool: base64 upload, mutual-exclusion (neither/both source), missing-filename, invalid-base64.

All new and existing attachment/server tests pass; `ruff`, `ruff-format` and `mypy` are clean on the changed files under the project config.

### DIFF
--- a/src/mcp_atlassian/confluence/attachments.py
+++ b/src/mcp_atlassian/confluence/attachments.py
@@ -167,6 +167,84 @@ class AttachmentsMixin(ConfluenceClient, AttachmentsOperationsProto):
             "failed": failed,
         }
 
+    def upload_attachment_from_content(
+        self,
+        content_id: str,
+        filename: str,
+        content: bytes,
+        comment: str | None = None,
+        minor_edit: bool = True,
+    ) -> dict[str, Any]:
+        """
+        Upload a single attachment from in-memory bytes.
+
+        This is the filesystem-free counterpart to ``upload_attachment``. It is
+        intended for deployments where the server cannot read host file paths
+        (for example, a remote or containerized MCP server): the caller provides
+        the raw file content directly instead of a path.
+
+        Args:
+            content_id: The Confluence content ID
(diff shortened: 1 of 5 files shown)


pkc-coder:
Please we need it. With this the MCP does not need to be local.

pkc-coder:
👍

Arbuzov:
Is this project alive?

sooperset:
This is a useful improvement: the in-memory upload path solves the remote-server case while keeping the existing file-path behavior intact. I also appreciate the duplicate-attachment versioning coverage, since that's where Cloud and DC behavior can drift.

PR #1227, opened 2026-04-01
fix(jira): allow return 'version' field

atline:
Fix `version` field also known as `Affects Version/s` not returned issue

## Description

`Affects Version/s:` should be standard field in Jira:

```
id: "versions"
name: "Affects Version/s"
custom: false
schema.system: "versions"
```

image

## Changes

- a versions list field to JiraIssue
- parse fields["versions"] in from_api_response
- emit it in to_simplified_dict when requested (keyed as versions, consistent with current API style)

## Testing

- [✓] Unit tests added/updated
- [✓] Manual checks performed:

image

image

## Checklist

- [✓] Code follows project style guidelines (linting passes).
- [✓] Tests added/updated for changes.
- [✓] All tests pass locally.

PR #1370, opened 2026-06-04
fix(jira): prefer exact Epic issue type matches

Niicolaa:
## Description

Jira issue creation could select a custom issue type such as `Program Epic` when it appeared before the standard `Epic` type in project metadata. Prefer the exact case-insensitive `Epic` match before falling back to localized and substring matching.

The original Affects Version/s model support is already present on `main` through #1227, so this branch now relies on that implementation instead of duplicating it.

## Changes

- Prefer an exact `Epic` issue type name when multiple Epic-like types exist.
- Preserve the existing localized and substring fallback behavior.
- Add a regression test for `Program Epic` appearing before `Epic`.
- Sync the branch with current `main`.

## Testing

- `uv run pytest tests/unit/ -x` (3,204 passed, 5 skipped)
- Jira Cloud E2E modules (28 passed, 8 OAuth-only skips)
- Jira Data Center E2E modules (51 passed, 1 environment skip)
- Real Jira lifecycle integration test

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,11 +1,5 @@
 {
   "permissions": {
-    "allow": [
-      "Edit:*",
-      "Write:*",
-      "MultiEdit:*",
-      "Bash:*"
-    ],
-    "defaultMode": "acceptEdits"
+    "defaultMode": "auto"
   }
 }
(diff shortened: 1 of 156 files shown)


Niicolaa:
It's still relevant please merge, thx :)

sooperset:
The regression test covering `Program Epic` before `Epic` is a useful guard against creating the wrong issue type, @Niicolaa. Thanks for keeping the localized fallback behavior intact as well.

PR #1379, opened 2026-06-09
fix(jira): parse full markdown table blocks and strip cell whitespace (#1343)

### LINKED ISSUE #1343: [Bug]: Markdown pipe tables get mangled by markdown→Jira-wiki conversion
## Description

Standard Markdown pipe tables passed to `jira_add_comment`, `jira_create_issue`, or `jira_update_issue` (description field) get mangled in the conversion to Jira wiki markup. The `|---|---|` separator row is misinterpreted; cell content drops; the resulting wiki table either fails to render as a table at all or renders with broken structure.

## Reproduction

Call `jira_add_comment` with a body containing a pipe table:

```markdown
| Region | Count | Roles |
|--------|-------|-------|
| east   | 10    | web=4, db=2 |
| west   | 8     | web=4, cache=2 |
```

## Expected

A Jira wiki table with one header row and two data rows. Either:

- Convert to Jira wiki table syntax: `||Region||Count||Roles||\n|east|10|web=4, db=2|\n|west|8|web=4, cache=2|`
- Or to ADF table nodes if the comment is going via the ADF path (issue #864)

## Actual

The Jira UI renders the table content as a mash of pipe-separated text on a single line, with no table structure. The `|---|---|` separator row appears as literal `---` segments. Bullet-list workarounds work fine, so this is specifically a table-conversion bug.

## Workaround

Replace pipe tables with bullet lists using bold labels:

```markdown
- **east** — 10 hosts: web=4, db=2
- **west** — 8 hosts: web=4, cache=2
```

This renders cleanly in the Jira UI.

## Environment

- mcp-atlassian: v0.21.1 via `uvx mcp-atlassian`
- Target: Jira Cloud (`*.atlassian.net`)

## Related

- #1340 — same converter family has problems with underscores inside inline code

## Willingness to contribute

Happy to test a fix or supply a tighter repro if needed.

simongonzalezdc:
## Problem

The `markdown_to_jira()` table converter only replaced `|` with `||` in the header row and deleted the separator, leaving data rows untouched with spaces around content (e.g. `| east |` instead of `|east|`). This caused Jira to render table cells with unwanted padding or as plain text.

Issue #1343.

### `src/mcp_atlassian/preprocessing/jira.py`

- Rewrote the markdown→Jira table loop to parse **full table blocks** (header + separator + 1+ data rows)
- Strip whitespace from each cell in both header and data rows
- Convert header delimiters to `||cell||` (Jira header syntax)
- Keep data row delimiters as `|cell|` (Jira data syntax)
- Drop the markdown separator line
- Support alignment separators (`:---:`, `------:`, etc.)

## Tests

- `tests/unit/preprocessing/test_markdown_table_to_jira.py`: 6 focused tests covering:
  1. Tables with padded cells (the original bug)
  2. Tables without padding
  3. Tables with alignment separators
  4. Single-column tables
  5. Non-table pipes left untouched
  6. Inline formatting preserved inside cells

## Verification

```bash
uv run pytest tests/unit/preprocessing/test_markdown_table_to_jira.py -v
# 6 passed

uv run pytest tests/unit/preprocessing/ -v
# 119 passed
```

Fixes #1343

### DIFF
--- a/src/mcp_atlassian/preprocessing/jira.py
+++ b/src/mcp_atlassian/preprocessing/jira.py
@@ -269,9 +269,11 @@ class JiraPreprocessor(BasePreprocessor):
         # Text formatting (bold, italic)
         output = re.sub(
             r"([*_])(.*?)\1",
-            lambda match: ("**" if match.group(1) == "*" else "*")
-            + match.group(2)
-            + ("**" if match.group(1) == "*" else "*"),
+            lambda match: (
+                ("**" if match.group(1) == "*" else "*")
+                + match.group(2)
+                + ("**" if match.group(1) == "*" else "*")
+            ),
             output,
         )
 
@@ -464,9 +466,7 @@ class JiraPreprocessor(BasePreprocessor):
         # Headers with = or - underlines
         output = re.sub(
             r"^(.*?)\n([=-])+$",
-            lambda match: (
-                f"h{1 if match.group(2)[0] == '=' else 2}. {match.group(1)}"
-            ),
+            lambda match: f"h{1 if match.group(2)[0] == '=' else 2}. {match.group(1)}",
             output,
             flags=re.MULTILINE,
         )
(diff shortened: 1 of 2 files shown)


sooperset:
Simon, the focused table tests are the valuable part here: they pin the padded-cell regression and the alignment separator case clearly. I merged this because the converter now handles the full table block instead of only rewriting the header.

PR #1380, opened 2026-06-10
fix(jira): fail loudly when assignee cannot be resolved in update_issue (silent no-op with modern account IDs)

kuznetsov-ai:
## Description

`jira_update_issue` with a valid modern Atlassian Cloud account ID (`712020:<uuid>`) reports **"Issue updated successfully" while the assignee never changes**. Two stacked bugs:

1. `_get_account_id()` only recognizes legacy account IDs (`assignee.startswith("5")`). Modern Cloud IDs like `712020:f653aab5-...` fail the heuristic, fall through to the user-search lookups (which query for the literal accountId string), find nothing, and raise `ValueError`.
2. `update_issue()` swallows that `ValueError` with `logger.warning`, so the PUT is skipped entirely and the tool returns the unchanged issue as a successful update.

The MCP client (LLM) has no way to detect this: the response is a success with the refetched — unchanged — issue. Observed in production against Jira Cloud (`712020:`-style accountIds are what Cloud issues `assignee.accountId` actually return today).

Overlaps with #1164 on the recognition half (kudos — same root cause); this PR additionally fixes the silent-failure half, accepts the documented `accountid:` prefix, and adds unit tests for all paths.

## Changes

- `jira/users.py::_get_account_id`: recognize the modern `<digits>:<uuid>` account ID format and strip the documented `accountid:` prefix (the format the `create_issue`/`update_issue` tool schemas advertise). Legacy `5...` IDs and email/username lookup behavior unchanged.
- `jira/issues.py::update_issue`: re-raise assignee resolution failures instead of swallowing them — an explicit assignee update that cannot be resolved now fails loudly instead of silently no-oping.
- Unit tests for both account ID formats, the prefix strip, and the loud failure (asserting the PUT is never issued).

## Testing

- [x] Manual checks performed: `verified against a real Jira Cloud instance — update_issue with a 712020:<uuid> assignee now issues the PUT and the assignee actually changes; an unresolvable user now raises instead of returning success. Full unit suite: 2581 passed, 5 skipped.`

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/issues.py
+++ b/src/mcp_atlassian/jira/issues.py
@@ -1106,7 +1106,13 @@ class IssuesMixin(
                             account_id = self._get_account_id(value)
                             self._add_assignee_to_fields(update_fields, account_id)
                         except ValueError as e:
-                            logger.warning(f"Could not update assignee: {str(e)}")
+                            # An explicit assignee update that cannot be resolved
+                            # must fail loudly. Swallowing it here means the PUT is
+                            # skipped (or runs without the assignee) while the call
+                            # still reports the issue as updated successfully.
+                            raise ValueError(
+                                f"Could not update assignee: {str(e)}"
+                            ) from e
                 elif key == "parent":
                     if isinstance(value, dict) and value.get("key"):
                         update_fields["parent"] = {"key": str(value["key"])}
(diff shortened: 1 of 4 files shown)


sooperset:
Merged, thanks Evgenii. The tests pin both the `712020:<uuid>` account ID path and the failure-before-PUT behavior, which is exactly the coverage this bug needed.

PR #1383, opened 2026-06-12
fix(jira): support project version creation on self-managed Jira (Dat…

justivome:
## Description

Self-managed Jira (Server/Data Center) only supports the v2 REST API. The `create_version` method was hardcoded to call `/rest/api/3/version`, which is exclusive to Jira Cloud and causes failures on self-hosted instances.

**References:**
- Jira Cloud API v3 – version creation: https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-project-versions/#api-rest-api-3-version-post
- Jira Server/DC API v2 – version creation: https://developer.atlassian.com/server/jira/platform/rest/v10000/api-group-version/#api-rest-api-2-version-post

## Changes

- Check `is_cloud` config flag in `create_version` to select the correct API version
- Jira Cloud uses `/rest/api/3/version` with `json=` payload
- Jira Server/Data Center uses `/rest/api/2/version` with `data=` payload

## Testing

- [x] Manual checks performed: verified version creation on self-managed Jira instance using API v2 endpoint

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/src/mcp_atlassian/jira/client.py
+++ b/src/mcp_atlassian/jira/client.py
@@ -382,7 +382,12 @@ class JiraClient:
         if description:
             payload["description"] = description
         logger.info(f"Creating Jira version: {payload}")
-        result = self.jira.post("/rest/api/3/version", json=payload)
+
+	if self.config.is_cloud:
+            result = self.jira.post("/rest/api/3/version", json=payload)
+        else:
+            result = self.jira.post("/rest/api/2/version", data=payload)
+
         if not isinstance(result, dict):
             error_message = f"Unexpected response from Jira API: {result}"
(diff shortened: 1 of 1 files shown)


sooperset:
Your diagnosis is correct: Jira Server/DC needs REST v2, and your patch selected the right endpoint for self-managed instances.

For this PR to merge, it would need:

1. A remaining version-creation bug on current main that the existing v2 implementation doesn't handle.
2. A regression test that fails on current main and passes with this PR.

Current main already uses v2 on both Cloud and DC, and its unit, Cloud, and DC tests all passed in my rerun. I recommend closing this as superseded by #1339 and #1350; no contributor action is needed.

PR #1386, opened 2026-06-16
fix(confluence): add /wiki prefix to Cloud attachment upload and delete URLs

### LINKED ISSUE #1321: [Bug]: confluence_upload_attachment fails on Confluence Cloud — direct API call constructs URL without /wiki/ prefix
## Description

`confluence_upload_attachment` (and `confluence_upload_attachments`) fail on **Confluence Cloud** with a generic "Failed to upload attachment" error. The page is created successfully, the file is found on disk, the auth is valid — but the upload returns 404. The cause is in `_upload_attachment_direct`: it hand-constructs the REST URL as `f"{base_url}/rest/api/content/.../child/attachment"`, which is the Server/DC path. Cloud serves Confluence at `/wiki/rest/api/...`, so the request hits a 404.

Other Confluence tools in the same server (`confluence_create_page`, `confluence_update_page`, etc.) work correctly against Cloud because they go through `atlassian-python-api`'s higher-level `Confluence` client, which knows about Cloud routing and prepends `/wiki/`. Only `_upload_attachment_direct` bypasses that layer and constructs the URL itself, so only attachment upload exhibits the bug.

## Reproduction

1. Configure an `mcp-atlassian` server against a Confluence Cloud site:
   - `CONFLUENCE_URL=https://<site>.atlassian.net`
   - `CONFLUENCE_USERNAME=<email>`, `CONFLUENCE_API_TOKEN=<token>`
2. Call `confluence_create_page(space_key=..., title=..., content=..., content_format="storage")` → succeeds, returns a page id.
3. Call `confluence_upload_attachment(content_id=<that page id>, file_path=<any local file>)` → returns `{"success": false, "error": "Failed to upload attachment <name> to content <id>"}`.

The server log shows the underlying 404:

```
ERROR - mcp-confluence - Direct API upload failed: 404 Client Error: Not Found for url: https://<site>.atlassian.net/rest/api/content/<id>/child/attachment
ERROR - mcp-confluence - Failed to upload attachment <name> to content <id>
```

## Root cause

In [`src/mcp_atlassian/confluence/attachments.py`](https://github.com/sooperset/mcp-atlassian/blob/main/src/mcp_atlassian/confluence/attachments.py), inside `_upload_attachment_direct`:

```python
# line 377
base_url = self.config.url.rstrip("/")
...
# line 380
url = f"{base_url}/rest/api/content/{content_id}/child/attachment"
```

For Cloud sites, `self.config.url` is the bare site URL (e.g. `https://example.atlassian.net`); the `/wiki/` prefix is normally added by `atlassian-python-api`'s `Confluence` client, not stored in the config. So the URL constructed here is missing the prefix on every Cloud deployment.

## Empirical confirmation

To isolate the cause from any other factors (auth, file content, content type, tool wrapping), I issued the identical multipart POST request to two URLs differing only in the `/wiki/` prefix, against the same Cloud site, with the same credentials, against a freshly-created page:

| URL | Result |
|---|---|
| `POST https://<site>.atlassian.net/rest/api/content/<id>/child/attachment` | **HTTP 404** — `{"detail":"No endpoint POST /rest/api/content/<id>/child/attachment"}` |
| `POST https://<site>.atlassian.net/wiki/rest/api/content/<id>/child/attachment` | **HTTP 200** — attachment created |

The URL prefix is the only difference. Adding `/wiki/` resolves the 404.

## Suggested fix

Two options, in increasing robustness:

1. **Detect Cloud and prepend `/wiki`:**

   ```python
   base_url = self.config.url.rstrip("/")
   if self.config.is_cloud:        # whatever the existing predicate is named
       base_url = f"{base_url}/wiki"
   url = f"{base_url}/rest/api/content/{content_id}/child/attachment"
   ```

2. **Route through `atlassian-python-api`'s built-in attachment helper** (`Confluence.attach_file()` or equivalent) so the Cloud-vs-Server prefixing logic isn't duplicated in `attachments.py`. This is the more future-proof fix — Cloud detection only has to live in one place, and any subsequent Atlassian URL changes are absorbed by the upstream library.

Note: switching from raw `self.confluence._session.post()` (as in #1306) to the higher-level helper is what gets the prefix automatically. Calling the session directly with a hand-constructed URL — which several recent PRs in this area do — does not pick up the prefix, even though it might look like it should.

## Related work

A few in-flight PRs touch the same function but don't address this bug:

- **#1253** — Adds POST/header/versioning fixes for Server/DC. URL construction stays at `{base_url}/rest/api/content/...` (no `/wiki/`). Cloud still 404s.
- **#1306** — Switches to `self.confluence._session.post(url, ...)`. The session object's `post()` takes whatever URL you pass it; the `/wiki/` prefix is applied by higher-level `Confluence` methods, not by the session. So this PR also leaves Cloud broken.

The fix above is independent of (and compatible with) both.

## Environment

- Image: `ghcr.io/sooperset/mcp-atlassian:latest`
- Image digest: `sha256:2a4bf26854fe05a03e6ba6033aeed4f6c184568f4749c24a895d6ac069ae47b1`
- `atlassian-python-api`: `4.0.7`
- Confluence: Cloud
- Auth: API token (Basic auth)

dsitze:
## Summary

Fixes the HTTP 404 reported in #1321 when uploading or deleting Confluence **Cloud** attachments via `confluence_upload_attachment` / `confluence_upload_attachments` (and the v1/token-auth delete path).

## Root cause

`_upload_attachment_direct` and the v1 branch of `delete_attachment` hand-build their REST URLs from `self.config.url`. On Cloud, `config.url` is the bare site URL (`https://site.atlassian.net`), but Cloud serves the REST API under `/wiki` (`https://site.atlassian.net/wiki/rest/api/...`). The hand-built URLs omit `/wiki`, so the requests 404. Every other Confluence tool works because it goes through `atlassian-python-api`'s higher-level client, which adds `/wiki`; only these two hand-built URLs miss it. (The issue confirms this empirically: identical POST differing only by `/wiki/` → 404 vs 200.)

## Fix

- New helper `_rest_base_url()` that prepends `/wiki` when `config.is_cloud`, guarded by an `endswith("/wiki")` check so configs that already include the prefix don't produce `/wiki/wiki`.
- Used at both hand-built call sites:
  - `_upload_attachment_direct`
  - the v1/token-auth branch of `delete_attachment`
- The OAuth delete path uses the v2 adapter and already constructs correct URLs, so it is left unchanged.

### Why not route through `attach_file()`?

`_upload_attachment_direct` deliberately bypasses `atlassian-python-api`'s `attach_file()` to support the `minorEdit` parameter (see the method's docstring). Routing through that helper would regress `minorEdit`, so the prefix is applied to the direct call instead.

## Tests

Added coverage in `tests/unit/confluence/test_attachments.py` for both upload and delete, asserting on the URL passed to the mocked session:

- **Cloud, bare URL** (`https://test.atlassian.net`) → URL contains the `/wiki` prefix
- **Cloud, URL already ending in `/wiki`** → no double `/wiki/wiki`
- **Server/DC** (`https://confluence.example.com`) → no `/wiki` prefix (unchanged behavior)

`uv run pytest` passes (full suite green) and `ruff-format` / `ruff` are clean.

Fixes #1321

### DIFF
--- a/src/mcp_atlassian/confluence/attachments.py
+++ b/src/mcp_atlassian/confluence/attachments.py
@@ -32,6 +32,23 @@ class AttachmentsMixin(ConfluenceClient, AttachmentsOperationsProto):
             )
         return None
 
+    def _rest_base_url(self) -> str:
+        """Return the REST API base URL, adding the Cloud ``/wiki`` prefix.
+
+        On Confluence Cloud, ``config.url`` is the bare site URL
+        (``https://site.atlassian.net``) but the REST API is served under
+        ``/wiki``. Hand-built URLs must include this prefix or requests 404.
+        The ``endswith`` guard avoids producing ``/wiki/wiki`` when the
+        configured URL already includes the prefix.
+
+        Returns:
+            The base URL to use for direct REST API calls.
+        """
+        base_url = self.config.url.rstrip("/")
+        if self.config.is_cloud and not base_url.endswith("/wiki"):
+            base_url = f"{base_url}/wiki"
+        return base_url
+
     def upload_attachment(
         self,
         content_id: str,
@@ -467,7 +484,7 @@ class AttachmentsMixin(ConfluenceClient, AttachmentsOperationsProto):
         """
         try:
             # Build the API endpoint URL
-            base_url = self.config.url.rstrip("/")
+            base_url = self._rest_base_url()
             url = f"{base_url}/rest/api/content/{content_id}/child/attachment"
 
             # Prepare headers (X-Atlassian-Token required for file uploads)
@@ -540,7 +557,7 @@ class AttachmentsMixin(ConfluenceClient, AttachmentsOperationsProto):
                     f"Using v1 API for token/basic authentication to delete attachment '{attachment_id}'"
                 )
                 # Use v1 API endpoint for deletion
-                base_url = self.config.url.rstrip("/")
+                base_url = self._rest_base_url()
                 url = f"{base_url}/rest/api/content/{attachment_id}"
                 response = self.confluence._session.delete(url)
                 response.raise_for_status()
--- a/tests/unit/confluence/test_attachments.py
+++ b/tests/unit/confluence/test_attachments.py
@@ -166,6 +166,68 @@ class TestAttachmentsMixin:
             assert call_args[1]["data"]["minorEdit"] == "false"
             # Note: comment is now in files dict as multipart form data, not in data dict
 
+    def _capture_upload_url(self, attachments_mixin, config_url: str) -> str:
+        """Run an upload with the given config URL and return the request URL.
+
+        Sets ``config.url`` (which drives ``is_cloud``), performs an upload with
+        all file I/O mocked, and returns the URL passed to ``session.put``.
+        """
+        attachments_mixin.config.url = config_url
+        self._mock_rest_api_upload(attachments_mixin)
+
+        with (
+            patch("os.path.exists", return_value=True),
+            patch("os.path.getsize", return_value=100),
+            patch("os.path.isabs", return_value=True),
+            patch("os.path.abspath", return_value="/absolute/path/test_file.txt"),
+            patch("os.path.basename", return_value="test_file.txt"),
+            patch("builtins.open", mock_open(read_data=b"test content")),
+        ):
+            attachments_mixin.upload_attachment(
+                "123456", "/absolute/path/test_file.txt"
+            )
+
+        return attachments_mixin.confluence._session.put.call_args[0][0]
+
+    def test_upload_attachment_cloud_adds_wiki_prefix(
+        self, attachments_mixin: AttachmentsMixin
+    ):
+        """Cloud bare site URL gets the /wiki prefix (otherwise the call 404s)."""
+        url = self._capture_upload_url(attachments_mixin, "https://test.atlassian.net")
+
+        assert (
+            url == "https://test.atlassian.net/wiki"
+            "/rest/api/content/123456/child/attachment"
+        )
+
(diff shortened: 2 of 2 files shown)


sooperset:
Maintainer recheck (merge-bucket triage):

**Scope/correctness** — The fix adds `/wiki` for Confluence Cloud only, via `_rest_base_url()` with an `endswith("/wiki")` guard, at the two hand-built REST URLs (`_upload_attachment_direct` and the v1/token delete branch). Behavior matrix, all covered by the new unit tests:
- Cloud, bare site URL → `/wiki` added (the #1321 fix)
- Cloud, URL already ending `/wiki` (the documented `CONFLUENCE_URL` form) → unchanged, no `/wiki/wiki`
- Server/DC → unchanged

Since the upload path is shared across Cloud and DC, I specifically confirmed DC is untouched (guarded by `is_cloud`) and the already-`/wiki` Cloud config is a no-op — so no regression to existing working configs.

**Verification** — CI is green across Python 3.10–3.13 (unit + integration), including the added URL-assertion tests. Note the Cloud e2e harness is configured with a `/wiki`-suffixed `CLOUD_E2E_CONFLUENCE_URL` (see `test_wiki_prefix_in_url`), so the guard makes this change a no-op there; the bare-URL bug path is exercised by the new unit tests rather than e2e.

**Status** — Mergeable, CI green, stale label removed. Branch is behind `main` (only non-conflicting test/docs commits; the changed file is untouched on main) — a maintainer can click "Update branch" before merging. Ready for human review.

Out of scope, non-blocking: OAuth Cloud uploads still route through `_upload_attachment_direct` against `config.url` rather than the `api.atlassian.com` session base — a pre-existing condition unaffected by this PR.

PR #1224, opened 2026-03-31
fix(jira): preserve embedded media on cloud description updates

sean-stallbaum-fortra:
## Summary

Fixes #1192.

When `jira_update_issue` updates a Jira Cloud description from Markdown/text, it was replacing the entire ADF document and dropping any existing UI-inserted media nodes. On Cloud, that can remove the embedded attachment itself.

This change preserves existing top-level ADF media blocks from the current Jira Cloud description before sending the v3 update payload.

## What changed

- Added a Cloud-only preservation step in `IssuesMixin.update_issue()` for Markdown/text description rewrites
- Fetched the current raw description from Jira Cloud `rest/api/3` before issuing the update
- Merged existing top-level `mediaSingle` / `mediaGroup` / `media` blocks into the outgoing ADF document
- Kept explicit caller-supplied ADF updates as replace semantics
- Added unit regression coverage for both the ADF merge helper and the Cloud update path

## Tests

Ran:

```bash
uv run pytest tests/unit/models/test_adf.py tests/unit/jira/test_issues.py -xvs
```

Result:
- `126 passed`

## Manual Jira Cloud validation

Validated against a real Jira Cloud issue using the local dev server build.

### Repro before fix

1. Create a disposable Jira Cloud issue.
2. Upload a small `.webm` attachment to the issue.
3. In the Jira web UI, embed that uploaded video inline in the description.
4. Confirm the raw v3 issue payload at `rest/api/3/issue/{issueKey}?fields=description,attachment` contains:
   - two attachments
   - a top-level `mediaSingle` node in `fields.description`
5. Call `jira_update_issue` with a Markdown/text description update.
6. Re-fetch the raw v3 issue payload.
7. Observe the bug:
   - the embedded attachment is removed
   - the `mediaSingle` node disappears from `fields.description`

### Verification after fix

1. Create or reset a disposable Jira Cloud issue.
2. Upload a small `.webm` attachment.
3. Embed it inline in the Jira description via the web UI.
4. Confirm the raw v3 payload at `rest/api/3/issue/{issueKey}?fields=description,attachment` shows:
   - both attachments present
   - a top-level `mediaSingle` node in `fields.description`
5. Run `jira_update_issue` with a Markdown/text description update.
6. Re-fetch the same raw v3 payload.
7. Confirm:
   - the updated paragraph text is present
   - the embedded attachment still exists
   - the top-level `mediaSingle` node is still present

### Observed result

After the fix, the raw Jira Cloud payload still contained:
- both attachments
[...]

PR #1402, opened 2026-06-23
feat(jira): attachment upload/delete, comment delete, and inline-media comments

ChekeEdd:
## Description

Adds four Jira tools that aren't in the server today, plus the supporting client methods, ADF model work, tests, and docs. The headline feature is **inline-media comments** on Jira Cloud — a comment whose body interleaves rich text and embedded screenshots (text → image → text → image), which a plain-text comment cannot express.

New MCP tools:
- `jira_upload_attachment` — upload from a local path or inline base64
- `jira_delete_attachment`
- `jira_delete_comment`
- `jira_add_comment_with_media` — **Jira Cloud**: screenshots embedded inline in the comment body, in order

How inline media works: each image is uploaded as a normal issue attachment, then its **Media Services file UUID** (required by ADF `media` nodes, and distinct from the numeric REST attachment id) is resolved from the redirect of `GET /rest/api/3/attachment/content/{id}` (`.../file/{uuid}/binary`). The comment is then posted as an ADF document via the v3 API, interleaving `paragraph` and `mediaSingle`/`media` nodes.

The media-id resolution is hardened for real-world setups:
- builds an **absolute** request URL (some `atlassian-python-api` versions return a relative `resource_url`, which made `requests` raise `MissingSchema`);
- falls back to following the full redirect chain (OAuth gateway `api.atlassian.com/ex/jira/{cloudId}`, proxies, multi-hop), matching the `/file/{uuid}` fragment host-agnostically;
- **rolls back** every attachment uploaded during the call if any later step fails, so a failed comment leaves no orphan attachments;
- attaches pixel `width`/`height` to the media node, parsed locally from the image header (stdlib only — PNG/JPEG/GIF/BMP/WebP, no Pillow dependency).

Cloud-only behaviour is gated on `is_cloud` (inline ADF media is not available on Server/DC).

## Changes

- `jira/attachments.py`: `upload_attachment_content`, `upload_attachment_from_path`, `delete_attachment`, `get_attachment_media_id` (+ absolute-URL/redirect-chain handling), image-dimension surfacing.
- `jira/comments.py`: `delete_comment`, `add_comment_adf`.
- `jira/client.py`: `_delete_api3` helper (v3 DELETE).
- `models/jira/adf.py`: `build_media_comment_adf` (interleaved text + inline media, optional width/height).
- `utils/media.py`: `get_image_dimensions` (stdlib header parser).
- `servers/jira.py`: the four tools, with attachment rollback on failure.
- `utils/toolsets.py`, `docs/tools-reference.mdx`, `README.md`: registration + docs.

## Testing

- [x] Manual checks performed: verified `jira_add_comment_with_media` against a live Jira Cloud instance — `text → screenshot → text → screenshot` renders inline; failure path rolls back the uploaded attachments.

Full unit suite: 2645 passed, 5 skipped. `ruff` and `mypy` (project pre-commit config) clean.

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,16 @@
 
 Model Context Protocol (MCP) server for Atlassian products (Confluence and Jira). Supports both Cloud and Server/Data Center deployments.
 
+> [!NOTE]
+> **This is an internal fork of [sooperset/mcp-atlassian](https://github.com/sooperset/mcp-atlassian)** (MIT licensed — see [`LICENSE`](LICENSE)), maintained for our team.
+> It adds the following Jira tools on top of upstream:
+> - `jira_upload_attachment` — add an attachment to an issue (local path or inline base64)
+> - `jira_delete_attachment` — delete an attachment by id
+> - `jira_delete_comment` — delete a comment from an issue
+> - `jira_add_comment_with_media` — post a Jira **Cloud** comment with screenshots embedded inline (text → image → text → image)
+>
+> All original copyright and the MIT license are preserved. To sync with upstream: `git fetch upstream && git merge upstream/main`.
+
 https://github.com/user-attachments/assets/35303504-14c6-4ae4-913b-7c25ea511c3e
 
 <details>
--- a/docs/tools-reference.mdx
+++ b/docs/tools-reference.mdx
@@ -63,13 +63,13 @@ Toolsets group related tools for easier management via `TOOLSETS` env var or `--
 |---------|:----:|-------|
 | `jira_issues` | Yes | `jira_get_issue`, `jira_search`, `jira_get_project_issues`, `jira_create_issue`, `jira_update_issue`, `jira_delete_issue`, `jira_move_issue`, `jira_batch_create_issues`, `jira_batch_get_changelogs` |
 | `jira_fields` | Yes | `jira_search_fields`, `jira_get_field_options` |
-| `jira_comments` | Yes | `jira_add_comment`, `jira_edit_comment` |
+| `jira_comments` | Yes | `jira_add_comment`, `jira_edit_comment`, `jira_delete_comment`, `jira_add_comment_with_media` |
 | `jira_transitions` | Yes | `jira_get_transitions`, `jira_transition_issue` |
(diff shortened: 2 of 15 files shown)


ChekeEdd:
Rebased onto latest `main` (`d494634`) — conflicts resolved, now mergeable.

Resolutions:
- `src/mcp_atlassian/models/jira/adf.py`: kept both this PR's `build_media_comment_adf` and #1224's `merge_adf_with_preserved_media` / `extract_top_level_media_nodes` — they are disjoint functions that happened to land at the same spot in the file.
- `tests/unit/jira/test_comments.py`: kept both this PR's `delete_comment` / `add_comment_adf` tests and the new `TestInternalCommentPublicParam` class.
- `tests/unit/utils/test_toolsets.py`: Jira tool count 56 → 60 (this PR adds 4).
- `src/mcp_atlassian/utils/toolsets.py`: the module docstring said `Groups 81 tools` while `main` actually registers 86 (56 Jira + 30 Confluence) — it looks like the count is updated in the test assertions but not in the docstring. I set it to the correct post-merge value, `90`. Happy to drop that one-line fix if you'd rather keep it out of this PR.

Full unit suite passes (2982 passed, 5 skipped); ruff and mypy clean.

This PR is still relevant — could the `stale` label be removed? I don't have permission to remove it myself.

sooperset:
The inline-media comment flow is a useful addition, and the redirect-chain handling, rollback behavior, and ADF-ordering tests cover real Cloud failure modes well. The image-dimension support is also a thoughtful detail.

I can't merge this as written because `upload_attachment_from_path` explicitly allows arbitrary absolute paths. That reopens the arbitrary-file-read and exfiltration issue addressed by the current attachment security hardening, and the PR is now conflicting with `main`.

For merge, please:

1. Constrain every caller-supplied upload path to the server workspace with the same safe-path boundary used by the current attachment code. Remove or invert the tests and docs that expect `/tmp` or other outside-workspace paths to succeed.
2. Rebase onto current `main`, preserve the attachment security fixes, and rerun the full unit suite, lint, mypy, and both Cloud and DC e2e coverage for the shared Jira REST changes.
3. Rework the new standalone tools against #1104. Keep the media work aligned with the planned media consolidation, and explain how comment deletion fits the outcome-oriented surface or fold it into the closest existing outcome tool.

Once those are in, I can review the security-sensitive path and the roadmap fit again.

PR #1403, opened 2026-06-23
fix(jira): accept comment alias for add comment

balcsida:
## Summary
- accept `comment` as a hidden compatibility alias for `jira_add_comment`'s `body` parameter
- add a FastMCP server regression test for the reported `comment` input shape

## Root cause
`jira_add_comment` advertised and required `body`, but a caller sent `comment`. FastMCP/Pydantic validates tool arguments before Jira logic runs, so the request failed with missing `body` and unexpected `comment`.

## Validation
- `uv run pytest tests/unit/servers/test_jira_server.py -q` -> `87 passed`
- `uv run pre-commit run --files src/mcp_atlassian/servers/jira.py tests/unit/servers/test_jira_server.py` -> fails in unrelated baseline mypy check: `src/mcp_atlassian/servers/main.py:363: error: "TTLCache" expects 3 type arguments, but 2 given [type-arg]`

### DIFF
--- a/src/mcp_atlassian/servers/jira.py
+++ b/src/mcp_atlassian/servers/jira.py
@@ -7,7 +7,7 @@ from typing import Annotated, Any
 
 from fastmcp import Context, FastMCP
 from mcp.types import BlobResourceContents, EmbeddedResource, ImageContent, TextContent
-from pydantic import Field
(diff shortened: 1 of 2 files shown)


balcsida:
Still relevant

PR #1406, opened 2026-06-24
fix(jira): convert [~accountid] mentions to ADF mention nodes

ChekeEdd:
## Summary

Comments posted to Jira Cloud that contain a user mention in the `[~accountid:<id>]` wiki syntax were stored as a **plain ADF text node**, so the Jira UI rendered the literal string `[~accountid:712020:…]`, the user was not notified, and the mention was not clickable. No exception was raised — a silent semantic mis-conversion in the markdown→ADF path.

## Root cause

On Cloud, both `jira_add_comment` and `jira_edit_comment` convert the Markdown body via `_markdown_to_jira` → `markdown_to_adf` → `_parse_inline_formatting` (`src/mcp_atlassian/models/jira/adf.py`). That inline parser handled bold/italic/code/links/strike but had **no rule for mentions**, so the mention token fell through to a literal text node.

## Fix

`_parse_inline_formatting` now recognises two mention forms and emits ADF `mention` nodes, placed **before** the link pattern so `@[name](id)` is not mis-parsed as a link:

- `[~accountid:<id>]` → `{"type": "mention", "attrs": {"id": "<id>"}}`
- `@[Display Name](<id>)` → `{"type": "mention", "attrs": {"id": "<id>", "text": "@Display Name"}}`

Because both comment tools share the `markdown_to_adf` path, the fix applies to add and edit alike. The reverse path (`adf_to_text`) already handles `mention` nodes, so read-back stays consistent.

### Note on "Defect 1" (stray leading backslash)
This fork's Cloud path uses the in-repo `markdown_to_adf` parser, which does **not** markdown-escape a leading `[`. I verified a mention at line start is converted with no stray `\` — so the escaping defect from the report does not reproduce here; only the missing-mention-node defect did.

## Verification

Before:
```json
{"type": "text", "text": "[~accountid:712020:3420e1b4-xxxx]"}
```
After:
```json
{"type": "mention", "attrs": {"id": "712020:3420e1b4-xxxx"}}
```

Authoritative check is the stored ADF (`GET /rest/api/3/issue/{KEY}/comment/{ID}`), not the text read-back.

## Tests

Added to `tests/unit/models/test_adf.py`:
- `[~accountid:<id>]` → mention node (literal wiki syntax must not survive)
- mention at line start (asserts no stray `\`)
- `@[Display Name](<id>)` → mention node with `text` attr
- regression: plain `[text](url)` link is unaffected

All unit tests pass (`tests/unit/models`, `tests/unit/jira/test_comments.py`, `tests/unit/preprocessing`); ruff + format clean; no new mypy errors introduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,16 @@
 
 Model Context Protocol (MCP) server for Atlassian products (Confluence and Jira). Supports both Cloud and Server/Data Center deployments.
 
+> [!NOTE]
+> **This is an internal fork of [sooperset/mcp-atlassian](https://github.com/sooperset/mcp-atlassian)** (MIT licensed — see [`LICENSE`](LICENSE)), maintained for our team.
+> It adds the following Jira tools on top of upstream:
+> - `jira_upload_attachment` — add an attachment to an issue (local path or inline base64)
+> - `jira_delete_attachment` — delete an attachment by id
+> - `jira_delete_comment` — delete a comment from an issue
+> - `jira_add_comment_with_media` — post a Jira **Cloud** comment with screenshots embedded inline (text → image → text → image)
+>
+> All original copyright and the MIT license are preserved. To sync with upstream: `git fetch upstream && git merge upstream/main`.
+
 https://github.com/user-attachments/assets/35303504-14c6-4ae4-913b-7c25ea511c3e
 
 <details>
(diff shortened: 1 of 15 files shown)


ChekeEdd:
Closing as superseded.

The `[~accountid:<id>]` → ADF mention conversion this PR adds is already implemented on `main` by #1307 — same regex (`[^\]]+` id capture, accepting both legacy 24-hex and `712020:UUID` ids), same `{"type": "mention", "attrs": {"id": ...}}` node, and the same mention-before-link ordering in `_parse_inline_formatting`.

The only behaviour unique to this PR was the prefix-less `@[Display Name](<accountId>)` form. #1228 covers the same intent with `@[Display Name](accountid:<accountId>)`, and the mandatory `accountid:` prefix is the safer design: without it, an ordinary markdown link written as `@[label](https://example.com)` is parsed as a mention with `id="https://example.com"`. Both forms occupy the same regex alternation slot, so they can't coexist — and upstream's guarded variant is the right one to keep.

Note this PR also carried the three commits from #1402; that PR has now been rebased onto current `main` and stands on its own.

PR #1408, opened 2026-06-24
fix(jira): autolink issue keys in cloud ADF

balcsida:
## Summary

- Link bare Jira issue keys when converting Markdown to Jira Cloud ADF.
- Pass the configured Jira base URL into the Cloud ADF converter.
- Add regression coverage for converter output and the Jira Cloud helper path.

## Root cause

Jira Cloud descriptions and comments are sent as ADF. Bare `PROJ-123` text in ADF is just text, so Jira's wiki-style issue key autolinker does not run over it. The converter now emits an ADF `link` mark pointing at `/browse/KEY` when a Jira base URL is available.

## Validation

- `uv run pytest -q` -> `2580 passed, 161 skipped`
- `uv run pre-commit run --files src/mcp_atlassian/jira/client.py src/mcp_atlassian/models/jira/adf.py tests/unit/jira/test_comments.py tests/unit/models/test_adf.py` -> passed
- `uv run pre-commit run --all-files` -> existing unrelated mypy failure on `src/mcp_atlassian/servers/main.py:363` (`TTLCache` expects 3 type arguments, but 2 given)

### DIFF
--- a/src/mcp_atlassian/jira/client.py
+++ b/src/mcp_atlassian/jira/client.py
@@ -10,6 +10,12 @@ from requests.exceptions import ConnectionError as RequestsConnectionError
 
 from mcp_atlassian.exceptions import MCPAtlassianAuthenticationError
 from mcp_atlassian.preprocessing import JiraPreprocessor
+from mcp_atlassian.utils.http import (
+    configure_circuit_breaker,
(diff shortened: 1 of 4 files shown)


balcsida:
Not stale

PR #1417, opened 2026-06-27
fix(http): cache multi-user token validation across requests

### LINKED ISSUE #1405: Per-request token validation hammers `/rest/api/user/current` and can trigger 429 (declared `token_validation_cache` is never used)
## Summary

In multi-user HTTP transport mode, every incoming request re-runs upstream
credential validation against Atlassian (Confluence: `GET /rest/api/user/current`,
Jira: `get_current_user_account_id()`). There is no cross-request caching of these
validation results, so a busy client produces one validation call **per request, per
service**, in addition to the actual tool call.

On self-hosted instances (especially behind a WAF / rate limiter) this extra,
constant traffic to a single endpoint can get throttled, surfacing as:

```
Error calling tool 'get_page': Invalid user Confluence token or configuration:
Confluence token validation failed with HTTPError: 429 Client Error:
Too Many Requests for url: https://<host>/rest/api/user/current
```

Notably, the codebase **already declares a cache that appears intended for exactly
this purpose, but it is never referenced anywhere**:

```python
# src/mcp_atlassian/servers/main.py:363
token_validation_cache: TTLCache[
    int, tuple[bool, str | None, JiraFetcher | None, ConfluenceFetcher | None]
] = TTLCache(maxsize=100, ttl=300)
```

`rg token_validation_cache src tests` returns only this declaration — it is dead code.

## Environment

- Version / commit: `ba72540` (current `main` at time of writing)
- Transport: streamable-HTTP, **multi-user** (per-request auth: header PAT / basic / OAuth-PAT)
- Atlassian: self-hosted **Confluence Server/DC** behind a corporate gateway

## Root cause

Fetchers are cached on `request.state`, which is a fresh object per HTTP request, so
the cache only dedupes *within* a single request — never across requests.

`src/mcp_atlassian/servers/dependencies.py`:

```python
# _get_fetcher() — cache lookup is per-request only
cached = getattr(request.state, spec.state_key, None)   # ~line 522
if cached:
    return cached
```

```python
# _create_and_validate() — validation call runs for every newly-created fetcher
fetcher = spec.fetcher_class(config=config)             # ~line 229
...
validation_data = spec.validate_fn(fetcher)             # ~line 234  <-- network call
spec.on_validated(...)
setattr(request.state, spec.state_key, fetcher)         # ~line 242
```

`validate_fn` per service (`_ServiceSpec`, ~lines 135–163):

- Jira: `lambda f: f.get_current_user_account_id()`
- Confluence: `lambda f: f.get_current_user_info()` → `GET /rest/api/user/current`

Because each new request builds a fresh fetcher, `validate_fn` fires every time.

### Secondary issue: 429 is misreported as a generic failure

In `src/mcp_atlassian/confluence/users.py` (`get_current_user_info`), only `401`/`403`
are special-cased; `429` falls through to the generic `HTTPError` branch and is
surfaced as a vague *"token validation failed"*, hiding the fact that it is a
rate-limit/backoff signal rather than a bad credential:

```python
except HTTPError as http_err:
    if http_err.response is not None and http_err.response.status_code in [401, 403]:
        ...  # clear auth error
    # 429 ends up here, reported as a generic HTTPError / "Invalid token"
    raise MCPAtlassianAuthenticationError(
        f"Confluence token validation failed with HTTPError: {http_err}"
    ) from http_err
```

## Impact

- Up to **2x** the necessary calls to Atlassian for simple tool usage (validation + work).
- Constant, repetitive load on a single endpoint (`/rest/api/user/current`), which is
  exactly the kind of pattern WAFs / DoS protection throttle first.
- A transient `429` is reported to users as an invalid-token error, which is misleading.
- The global/single-user fallback path does not validate at all, so this only affects
  multi-user deployments — i.e. the deployments most likely to be under load.

## Suggested resolution (for maintainer to decide)

The minimal, low-risk fix is to **wire up the already-declared validation cache** so a
given credential is validated at most once per TTL window, instead of once per request.

1. Introduce a module-level cache keyed across requests (reuse/relocate the existing
   `token_validation_cache`, or add a dedicated one in `dependencies.py`):
   - Key: `(spec.name, sha256(credential))` where *credential* is the resolved
     token / `username:api_token` / OAuth access token. **Hash it — never store raw
     tokens in memory keys/logs.**
   - Value: the `validation_data` returned by `validate_fn` (so `on_validated` can
     still populate `request.state`).
2. In `_create_and_validate()`, before calling `spec.validate_fn`:
   - On **cache hit**: skip the network call, reuse cached `validation_data`, still run
     `on_validated(...)` and set `request.state`.
   - On **cache miss**: validate, then store `validation_data` in the cache.
3. Make it configurable, e.g.:
   - `MCP_ATLASSIAN_VALIDATION_CACHE_TTL` (default `300`, set `0` to disable)
   - `MCP_ATLASSIAN_VALIDATION_CACHE_MAXSIZE` (default `100`)

Effect: at most one `/rest/api/user/current` (and one Jira validation) per credential
per TTL, regardless of request volume.

**Tradeoff to document:** a credential revoked upstream stays "valid" locally for up to
the TTL. The TTL/disable knob lets operators choose their own balance; the existing
declaration already implies a 5-minute window is acceptable.

### Secondary improvement (independent, small)

In `confluence/users.py` (and the Jira equivalent), special-case `429` so it is reported
as a rate-limit/backoff condition rather than a generic "invalid token" error. This is a
diagnostics-only change — **no automatic retry/backoff is proposed here** (deliberately
left out to avoid amplifying load during an upstream 429 storm).

## Related issues

- #1323 — validation hits `/rest/api/user/current` (+ deprecated `/rest/api/space`)
  on multi-user setups (same code path).
- #884 — 4xx from self-hosted installs behind DoS protection; workaround was setting a
  custom `User-Agent` via `*_CUSTOM_HEADERS`. Same family of symptom, different lever.
- #850 / #1301 — multi-user per-request config (the path that re-validates each request).

syf2211:
## Description

Fixes #1405

In multi-user HTTP transport mode, every request re-validated upstream credentials against Atlassian (`/rest/api/user/current`, Jira `myself()`), even when the same credential was validated moments earlier. A `token_validation_cache` existed in `main.py` but was never wired up.

This PR adds a cross-request validation cache in `dependencies.py` and removes the dead declaration from `main.py`.

## Changes

- Cache successful `validate_fn` results keyed by SHA-256(service + url + credentials)
- Configurable via `MCP_ATLASSIAN_VALIDATION_CACHE_TTL` (default 300, `0` disables) and `MCP_ATLASSIAN_VALIDATION_CACHE_MAXSIZE` (default 100)
- Report HTTP 429 during token validation as rate-limit errors (Jira + Confluence) instead of generic invalid-token failures
- Document new env vars in `.env.example`

## Testing

- [x] Manual checks performed: targeted pytest for dependencies + user validation modules

```
uv run pytest tests/unit/servers/test_dependencies.py tests/unit/confluence/test_users.py tests/unit/jira/test_users.py -q
# 139 passed
```

## Checklist

- [x] All tests pass locally.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -138,6 +138,12 @@ CONFLUENCE_API_TOKEN=your_confluence_api_token_here
 
 # --- HTTP Timeout ---
 # Timeout in seconds for HTTP requests to Atlassian APIs. Default is 75 seconds.
+
+# --- Multi-user token validation cache ---
+# Cache successful credential validation across HTTP requests (multi-user mode).
+# Reduces repeated calls to /rest/api/user/current and Jira myself() during bursts.
+#MCP_ATLASSIAN_VALIDATION_CACHE_TTL=300
+#MCP_ATLASSIAN_VALIDATION_CACHE_MAXSIZE=100
 #JIRA_TIMEOUT=75
 #CONFLUENCE_TIMEOUT=75
 
--- a/src/mcp_atlassian/confluence/users.py
+++ b/src/mcp_atlassian/confluence/users.py
@@ -83,6 +83,15 @@ class UsersMixin(ConfluenceClient):
                 raise MCPAtlassianAuthenticationError(
                     f"Confluence token validation failed: {http_err.response.status_code} from /rest/api/user/current"
                 ) from http_err
+            if http_err.response is not None and http_err.response.status_code == 429:
+                logger.warning(
+                    "Confluence token validation rate-limited with HTTP 429 "
+                    "for /rest/api/user/current."
+                )
+                raise MCPAtlassianAuthenticationError(
+                    "Confluence token validation rate-limited (HTTP 429) by "
+                    "/rest/api/user/current. Retry later or reduce request volume."
+                ) from http_err
             logger.error(
                 f"HTTPError when calling Confluence /rest/api/user/current: {http_err}",
                 exc_info=True,
--- a/src/mcp_atlassian/jira/users.py
+++ b/src/mcp_atlassian/jira/users.py
@@ -8,6 +8,7 @@ import requests
 from requests.exceptions import HTTPError
 from unidecode import unidecode
 
+from mcp_atlassian.exceptions import MCPAtlassianAuthenticationError
 from mcp_atlassian.models.jira.common import JiraUser
 from mcp_atlassian.utils.decorators import handle_auth_errors
 
@@ -91,6 +92,15 @@ class UsersMixin(JiraClient):
             self._current_user_account_id = account_id
             return account_id
         except HTTPError as http_err:
+            if http_err.response is not None and http_err.response.status_code == 429:
+                logger.warning(
+                    "Jira token validation rate-limited with HTTP 429 while "
+                    "calling myself()."
+                )
+                raise MCPAtlassianAuthenticationError(
+                    "Jira token validation rate-limited (HTTP 429) while "
+                    "validating credentials. Retry later or reduce request volume."
+                ) from http_err
             response_content = ""
             if http_err.response is not None:
                 try:
--- a/src/mcp_atlassian/servers/dependencies.py
+++ b/src/mcp_atlassian/servers/dependencies.py
@@ -13,7 +13,6 @@ from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
 from cachetools import TTLCache
-
 from fastmcp import Context
 from fastmcp.server.dependencies import get_access_token, get_http_request
 from starlette.requests import Request
--- a/src/mcp_atlassian/servers/dependencies.py
+++ b/src/mcp_atlassian/servers/dependencies.py
@@ -6,9 +6,13 @@ Provides get_jira_fetcher and get_confluence_fetcher for use in tool functions.
 from __future__ import annotations
 
 import dataclasses
+import hashlib
 import logging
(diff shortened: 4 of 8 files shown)


sooperset:
The service, URL, tenant, and credential-digest keying is careful, and keeping fetchers request-local avoids cross-user session leakage.

1. Concurrent requests for the same key still duplicate upstream validation.

**`src/mcp_atlassian/servers/dependencies.py:445`** — **BLOCKER:** The cache lookup and insert are locked separately, leaving `validate_fn` outside the critical section. Two simultaneous requests for the same credential can both observe `_CACHE_MISS` and both call Atlassian, which recreates the 429 spike this change is meant to prevent. A focused 2-thread reproduction records 2 validation calls here, while #1423 kept lookup, validation, and insertion atomic. Required for merge: use per-key single-flight behavior or an equivalent mechanism, and add a concurrent regression test that asserts one validation call, separate request-local fetchers, and no cached value after failure.

sooperset:
Thanks @syf2211 — your multi-user validation cache work (dependency-layer wiring, configurable TTL/maxsize, and clearer Confluence 429 diagnostics) shipped in `main` via the consolidated #1481. Closing as landed; full credit to you for the design. 🙏

PR #1425, opened 2026-07-03
feat(jira): server-side internal-only comment guard for JSM projects

JeanBaptisteRenard:
## Motivation

Teams running JSM (Jira Service Management) ServiceDesk projects through
this MCP server want a way to guarantee that automation can *never* post
customer-visible content to certain projects — even if a client forgets to
pass `public=False`, or edits a comment that later needs to stay internal.
Today this is entirely the caller's responsibility: `jira_add_comment`
defaults to a public comment when `public` is omitted, and `jira_edit_comment`
has no way to know (or care) whether the comment it's about to overwrite is
currently public or internal.

## What this adds

A new, strictly opt-in env var `JIRA_INTERNAL_ONLY_PROJECTS`
(comma-separated project keys). When set:

- `jira_add_comment` rejects any call targeting a listed project unless
  `public` is exactly `False`.
- `jira_edit_comment` fetches the target comment's current visibility via
  the ServiceDesk API and rejects edits to comments that are currently
  public (fail-closed: unverifiable visibility refuses the edit).
- `jira_transition_issue`'s `comment` argument and
  `jira_create_issue_link`'s comment payload are rejected for listed
  projects — both post standard, potentially customer-visible Jira
  comments with no way to force them internal (the link guard checks both
  sides of the link, erring safe). The transition/link itself, without a
  comment, is unaffected.
- Issue keys are normalized (stripped) before matching, so
  whitespace-padded keys cannot bypass the guard.

Known scoping: `jira_add_worklog`'s comment is intentionally not guarded
(worklog entries are not portal-visible to JSM customers by default); see
the coverage map in the `comments.py` module docstring.

Unset (the default), this is a complete no-op — no behavior change for
existing deployments.

## Also included

A docstring fix on both comment tools: the `public` parameter's description
claimed the ServiceDesk API uses plain text. In practice the body is shipped
as a raw string that Jira Cloud renders server-side and stores as ADF
(observed to render markdown constructs on Cloud, though without the
client-side markdown→ADF guarantees of the regular comment path). Wording
updated to match.

## Testing

New unit tests cover: unlisted projects unaffected, listed+public→reject,
listed+internal→pass, edit of a public comment→reject, edit of an internal
comment→pass, fail-closed visibility lookup, whitespace-padded key bypass
class, transition-comment and link-comment guards (reject/pass matrices),
unset env→no-op, and malformed env input tolerated.

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -183,6 +183,18 @@ MCP_VERY_VERBOSE=true   # Enables DEBUG level logging (equivalent to 'mcp-atlass
 # Optional: Comma-separated list of Jira project keys to limit searches and other operations to.
 #JIRA_PROJECTS_FILTER=PROJ,DEVOPS
 
+# Optional: Comma-separated list of JSM/Service Desk project keys where
+# jira_add_comment/jira_edit_comment enforce internal-only (non-customer-visible)
+# comments server-side. add_comment rejects any call where public is not exactly
+# false; edit_comment rejects edits to comments that are currently public;
+# transition comments (jira_transition_issue) and issue-link comments
+# (jira_create_issue_link) are rejected when they touch a listed project.
+# Worklog comments are intentionally NOT guarded (not portal-visible on JSM).
+# Unset/empty = feature disabled, zero behavior change.
+# NOTE: listing a NON-JSM project key hard-blocks all comment automation on it
+# (only public=false is accepted, and the ServiceDesk API 404s on non-JSM issues).
+#JIRA_INTERNAL_ONLY_PROJECTS=CC
+
 # --- SLA Metrics Configuration ---
 # Configure how SLA metrics are calculated for Jira issues.
 # Default metrics to calculate (comma-separated).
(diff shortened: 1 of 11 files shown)


sooperset:
Thank you for this, @JeanBaptisteRenard — the internal-only comment guard shipped in 02de188 (via #1478), and the design is yours: the fail-closed visibility handling and the idea of an opt-in project allowlist both came from this PR.

This branch couldn't take the rebased history under our maintainer token, so the work was continued on #1478 rather than force-pushed here. What landed on top of your original:

- the MCP tool boundary no longer drops an explicit `public=false` (it was being coerced to `None`, so the guard read it as "omitted" and refused legitimate internal comments) — a bare `false` is now honored only for a configured project, since some MCP clients auto-fill omitted optional booleans;
- a failed ServiceDesk request is never retried as an ordinary Jira comment, which could otherwise have published it to the customer portal;
- the ServiceDesk path is now verified end-to-end against real JSM projects on **both** Cloud and Data Center, which is what this PR had been blocked on.

Closing as superseded — the credit stays with you. Thanks again for the careful security posture in the original.

PR #1434, opened 2026-07-06
fix(security): allowlist ATLASSIAN_OAUTH_CLIENT_STORAGE_FACTORY imports

failsafesecurity:
## Summary

When `ATLASSIAN_OAUTH_CLIENT_STORAGE_MODE=factory`, the server must not import an arbitrary Python module from `ATLASSIAN_OAUTH_CLIENT_STORAGE_FACTORY`.

## Security fix

- Rejects factory paths that are not explicitly allowlisted before importing them.
- Ships the allowlisted `mcp_atlassian.storage.redis:factory` adapter used by the Helm documentation.
- Removes the unshipped PostgreSQL entry instead of advertising a path that cannot work with this release.
- Keeps downstream custom factories opt-in through an explicit allowlist change in the downstream package.

## Testing

- Unallowlisted `os:system` is rejected before import.
- The shipped Redis factory is loaded through the production allowlist without patching it in the test.
- `uv run pytest tests/unit/servers`
- `uv run pytest tests/e2e/cloud --cloud-e2e`
- `uv run pytest tests/e2e --dc-e2e`

## Disclosure

Reported via GitHub Advisory (PVRA). Env-var control is a separate trust boundary, but importing an unvalidated module path is a known RCE vector.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -108,7 +108,17 @@ The factory callable should return an async key/value compatible storage object
 used by FastMCP OAuth proxy client registration storage. Factory paths must be
 listed in `ALLOWED_STORAGE_FACTORIES`; unknown paths are rejected before Python
 imports the requested module. Add a downstream factory path to that allowlist
-when building an image with a custom storage backend.
+when building an image with a custom storage backend. The built-in Redis
+factory accepts the backend's keyword arguments through
+`configJson`, for example:
+
+```yaml
+oauthClientStorage:
+  mode: factory
+  factory:
+    importPath: "mcp_atlassian.storage.redis:factory"
+    configJson: '{"url":"redis://redis:6379/0"}'
+```
 
 ### Health Checks and Readiness Probe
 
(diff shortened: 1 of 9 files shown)


failsafesecurity:
Hi! Friendly follow-up on this security fix. Happy to adjust the approach if you have concerns about backwards compatibility, tests, or scope. If you'd prefer a different disclosure channel such as private vulnerability reporting or a security contact, I can resubmit there.

failsafesecurity:
Hi - friendly follow-up on this security fix PR. Happy to rebase, split the change, or adjust the approach if there are compatibility or scope concerns. If you prefer a private channel for security reports, I can also resubmit via PVRA/security@.

failsafesecurity:
Hi! Friendly follow-up on this security fix. Happy to adjust the approach if you have concerns about backwards compatibility, tests, or scope. If you'd prefer a different disclosure channel such as private vulnerability reporting or a security contact, I can resubmit there.

PR #1435, opened 2026-07-06
fix(security): add format validation to AtlassianOpaqueTokenVerifier

failsafesecurity:
## Summary

`AtlassianOpaqueTokenVerifier.verify_token()` accepted any non-empty string as a valid token without cryptographic verification, introspection, or format checking. When combined with the OAuth proxy (`ATLASSIAN_OAUTH_PROXY_ENABLE=true`), any client that can reach the MCP endpoint can authenticate with a fabricated token and access all Jira/Confluence tools.

## Impact

- Complete auth bypass when OAuth proxy is enabled
- Any non-empty string is accepted as valid authentication (e.g., `Bearer fake-token-123`)
- No way to distinguish legitimate tokens from fabricated ones

## Fix

Add minimum length (20 chars) and character format (`^[A-Za-z0-9._-]+$`) validation. Tokens that don't match Atlassian token format are rejected. This is a defense-in-depth check — full token introspection should be added when Atlassian provides a stable endpoint.

## Testing

- `verify_token("fake")` → returns `None` (too short)
- `verify_token("token with spaces!")` → returns `None` (invalid chars)
- `verify_token("aB3dEfGhIjKlMnOpQrStUv")` → returns `AccessToken` (valid format)

## Disclosure

Reported via GitHub Advisory (PVRA). The opaque token design is documented as intentional, but accepting arbitrary non-empty strings provides no authentication guarantee.

### DIFF
--- a/src/mcp_atlassian/servers/main.py
+++ b/src/mcp_atlassian/servers/main.py
@@ -787,7 +787,12 @@ def _build_auth_provider() -> HardenedOAuthProxy | None:
     allowed_client_redirect_uris = _get_allowed_redirect_uris()
     allowed_grant_types = _get_allowed_grant_types()
     require_consent = is_env_truthy("ATLASSIAN_OAUTH_REQUIRE_CONSENT", "true")
-    verifier = AtlassianOpaqueTokenVerifier(required_scopes=scopes)
+    verifier = AtlassianOpaqueTokenVerifier(
+        instance_url=instance_url,
+        is_cloud=is_cloud,
+        cloud_id=os.getenv("ATLASSIAN_OAUTH_CLOUD_ID") if is_cloud else None,
+        required_scopes=scopes,
+    )
 
     return HardenedOAuthProxy(
         upstream_authorization_endpoint=upstream_authorize,
--- a/src/mcp_atlassian/utils/token_verifier.py
+++ b/src/mcp_atlassian/utils/token_verifier.py
@@ -1,35 +1,203 @@
-"""Token verifier for Atlassian opaque OAuth tokens.
+"""Token verifier for Atlassian OAuth tokens.
 
 FastMCP's OAuthProxy requires a TokenVerifier for loaded upstream access tokens.
-Atlassian OAuth tokens are opaque in many environments and there is no stable
-JWKS endpoint for verification, so we accept non-empty tokens and attach the
-required scopes.
+Atlassian tokens are opaque, so verification must be done via an upstream API
+call instead of local JWT/JWKS checks.
 """
 
 from __future__ import annotations
 
-import re
+import hashlib
+import logging
 import time
+from typing import Any
+from urllib.parse import urlparse
 
+import httpx
+from cachetools import TTLCache
 from fastmcp.server.auth.auth import AccessToken, TokenVerifier
+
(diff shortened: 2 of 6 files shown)


sooperset:
The report correctly identifies that the OAuth proxy accepts any non-empty bearer token, and the negative examples are useful. The current patch can't merge yet: a format check still accepts fabricated tokens that match the pattern, and it rejects the existing `"opaque-token"` unit-test fixture, which is why all Python-version test jobs fail.

1. Validate the token against the configured Atlassian Cloud resource or Data Center user endpoint, failing closed on unsuccessful responses, malformed responses, missing resources or scopes, and network errors. A format check can be an early filter, but it can't be the authentication decision.
2. Add regression coverage for accepted and rejected tokens, update the fixture, and cover the Cloud plus Jira and Confluence Data Center paths.
3. Run the unit suite and the Cloud/DC end-to-end checks through the OAuth proxy verifier, including valid and invalid or revoked tokens.

PR #1282 is already working on the upstream-validation design, so please coordinate with that implementation or rebase this PR onto the consolidated approach. The focus on fabricated-token rejection is valuable, and I'll review the updated evidence once the tests are green.

failsafesecurity:
Pushed a follow-up that switches the opaque-token verifier from format checks to upstream validation:

- Cloud tokens are now checked against `oauth/token/accessible-resources`
- Data Center tokens are now checked against authenticated current-user endpoints
- verifier now fails closed on bad responses, missing scopes, and request errors
- OAuth provider wiring now passes the instance context into the verifier
- added regression coverage for accepted/rejected tokens plus Cloud and DC paths

Targeted tests run locally:
`python -m pytest tests/unit/utils/test_token_verifier.py tests/unit/servers/test_oauth_proxy_build.py`

failsafesecurity:
Hi - friendly follow-up on this security fix PR. Happy to rebase, split the change, or adjust the approach if there are compatibility or scope concerns. If you prefer a private channel for security reports, I can also resubmit via PVRA/security@.

failsafesecurity:
Hi! Friendly follow-up on this security fix. Happy to adjust the approach if you have concerns about backwards compatibility, tests, or scope. If you'd prefer a different disclosure channel such as private vulnerability reporting or a security contact, I can resubmit there.

failsafesecurity:
Hi @sooperset — friendly ping on the updated token validation. I pushed the strengthened verifier (Cloud + Data Center validation, fail-closed on errors) a while back. Happy to adjust if you have any concerns about the approach or scope.

AmirF194 (commented):
Static review against d5a403d1, no live OAuth deployment to exercise the upstream calls.

The fix reads correctly: `_verify_cloud_token`/`_validate_dc_token` replace the old accept-any-token check with a real upstream call, and `_matches_cloud_instance` ties a Cloud token to the configured tenant.

Worth flagging before merge: #1282 implements the same fix for the same vulnerability (upstream `accessible-resources` validation, same TTL-cache shape), opened three months earlier. It bundles an unrelated CLI secret-flag change; this PR stays scoped to the verifier. Did not check whether the two branches conflict.

failsafesecurity:
Hi @sooperset — friendly ping. @AmirF194 reviewed the strengthened verifier favorably on Aug 26. Happy to rebase or adjust if there's anything blocking merge. Thanks for reviewing.

PR #1438, opened 2026-07-06
fix(confluence): use POST for attachment upload on Confluence Server/Data Center

### LINKED ISSUE #1163: upload_attachment / upload_attachments always fails on Confluence Data Center
## Bug Description

`upload_attachment` and `upload_attachments` tools always fail when uploading files to Confluence Data Center, regardless of file type (PNG, SVG) or file size.

## Environment

- **MCP server**: `mcp-atlassian` (installed via `uvx mcp-atlassian`)
- **Confluence**: Data Center (not Cloud) at a private instance
- **uvx version**: 0.9.22
- **VS Code** with Copilot Chat (MCP integration)

## Steps to Reproduce

1. Connect to a Confluence Data Center instance using personal token auth
2. Verify page access works (read, update pages all succeed)
3. Try to upload a file:
   ```
   upload_attachment(content_id="<valid_page_id>", file_path="/path/to/image.png", comment="test")
   ```

## Expected Behavior

File should be uploaded as an attachment to the specified page.

## Actual Behavior

Always returns:
```json
{
  "message": "Attachment uploaded successfully",
  "attachment": {
    "success": false,
    "error": "Failed to upload attachment <filename> to content <page_id>"
  }
}
```

Note the contradictory response: message says "successfully" but success is `false`.

## What Works

- `get_page` — works fine
- `update_page` — works fine (both wiki and markdown formats)
- `get_attachments` — lists existing attachments fine
- `search` — works fine
- Manual upload via Confluence UI — works fine (same user, same page)

## What Was Tried

- PNG files of various sizes (39KB to 175KB)
- SVG files (4-10KB)
- Single file via `upload_attachment`
- Multiple files via `upload_attachments`
- Different page IDs (multiple pages, all editable by the same user)

All fail with the same error. Since all other write operations (update_page) succeed with the same credentials, this appears to be a bug in the attachment upload implementation for Confluence Data Center (as opposed to Cloud).

## Workaround

Upload attachments manually via the Confluence UI, then reference them in page content via the MCP update_page tool.

kakao-edgar-p:
## Description

`confluence_upload_attachment` / `confluence_upload_attachments` always fail on Confluence **Server/Data Center**, while `get_page` / `update_page` work fine.

Root cause: the direct upload call in `_upload_attachment_direct` uses **PUT** on `/rest/api/content/{id}/child/attachment`. 
Data Center rejects PUT on that collection endpoint with **405 Method Not Allowed** — it only accepts **POST**.
Confluence Cloud keeps PUT because it provides create-or-update (new version on existing filename), which is not available on Server/DC.

Fixes: #1163

Refs (official REST API docs):
- Cloud (PUT create-or-update + POST): https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-content---attachments/
- Server/Data Center (POST only): https://developer.atlassian.com/server/confluence/rest/v1010/api-group-attachments/

## Changes

- Branch the upload HTTP method by deployment: `config.is_cloud` → **PUT** (Cloud), else **POST** (Server/DC).
- Add explanatory comment + official doc links at the call site.
- Add unit tests asserting Cloud uses PUT and Server/DC uses POST.

## Testing

- [x] Unit tests added/updated (`tests/unit/confluence/test_attachments.py`: `test_upload_attachment_cloud_uses_put`, `test_upload_attachment_server_dc_uses_post`; updated DC URL test to POST)
- [x] Manual checks performed: `uploaded PNG attachments to a real Confluence Data Center instance (PAT auth) — previously "success: false", now succeeds`

## Checklist

- [x] All tests pass locally.
- [x] Documentation updated (if needed) — n/a (code comment + doc links added)

### DIFF
--- a/src/mcp_atlassian/confluence/attachments.py
+++ b/src/mcp_atlassian/confluence/attachments.py
@@ -575,12 +575,25 @@ class AttachmentsMixin(ConfluenceClient, AttachmentsOperationsProto):
             if minor_edit is not None:
                 data["minorEdit"] = str(minor_edit).lower()
 
-            # Use PUT to support creating new versions of existing attachments
-            # PUT will create a new attachment if it doesn't exist, OR create a new
-            # version if an attachment with the same filename already exists
-            response = self.confluence._session.put(
-                url, headers=headers, files=files, data=data
-            )
+            # HTTP method differs by deployment (see #1163 / #1202):
+            #   * Confluence Cloud accepts PUT on the collection endpoint, which
+            #     creates the attachment, or a new version if the filename exists.
+            #   * Confluence Server/Data Center rejects PUT here with
+            #     405 Method Not Allowed and requires POST (POST also creates a new
+            #     version when the filename already exists on Server/DC).
+            # Official REST API docs:
+            #   Cloud (PUT create-or-update + POST):
+            #     https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-content---attachments/
+            #   Server/Data Center (POST only; PUT on collection -> 405):
+            #     https://developer.atlassian.com/server/confluence/rest/v1010/api-group-attachments/
+            if self.config.is_cloud:
+                response = self.confluence._session.put(
+                    url, headers=headers, files=files, data=data
+                )
+            else:
+                response = self.confluence._session.post(
+                    url, headers=headers, files=files, data=data
+                )
             response.raise_for_status()
 
             # Parse response
--- a/tests/unit/confluence/test_attachments.py
+++ b/tests/unit/confluence/test_attachments.py
@@ -105,13 +105,14 @@ class TestAttachmentsMixin:
 
         mock_response = Mock()
         if raise_error:
-            # Changed from .post to .put to match implementation
+            # Mock both verbs: Cloud uses PUT, Server/DC uses POST.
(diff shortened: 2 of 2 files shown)


kakao-edgar-p:
Hi @sooperset,

It looks like there may be some duplicate PRs.

I don't have a strong preference on which approach is taken, but I believe this functionality is quite important for users of these MCP servers.

Thanks again for maintaining this project and for all your work on it!

sooperset:
@kakao-edgar-p, you correctly identified that Confluence Server/Data Center needs POST on the attachment collection endpoint, and your tests clearly separate the Cloud and Server/DC paths.

That Server/DC fix is already covered by merged [#1253](https://github.com/sooperset/mcp-atlassian/pull/1253). It also includes the `X-Atlassian-Token` correction, same-name versioning fallback, and verified upload plus re-upload coverage for both Cloud and Server/DC. The merged implementation uses POST for the shared path, and the Cloud re-upload test passes, so this PR no longer has a separate user-facing fix to merge.

If you can reproduce a Cloud-only failure with the current POST path, please open a focused follow-up with:
1. The failing Cloud behavior and response.
2. A regression test that fails before the fix and passes after it.

I'm closing this PR in favor of [#1253](https://github.com/sooperset/mcp-atlassian/pull/1253).

PR #1447, opened 2026-07-09
fix(confluence): expand content metadata in CQL search results

### LINKED ISSUE #1446: confluence_search returns empty created/updated/author for every result
### Describe the bug

Every result returned by the `confluence_search` tool (and the underlying `ConfluenceFetcher.search()` CQL path) has empty `created`, `updated` and `author`, even though `confluence_get_page` returns those fields correctly for the same pages.

### Root cause

`SearchMixin.search()` calls `self.confluence.cql(cql=cql, limit=limit)` without an `expand` argument (`src/mcp_atlassian/confluence/search.py`). `cql()` hits `GET /rest/api/search`, whose results nest the page under a `content` object. Without expanding the content's history/version, the API omits that data, so `ConfluencePage.from_api_response()` has nothing to read and `created`/`updated`/`author` fall back to empty strings.

Note that on this endpoint the expand keys must be prefixed with `content.`; a bare `expand="history,version"` is **silently ignored** (verified against a Confluence Server/DC 8.x instance — only `content.history,content.version` populates the fields).

### Expected behavior

Search results should carry the same basic metadata (`created`, `updated`, `author`, `version`) that `get_page` already provides.

### Environment

- mcp-atlassian: `main`
- Deployment: Confluence Server/DC 8.x (Personal Access Token auth)

Wujidadi:
## Summary

`confluence_search` returned every result with empty `created` / `updated` / `author`. `SearchMixin.search()` called `cql()` without an `expand`, so `GET /rest/api/search` never returned the history/version data that `ConfluencePage.from_api_response()` parses.

This expands `content.history,content.version` on the CQL query. On `/rest/api/search` the page is nested under a `content` object, so the expand keys need the `content.` prefix — a bare `history,version` is **silently ignored** (verified against a live Confluence Server/DC 8.x instance; only the prefixed form populates the fields). Same fix in spirit as expanding `history` for `get_page`.

## Changes

- `src/mcp_atlassian/confluence/search.py`: add `expand="content.history,content.version"` to the `cql()` call.
- `tests/unit/confluence/test_search.py`: update the existing `cql` call assertions and add `test_search_expands_content_metadata`, which asserts `created`/`updated`/`author`/`version` flow through to the returned pages.

## Trade-off

CQL searches often return many results, so expanding history/version makes each result heavier (`get_page`, being a single page, has no such cost). I kept it always-on because "search results should carry basic metadata" matches user expectations and mirrors what `get_page` already returns. Happy to make it opt-in via a parameter if you'd prefer.

## Independent of #1445

This is orthogonal to #1445 (which adds `version_author` / `version_date` to `to_simplified_dict`). This PR is based on `main` and does not depend on it; once #1445 merges, those two fields will additionally flow into search results through the same `content.version` expand.

## Testing

- `uv run pytest tests/unit/confluence/test_search.py tests/unit/models/ -q` → **301 passed, 5 skipped**
- `uv run pre-commit run --files src/mcp_atlassian/confluence/search.py tests/unit/confluence/test_search.py` → all hooks pass (ruff, ruff-format, mypy)
- End-to-end against a live Confluence Server/DC 8.x: `confluence_search` results now include `created`/`updated`/`author` that were previously empty.

Closes #1446

### DIFF
--- a/src/mcp_atlassian/confluence/search.py
+++ b/src/mcp_atlassian/confluence/search.py
@@ -64,8 +64,14 @@ class SearchMixin(ConfluenceClient):
 
             logger.info(f"Applied spaces filter to query: {cql}")
 
-        # Execute the CQL search query
-        results = self.confluence.cql(cql=cql, limit=limit)
+        # Execute the CQL search query. Expand content.history and
+        # content.version so each result carries created/updated/author and
+        # version metadata; on the /rest/api/search endpoint these nested
+        # properties require the "content." prefix (a bare "history,version"
+        # is silently ignored).
+        results = self.confluence.cql(
+            cql=cql, limit=limit, expand="content.history,content.version"
+        )
 
         # Convert the response to a search result model
         search_result = ConfluenceSearchResult.from_api_response(
--- a/tests/unit/confluence/test_search.py
+++ b/tests/unit/confluence/test_search.py
@@ -59,13 +59,78 @@ class TestSearchMixin:
         result = search_mixin.search("test query")
 
         # Verify API call
-        search_mixin.confluence.cql.assert_called_once_with(cql="test query", limit=10)
+        search_mixin.confluence.cql.assert_called_once_with(
+            cql="test query", limit=10, expand="content.history,content.version"
+        )
 
         # Verify result
         assert len(result) == 1
         assert result[0].id == "123456789"
         assert result[0].title == "Test Page"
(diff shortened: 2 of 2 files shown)


GustavR (approved):
LGTM ✅ Confirmed against our Confluence DC (was still empty on 0.22.1) — search results carry the dates now instead of us chaining a get_page per hit. Nice regression test too. Thanks! 🙏

PR #1490, opened 2026-07-13
fix(jira): stop JIRA_INTERNAL_ONLY_PROJECTS failing open on invisible characters

sooperset:
Follow-up to #1478 (merged as 02de188), from an adversarial review of what shipped.

## The bug

`JIRA_INTERNAL_ONLY_PROJECTS` fails **open** on invisible characters. `str.strip()` does not remove zero-width characters or a BOM, so a key pasted from a browser or a spreadsheet can carry one:

```
"CC"          -> {'CC'}          guarded
"CC​"    -> {'CC​'}    NOT guarded  (silently)
"﻿CC"    -> {'﻿CC'}    NOT guarded  (silently)
```

The key then never matches its own issues, `_is_internal_only_project()` returns `False`, and the comment goes out **customer-visible** — while the operator believes the project is protected. There is no error and no warning.

For a setting whose entire job is keeping automation off a customer-facing portal, that is the wrong direction to fail.

## The fix

- Normalize zero-width characters and the BOM away, on **both** sides of the comparison — the configured keys *and* the project key extracted from an issue key. If only one side normalized, the guard would silently stop matching.
- **Warn** on an entry that still cannot match any issue after normalization, instead of dropping it silently. An operator who typos a key should find out from the logs, not from a customer.

## Also: the e2e tests weren't watching the edit path

`edit_comment` does its ServiceDesk visibility pre-check, but performs the actual write through the **core Jira API**. The tests only asserted the returned comment id — so if that write ever dropped the ServiceDesk internal-comment property, the text would reach the customer portal while the tests still passed.

Verified against a real JSM project that the property does survive the edit today, so this is a coverage gap rather than a live bug — but nothing was watching it. Both the Cloud and DC tests now re-assert `_fetch_servicedesk_comment_is_public(...) is False` after the edit.

## Verification

- unit: **3451 passed** (5 new)
- Cloud JSM e2e: **1 passed, 0 skipped** — real `service_desk` project
- DC JSM e2e: **1 passed, 0 skipped** — real Jira Service Management DC instance
- full DC e2e: **101 passed**, no regression

### DIFF
--- a/src/mcp_atlassian/jira/client.py
+++ b/src/mcp_atlassian/jira/client.py
@@ -29,7 +29,7 @@ from mcp_atlassian.utils.urls import make_ssrf_redirect_hook
 from mcp_atlassian.utils.user_agent import get_default_user_agent
 
 from ..models.jira.adf import markdown_to_adf
-from .config import JiraConfig
+from .config import JiraConfig, normalize_project_key
 
 # Configure logging
 logger = logging.getLogger("mcp-jira")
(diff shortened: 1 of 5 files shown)


sooperset:
Superseded by #1491 (merged), which carries the fix forward. Closing.

PR #1508, opened 2026-07-18
fix(confluence): separate lists that follow paragraph lines in markdown

### LINKED ISSUE #1493: [Bug]: Confluence Server/DC: Markdown list immediately after a bold header line collapses into flattened text instead of rendering as a list
### Prerequisites

### Bug Description

When adding a comment to a Confluence Server/Data Center page via confluence_add_comment, Markdown input containing an unordered list that immediately follows a bold header line (single newline, no blank line before the list) gets flattened into one paragraph. The - list markers render as literal characters instead of being converted into a proper storage-format <ul><li> list.

### Steps to Reproduce

Call confluence_add_comment with this body:

**Header:**
- item one
- item two

Some other paragraph.

### Expected Behavior

The list should render as an actual bulleted list under "Header:", per standard CommonMark rules (an unordered list marker is allowed to interrupt a preceding paragraph without a blank line in between).

### Actual Behavior

```shell
The header and both list items get merged into a single paragraph, with the - characters left in as literal text instead of becoming list markup:

Header: - item one - item two
```

### mcp-atlassian Version

0.21.0 (docker tag: latest, revision c450a161)

### Installation Method

Docker

### Operating System

Windows

### Python Version

3.13.12 (in-container; host is Windows 11 Enterprise 10.0.26100, container runs linux/amd64 via Docker Desktop)

### Atlassian Instance Type

Confluence Server / Data Center

### Client Application

Claude Code CLI v2.1.203 (running as the VS Code extension)

### Additional Context

Environment: mcp-atlassian Docker image ghcr.io/sooperset/mcp-atlassian:latest (pulled 2026-07-14), Confluence Server/Data Center, tool used confluence_add_comment, transport stdio via Docker.

Additional Notes:
- Adding a blank line before the list (**Header:**, then blank line, then - item one / - item two) works around the issue and produces a correctly rendered list.
- Reproduces consistently: every list not preceded by a blank line gets flattened the same way, across multiple headers/lists in the same comment body.
- Bold, italics, and inline code all convert correctly - this appears isolated to block-level list detection.
- Possibly specific to the Server/DC storage-format conversion path, since Cloud instances use ADF and may have more complete list-conversion support. Not verified whether Cloud hits the same issue.

Suggested Fix:
When converting Markdown to Confluence storage format for Server/DC, treat an unordered list start (-, *, +) as able to interrupt a preceding paragraph without requiring a blank line first, matching CommonMark's list-interrupts-paragraph behavior.

syf2211:
## Summary

Insert a preprocessing blank line before Markdown list markers that directly follow a paragraph line, so unordered and ordered lists render correctly in Confluence storage format.

## Motivation

Python-Markdown (used by md2conf) does not implement CommonMark rule that a list may interrupt a paragraph without a blank line. Input like `**Header:**
- item one` was flattened into a single paragraph with literal `-` characters instead of a `<ul>` list (reported in #1493).

## Changes

- Add `ConfluencePreprocessor._ensure_list_paragraph_separation()` and call it at the start of `markdown_to_confluence_storage()`.
- Skip preprocessing inside fenced code blocks; do not alter lists that already follow headings, blockquotes, or other list items.
- Add `TestListParagraphSeparation` regression tests covering unordered lists, ordered lists, idempotency, headings, nested lists, and code fences.

## Tests

```bash
uv run pytest tests/unit/preprocessing/test_content_processing.py::TestListParagraphSeparation -v
uv run pytest tests/unit/preprocessing/ -q
```

Result: 6 new tests pass; all 173 preprocessing unit tests pass.

## Notes

- Lists immediately after a blockquote line (`> quote
- item`) are intentionally unchanged to avoid disturbing blockquote structure.

Fixes #1493

### DIFF
--- a/src/mcp_atlassian/preprocessing/confluence.py
+++ b/src/mcp_atlassian/preprocessing/confluence.py
@@ -78,6 +78,8 @@ class ConfluencePreprocessor(BasePreprocessor):
             "menu",
             "menuitem",
             "nav",
+            "noframes",
+            "noembed",
             "ol",
             "p",
             "pre",
@@ -106,6 +108,9 @@ class ConfluencePreprocessor(BasePreprocessor):
         re.IGNORECASE,
     )
     _HTML_COMMENT_OPEN_PATTERN = re.compile(r"^ {0,3}<!--")
+    _HTML_PROCESSING_INSTRUCTION_OPEN_PATTERN = re.compile(r"^ {0,3}<\?")
+    _HTML_DECLARATION_OPEN_PATTERN = re.compile(r"^ {0,3}<!(?!--|\[)")
+    _HTML_CDATA_OPEN_PATTERN = re.compile(r"^ {0,3}<!\[CDATA\[")
 
     def __init__(self, base_url: str) -> None:
         """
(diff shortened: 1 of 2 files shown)


syf2211:
Thanks for the detailed review. I've pushed an update that addresses both blockers:

1. **List-interrupt predicate** — separated `_is_list_line()` (general detection) from `_may_interrupt_paragraph_list_line()` (stricter CommonMark interrupt rules). Only unordered items with content and ordered lists starting at `1.` can interrupt a paragraph now. Added regression tests for `Intro\n2. prose` and `Intro\n- `.

2. **Raw HTML blocks** — the scanner now tracks whole-line HTML block open/close tags (e.g. `<pre>`) and skips blank-line injection inside them. Added a `<pre>` regression test asserting exact content preservation.

All 14 `TestListParagraphSeparation` tests pass locally.

Re: Data Center e2e — understood that the configured instance was unavailable; happy to re-run once it's back if needed.

syf2211:
Thanks for the second review — pushed 1654dac addressing both blockers:

1. **Fence ordering** — active raw HTML, comment, CDATA, and processing-instruction states are now handled before fence recognition. Literal `` ``` `` inside `<pre>` or `<!-- -->` no longer leaves `in_fence` set after the block ends. Regression tests cover both cases plus a real list after `</pre>` / `-->`.

2. **Additional raw HTML forms** — added handling for processing instructions (`<?...?>`), CDATA (`<![CDATA[...]]>`), declarations (`<!...>`), and `noframes`/`noembed` block tags. Byte-for-byte helper regressions plus post-block list resumption tests for each form.

All 30 `TestListParagraphSeparation` tests pass locally. Ready for another look.

PR #1509, opened 2026-07-19
fix(jira): return epic children when the first lookup finds none

trevorh:
Fetching an epic's children tries several lookup methods in turn (scoped
search, parent field, Epic Link, …). If the first method ran but found
nothing, the code stopped there and reported no children — because it
checked the result object for truthiness, and that object is always
truthy even when empty.

Now it checks whether the result actually contains issues, so an empty
result falls through to the remaining methods. Includes a regression test.

### DIFF
--- a/src/mcp_atlassian/jira/epics.py
+++ b/src/mcp_atlassian/jira/epics.py
@@ -501,7 +501,11 @@ class EpicsMixin(
                     logger.info(f"Trying to get epic issues with issueFunction: {jql}")
 
                     search_result = self.search_issues(jql, start=start, limit=limit)
-                    if search_result:
+                    # A JiraSearchResult is always truthy (a populated model),
+                    # so gate on its issues: an empty result must fall through
+                    # to the parent / Epic Link strategies below rather than
+                    # short-circuiting the whole discovery chain with [].
+                    if search_result.issues:
                         logger.info(
                             f"Successfully found {len(search_result.issues)} issues for epic {epic_key} using issueFunction"
                         )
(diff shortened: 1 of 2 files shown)


src/mcp_atlassian/jira/epics.py
```diff
@@ -501,7 +501,11 @@ def get_epic_issues(
                     logger.info(f"Trying to get epic issues with issueFunction: {jql}")
 
                     search_result = self.search_issues(jql, start=start, limit=limit)
-                    if search_result:
+                    # A JiraSearchResult is always truthy (a populated model),
+                    # so gate on its issues: an empty result must fall through
+                    # to the parent / Epic Link strategies below rather than
+                    # short-circuiting the whole discovery chain with [].
+                    if search_result.issues:
```
Sanjays2402, line 508: does this need to distinguish "no matches" from an empty page? On Server/DC, `start` is honored, so `issues=[]` can still have `total > 0`; falling through then switches to a different JQL strategy instead of returning the requested empty page.

trevorh: Hey @Sanjays2402, good catch. It is a real edge on Server/DC for a direct `start > 0` call, where an empty & past the end page falls through to a different strategy instead of returning an empty page. However, want to point out that it's currently dormant: `get_epic_issues` has no callers beyond the epic children fetch and it iself isn't exposed as a tool.

The epic summary work in #1285 sidesteps it. It fetches the first page at `start=0` and then pages follow-up Server/DC requests against the single winning JQL rather than rerunning the discovery chain, so a later empty page ends paging cleanly instead of switching strategies. If we want to harden the general `get_epic_issues`
  pagination contract itself, that's a good small follow up. I'm happy to pick it up separately - lmk.

PR #1518, opened 2026-07-22
fix(jira): honor explicit null when clearing an issue's parent

### LINKED ISSUE #1500: [Bug]: jira_update_issue silently drops `parent: null` (clearing epic membership fails but returns success)
### Summary

`jira_update_issue` **silently drops `parent: null`** (and a null Epic Link customfield). The field is skipped, the parent is left unchanged, but the tool still returns a **success** response — so callers cannot tell the write failed. **Setting** a parent works fine; only **clearing** it is broken. The net effect is that it's impossible to unlink an issue from its epic via `jira_update_issue`, and the failure is silent.

### Environment

- `mcp-atlassian` 0.21.1
- Jira **Cloud**, company-managed project
- Transport: stdio; auth: API token (Basic)

### Steps to reproduce

1. Link an issue to an epic (works):
   ```
   jira_update_issue(issue_key="PROJ-123", fields='{"parent": {"key": "EPIC-1"}}')
   ```
   → `parent` and the legacy Epic Link are both set. ✅
2. Try to clear it:
   ```
   jira_update_issue(issue_key="PROJ-123", fields='{"parent": null}')
   ```
   → Tool returns success, but the parent is **unchanged**. Server log:
   ```
   WARNING - mcp-jira - Invalid parent value for issue PROJ-123: None
   ```
3. Same for clearing the Epic Link customfield directly:
   ```
   jira_update_issue(issue_key="PROJ-123", fields='{"customfield_10014": null}')
   ```
   → Server log: `WARNING - mcp-jira - Skipping field 'customfield_10014' due to formatting error or invalid value.` Field unchanged; success returned.

### Expected behavior

Either actually clear the field when `null` is passed, **or** return an explicit error. Silently skipping a requested field while reporting success is data-loss-masking: automation believes the epic was unlinked when it wasn't.

### Actual behavior

The field is silently left unchanged and a success response is returned. No error surfaces to the caller.

### Impact

Any workflow that needs to remove an issue from an epic (e.g. re-triage / re-parenting automation) silently no-ops. Because there's no error, downstream logic assumes success and the issue lingers in the old epic indefinitely.

### Verified workaround (Cloud)

A direct REST call bypasses the tool's field handling and works:
```
PUT /rest/api/3/issue/{key}
{"fields": {"parent": null}}          # HTTP Basic auth (email:token) → 204
```
On Cloud company-managed projects this clears the native `parent` **and** the synced legacy Epic Link, dropping the issue from `"Epic Link" = <epic>` membership.

### Related

- #1221 — feature request to support unlinking an issue from an epic. This issue is the underlying **silent-failure bug**: `jira_update_issue` already accepts `parent`/Epic-Link `null` and returns success, so at minimum it should error instead of silently no-op'ing (and ideally clear the field, which the REST PUT above shows is possible).

amitvijapur:
Fixes #1500

## Cause

IssuesMixin.update_issue previously dropped an explicit null or empty parent value. No PUT was sent, but the call returned the unchanged issue as if the clear had succeeded.

## Fix

- On Jira Cloud, treat None and an empty string as an explicit clear and send a null parent value.
- On Jira Server/Data Center, reject parent clearing before any update request with a clear ValueError. The error directs users to clear the Epic Link custom field directly, using the field ID returned by jira_search_fields.
- Preserve existing parent-setting behavior and continue warning on genuinely invalid parent values.
- Document the Cloud-only limitation and Server/DC alternative in the tool schema, override, and generated reference.

## Evidence

- Edition-specific unit regressions cover Cloud null transmission and Server/DC early rejection for both keyword and fields input styles.
- 352 targeted Jira issue/server tests passed.
- 3,890 unit tests passed; mocked integration tests passed after clearing the host-specific NO_PROXY setting.
- 38 Jira DC end-to-end tests passed.
- A live Jira Cloud lifecycle set a Task parent, cleared it with parent=None, and confirmed the parent was absent on re-fetch.
- Generated tool documentation check passed for all 98 tools, and all eight PR checks are green.


amitvijapur:
Fixed in 0fe771f: the `fields` example now uses double-quoted JSON (`{"key": "PROJ-123"}`) so it survives `json.loads`, and `docs/tools/jira-issues.mdx` is regenerated with `scripts/generate_tool_docs.py`.

Also merged main in at 2633ff4 to clear the stale base. Full unit suite passes locally: 3722 passed, 5 skipped.

The changes-requested is still sitting on the 22 July snapshot, so this needs a re-review to unblock whenever you get a chance.

amitvijapur:
Bumping this one. The requested change landed in 0fe771f on 1 August, and the branch is current with main at 7d760dd with all eight checks green.

The `CHANGES_REQUESTED` is still sitting on the 22 July review of 45e2ca6, which predates the fix, so the PR stays blocked until someone re-reviews. The flagged example is now valid double-quoted JSON and `docs/tools/jira-issues.mdx` has been regenerated from `scripts/generate_tool_docs.py`.

Nothing outstanding from my side.

amitvijapur:
Rebased onto current main in `44f3f70` — the branch is no longer conflicting.

Three conflicts, all additive, so both sides are kept in each:
- `servers/jira.py` and `docs/tools/jira-issues.mdx`: main added the `{status:color=green|title=Done}` lozenge to the `fields` description while this branch added the parent set/clear text. The merged description carries both.
- `tests/unit/servers/test_jira_server.py`: main added `test_update_issue_plain_text_fields_error_names_fields`, this branch added `test_update_issue_clears_parent_with_json_null`. Both are present and both pass.

`pytest tests/unit/servers/test_jira_server.py tests/unit/jira/test_issues.py` is 352 passed on the merge commit.

On the review from 22 July: the `fields` schema example was corrected to JSON syntax in `0fe771f` on 1 August, so that point is addressed — the review just predates the fix. Happy to take another look if anything else is outstanding.

Disclosure: written with AI assistance, reviewed by me before pushing.

PR #1519, opened 2026-07-22
fix(confluence): enforce CONFLUENCE_SPACES_FILTER outside confluence_search

### LINKED ISSUE #1495: [Bug]: CONFLUENCE_SPACES_FILTER is only enforced in confluence_search, not in other Confluence read tools
### Prerequisites

### Bug Description

`CONFLUENCE_SPACES_FILTER` is documented (`.env.example:233`) as limiting "searches **and other operations**" to an allowlist of spaces, but the code only enforces it inside `confluence_search`. Every other Confluence read tool (`confluence_get_page`, `confluence_get_space_page_tree`, `confluence_get_page_children`, `confluence_get_page_ancestors`, `confluence_get_comments`, etc.) ignores the filter entirely, so an agent (or user) can trivially read content from spaces outside the configured allowlist as long as the underlying credentials have access.

### Steps to Reproduce

1. Set `CONFLUENCE_SPACES_FILTER=ANYSPACE` (a single, narrow space).
2. Ask an MCP client to search for content that actually lives in a different space (e.g. `OTHERSPACE`).
3. `confluence_search` correctly returns an empty result set — the filter works as intended here.
4. The agent falls back to `confluence_get_space_page_tree(space_key=OTHERSPACE, ...)` and then `confluence_get_page(page_id=...)` for a page found in that tree.
5. Both calls succeed and return full page content from `OTHERSPACE`, even though it's outside the configured `CONFLUENCE_SPACES_FILTER` allowlist.

### Expected Behavior

`CONFLUENCE_SPACES_FILTER` should be a hard boundary enforced consistently across all Confluence tools that accept or resolve a `space_key`/page, not just `confluence_search`. Compare with the Jira equivalent (`JIRA_PROJECTS_FILTER`), which is enforced both in search (`src/mcp_atlassian/jira/search.py`) and in direct single-item lookups — `jira_get_issue` explicitly rejects issues whose project prefix isn't in the allowlist (`src/mcp_atlassian/jira/issues.py:104-123`).

### Actual Behavior

```shell
Only `confluence_search` applies the filter, via `SearchMixin.search()` (`src/mcp_atlassian/confluence/search.py:113-115`):


for filter_str in (self.config.spaces_filter, spaces_filter):
    if filter_str:
        cql = self._and_spaces_filter(cql, filter_str)


No equivalent check exists in `src/mcp_atlassian/confluence/pages.py`, which backs `get_page`, `get_space_page_tree`, `get_page_children`, `get_child_pages`, `create_page`, etc. — none of these read `self.config.spaces_filter` at all.
```

### mcp-atlassian Version

v0.22.1

### Installation Method

### Operating System

### Python Version

3.14.4

### Atlassian Instance Type

Confluence Server / Data Center

### Client Application

opencode + mcp via uvx

### Additional Context

_No response_

amitvijapur:
Fixes #1495

## Summary

Enforce `CONFLUENCE_SPACES_FILTER` before every supported Confluence content read or write, including page creation, updates, section updates, moves, copies, ancestors, history and version diffs, analytics, attachments and images, permissions, restrictions, labels, templates, and comment replies.

Cloud OAuth and Cloud subtype operations resolve page and space IDs through the matching v2 adapter. Unknown v2 space IDs fail closed instead of being treated as allowed.

## Validation

- `uv run pytest -q tests/unit/confluence/test_pages.py tests/unit/confluence/test_analytics.py tests/unit/confluence/test_v2_adapter.py` — 196 passed
- `uv run pytest tests/unit/ -xvs` — 3744 passed, 5 skipped
- `uv run pytest tests/e2e/test_confluence_dc_operations.py tests/e2e/test_images_dc.py --dc-e2e` — 21 passed
- `pre-commit run --all-files` — passed
- `uv run python scripts/generate_tool_docs.py --check` — passed

The full Cloud suite reached 93 passed and 15 skipped; one unrelated restrictions test received HTTP 403 from the configured Cloud account.


amitvijapur:
43ea6bf looks right, and it closes the two I missed. `update_page` takes `parent_id` straight from the tool layer, so it had the same exposure as create rather than a narrower one, and the parent on `move_page` did need checking even when a target space is supplied. Selecting the adapter before the parent lookup is also the better shape, given the subtype path writes through v2 on non-OAuth Cloud.

Merged main in at d0f885a to clear the stale base. Full unit suite green locally: 3789 passed, 5 skipped.

Nothing outstanding from my side, so this is ready for the security review whenever you are.

amitvijapur:
@sooperset when you get a chance, this one is waiting on the maintainer security review flagged in the 22 July review, which is the last gate on it.

Since then the boundary has been closed across the remaining paths, including the attachment and Cloud v2 gaps, and the branch is current with main at 084e331 with all checks green. Full unit suite passes locally: 3836 passed, 5 skipped.

Happy to split it into smaller pieces if that makes the access-control review easier.

amitvijapur:
Closing this in favour of a reviewable sequence.

This PR tried to cover every Confluence tool in one change: 25 files and +2494. Looking at what actually merges here, the median fork PR is 3 files and +70 lines and the 75th percentile is 5 files and +172, so this was never going to get the review it needed. That is my mistake in how I scoped it, not a comment on anyone's responsiveness.

The change splits along a real seam rather than arbitrarily. Some tools already hold the space key and can be checked with no extra API call; the rest are addressed by page or content ID and need a lookup first, which changes the request count for filtered clients and deserves to be reviewed on its own.

First slice is #1648 — 3 files, +78, covering the space-listing disclosure in `get_spaces` and `get_user_contributed_spaces`. I will follow with the remaining tools in similar increments once that lands, so each one can be judged on its own.

No work is lost; the full change stays on my fork's branch.

PR #1527, opened 2026-07-25
fix(confluence): convert code macros to fenced Markdown blocks (read-only)

hacctarr:
### Bug Description

`BasePreprocessor.process_html_content` has no handler for
`<ac:structured-macro ac:name="code">`, so storage-format code blocks reach
markdownify as unknown tags. markdownify strips the tags but keeps their
text. The `language` and `linenumbers` parameter values end up glued to the
first code line (`texttrue` for a text-language block with line numbers), the
fence is gone, and whitespace normalization flattens the body's indentation.
Every page with a code block comes back corrupted in markdown mode. Nothing
downstream can repair it; the fencing and indentation are destroyed
server-side.

### Steps to Reproduce

1. Create a Confluence page with a code macro:
   `<ac:structured-macro ac:name="code"><ac:parameter ac:name="language">text</ac:parameter><ac:parameter ac:name="linenumbers">true</ac:parameter><ac:plain-text-body><![CDATA[indented\n  body]]></ac:plain-text-body></ac:structured-macro>`
2. Fetch it with `confluence_get_page` in default markdown mode.
3. Inspect the returned content.

### Expected Behavior

A fenced code block with the language tag, body verbatim:

    ```text
    indented
      body
    ```

### Actual Behavior

`texttrueindented body` (parameters leaked as text, no fence, indentation
collapsed).

### Fix

Adds `_process_code_macros_in_soup`, which swaps `code` and `noformat` macros
for `<pre><code class="language-X">` before the markdown conversion. `md()`
now gets a `code_language_callback` so the language class lands on the fence
(markdownify ignores `language-*` classes without one).

Two design calls worth flagging:

- Markdown path only. `processed_html` is serialized before the macro
  rewrite, so the HTML output keeps the original storage-format macros and
  the existing `test_confluence_macro_preservation` passes unchanged.
- If a macro's body text is unreachable, the macro is left alone instead of
  emitting an empty fence that would silently drop content.

The markdown-to-storage direction is untouched.

### Tests

Four new tests in `tests/unit/preprocessing/test_code_macro_conversion.py`
cover the fence and language tag, the parameter leak, CDATA indentation
(box-drawing characters included), the unchanged HTML output, and multiple
macros on one page. Full suite: 3677 passed, 240 skipped (all skips are the
marker-gated e2e/integration tests). Conversion behavior is verified on both
ends of the declared beautifulsoup4 range (the 4.12.3 floor and the locked
4.14.3).

### DIFF
--- a/src/mcp_atlassian/preprocessing/base.py
+++ b/src/mcp_atlassian/preprocessing/base.py
@@ -88,7 +88,11 @@ def _fenced_code_block(code_text: str, language: str) -> str:
 
 def _prefix_code_block(code_block: str, prefix: str) -> str:
     """Add a Markdown container prefix to every line in a code block."""
-    return code_block.replace("\n", f"\n{prefix}")
+    return re.sub(
+        r"\r\n|\r|\n",
+        lambda match: f"{match.group()}{prefix}",
+        code_block,
+    )
 
 
 def _restore_inline_code_macro(
@@ -175,11 +179,16 @@ def _restore_code_macro_block(
         return text.replace(placeholder, code_block, 1)
 
     body = code_block[stored_opening_end : -len(stored_fence)]
+    # The closing line contains only markdownify's continuation prefix and
+    # the fence, making it the reliable prefix for every restored body line.
+    # The placeholder line can also carry list or quote context, but unlike
+    # the closing line it may be adjacent to content in malformed HTML.
+    continuation_prefix = closing_line[: closing_match.start()]
     container_prefix = opening_line[: opening_match.start()]
     stored_opening_line = code_block[:stored_opening_end].rstrip("\n")
     restored = (
         f"{container_prefix}{stored_opening_line}\n"
-        f"{line_before}{_prefix_code_block(body, line_before)}"
+        f"{continuation_prefix}{_prefix_code_block(body, continuation_prefix)}"
         f"{stored_fence}"
     )
 
--- a/src/mcp_atlassian/preprocessing/base.py
+++ b/src/mcp_atlassian/preprocessing/base.py
@@ -86,33 +86,100 @@ def _fenced_code_block(code_text: str, language: str) -> str:
     return f"{fence}{language}\n{code_text}{closing_newline}{fence}"
 
 
-_CONTAINER_MARKERS_RE = re.compile(
-    r"(?:[ \t]*(?:>|[*+-][ \t]|\d{1,9}[.)][ \t]))*[ \t]*"
-)
+def _markdown_container_prefix(line: str) -> tuple[str, str]:
+    """Return a Markdown container prefix and its continuation indentation."""
+    prefix = ""
+    continuation = ""
+    position = 0
+
+    while position < len(line):
+        quote_match = re.match(r">[ ]?", line[position:])
+        if quote_match:
+            token = quote_match.group()
+            prefix += token
+            continuation += token
+            position += len(token)
+            continue
+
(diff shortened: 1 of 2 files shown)


hacctarr:
Fixed in d4f646b. The whitespace guard is gone: empty and whitespace-only bodies now convert (an empty body renders as an empty fenced block with the language tag), so the parameter values no longer leak. Regression tests added for both cases. The guard's original premise didn't hold anyway; CDATA bodies are reachable via get_text() across the supported bs4 range (checked 4.12.3 and 4.14.3).

ebarkhordar (commented):
The macro body is restored by substituting the placeholder into the finished markdown, and that substitution has no idea where the placeholder ended up. When a code macro sits inside a list item or a table cell, the fence lands mid-line and is no longer a fence.

At head b65c9400:

```python
from mcp_atlassian.preprocessing.base import BasePreprocessor

MACRO = ('<ac:structured-macro ac:name="code">'
         '<ac:parameter ac:name="language">python</ac:parameter>'
         '<ac:plain-text-body><![CDATA[run me]]></ac:plain-text-body>'
         '</ac:structured-macro>')
p = BasePreprocessor(base_url="https://c.example.com")

_, md = p.process_html_content("<ul><li>step one:" + MACRO + "</li><li>step two</li></ul>")
# '* step one:```python\nrun me\n```\n* step two'
```

Rendered (markdown-it-py, commonmark + table) that is `<ul><li>step one:```python run me</li></ul>` followed by `<pre><code>* step two</code></pre>`: the body is absorbed into the item text and the next item turns into an indented code block. Inside a table cell the row splits in two and the trailing pipe opens a block with `class="language-|"`.

markdownify handles both positions itself. `<pre><code>` inside an `<li>` comes back as `'* step one:\n\n  ```\n  run me\n  ```\n* step two'`, and in a table cell it collapses onto one line so the row survives.

Routing the body around markdownify is still the right call for the fence width, since markdownify emits exactly three backticks regardless of the body and `print('```')` does come back broken through it. It is the restoration that needs the block context back: a blank line plus the enclosing indent when the placeholder is not at the start of a line, and an inline span rather than a fence inside a table row. Code macros in table cells are common on Confluence documentation pages, so it is worth covering before this lands.

hacctarr:
Fixed in 623ffd3. Placeholder restoration now derives a continuation prefix from the container markers on the placeholder's line: `>` markers repeat verbatim on every restored line (a bare line would end the blockquote), and list markers become equivalent indentation (repeating them would create new items), so nesting like `* > ` composes correctly. Added regression tests for a macro inside a list item (exact whitespace asserted) and inside a blockquote with embedded backticks. Full suite passes.

hacctarr:
The table-cell case @ebarkhordar raised is still broken at 918a0a1: markdownify collapses the pre/code wrapper inside a cell to an inline ` ``` placeholder ``` ` span, restoration hits the naive-replace fallback, and the spliced multi-line fence splits the row. Fixed in c53e097: restoration now detects the backtick span around the placeholder and rebuilds it with the body collapsed onto one line, mirroring markdownify's native pre-in-cell handling, with the span widened past any backtick run in the body. Regression tests added for single-line, multi-line (row survival asserted), and embedded-backtick bodies. Full suite passes.

hacctarr:
All items from the 2026-07-28 review are addressed: the table-cell restoration gap is fixed (inline-span handling with row-survival regression tests), and the branch is rebased cleanly onto current main with the full suite and CI green. @sooperset could you re-trigger the review when you get a chance? The @claude trigger 401s for non-collaborators, so I cannot summon it from my side.

PR #1528, opened 2026-07-25
fix(auth): defer OAuth challenge to auth provider and FastMCP

### LINKED ISSUE #1257: [Feature]: Add WWW-Authenticate: Bearer resource_metadata to 401 responses when OAuthProxy is configured (RFC 9728)

When OAuthProxy is configured as the server's auth provider, unauthenticated requests to the MCP HTTP endpoint return a plain 401 Unauthorized with no WWW-Authenticate header. Reactive MCP clients (Cursor, Claude Desktop, and any client following [MCP 2025-06-18 spec §4 (Authorization)](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization)) expect this header to bootstrap OAuth automatically. Without it, they have no machine-readable way to discover the authorization server and either stall or require manual configuration.

When OAuthProxy is attached and an unauthenticated request is rejected with 401, include:

WWW-Authenticate: Bearer resource_metadata="https://your-server.com/.well-known/oauth-protected-resource/mcp"
per [RFC 9728 §5.1](https://www.rfc-editor.org/rfc/rfc9728#section-5). The resource_metadata URL should point at the /.well-known/oauth-protected-resource/<mcp-path> endpoint that FastMCP already registers whenever OAuthProxy is active — no new endpoints needed. The header value can be constructed from OAuthProxy.base_url (already set at construction time) and the MCP mount path.

_No response_

### Use Case

A developer configures OAuthProxy and points a supported MCP client at the server URL. The expected end-to-end flow is:

Client calls POST /mcp without a token → 401 + WWW-Authenticate
Client GETs resource_metadata URL → reads existing FastMCP PR metadata JSON
Client discovers authorization server, runs DCR + OAuth
Client retries POST /mcp with Authorization: Bearer <token>
Currently step 1 produces a 401 with no headers, so the client never reaches step 2. The fix is the single missing header that FastMCP is already in a position to emit.

### Additional Context

- This is required by the [MCP 2025-06-18 spec §4 (Authorization)](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization)
- FastMCP already serves /.well-known/oauth-protected-resource/<mcp-path> when OAuthProxy is mounted — the metadata the header advertises already exists and is correct. Only the 401 challenge is missing.
- The change is isolated to the ASGI layer that intercepts unauthenticated requests: read OAuthProxy.base_url and the MCP mount path, build the header value, append it to the 401 response.
- References: [RFC 9728](https://www.rfc-editor.org/rfc/rfc9728), [RFC 8414](https://www.rfc-editor.org/rfc/rfc8414)

cblacknygc:
## Description

When an OAuth provider is configured, `UserTokenMiddleware` currently rejects MCP requests with no credentials before FastMCP can handle them. Clients therefore receive a local `401` without the RFC 9728 discovery challenge needed to start OAuth.

Without the RFC 9728 challenge, MCP clients cannot discover authorization-server metadata or DCR endpoints, so OAuth fails before authorization begins. Restoring this spec-defined flow enables standards-based clients—including Codex-compatible MCP connectors—to register and authenticate without static or sentinel credentials.

This change defers only the OAuth-provider + missing-credentials case to FastMCP, which returns:

```http
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Bearer resource_metadata="https://host/.well-known/oauth-protected-resource/mcp"
```

Middleware-only deployments remain fail-closed, authenticated bearer parsing and request state are preserved, and `ALLOW_GLOBAL_CRED_FALLBACK=false` continues to prevent fallback to operator credentials.

The PR also binds each validated outer FastMCP bearer to its mapped upstream Atlassian token set. This preserves the correct Jira OAuth token for tool calls and prevents cross-user MCP session reuse.

Fixes: #1257

## Changes

- Defer missing OAuth credentials to FastMCP so it can emit the protected-resource discovery challenge.
- Preserve middleware rejection when no OAuth provider exists, including PAT/basic and external-gateway configurations.
- Resolve validated outer bearer tokens to their associated upstream Jira OAuth token and stable session owner.
- Reject invalid, unknown, replayed, and cross-provider/cross-user tokens before they can reach Jira or an MCP tool handler.
- Add ASGI and unit regression coverage for discovery, fail-closed behavior, request state, upstream-token resolution, and session isolation.

## Testing

- [ ] Integration tests passed (not run; no live Atlassian credentials required)
- [x] Manual checks performed: ASGI route probe before and after the fix

Validation completed:

- Focused auth/server suite: `207 passed`
- Full unit suite: `3682 passed, 5 skipped`
- `pre-commit run --all-files`: all hooks passed, including Ruff and mypy
- ASGI probe: missing OAuth credentials now return `401` with the expected `WWW-Authenticate: Bearer resource_metadata=...` header

## Compatibility and scope

- No `IGNORE_HEADER_AUTH`, proxy sentinel token, or global credential fallback is introduced.
- PAT/basic authentication and external-gateway behavior retain their existing paths.
- Path-qualified protected-resource discovery works behind a public URL prefix. FastMCP's separate path-qualified authorization-server metadata routing limitation is tracked upstream in PrefectHQ/fastmcp#4390 and is intentionally outside this PR.

## Checklist

- [x] All tests pass locally.
- [x] Documentation updated (not required; no public tool signature or registration changed).

### DIFF
--- a/src/mcp_atlassian/servers/main.py
+++ b/src/mcp_atlassian/servers/main.py
@@ -487,15 +487,16 @@ class UserTokenMiddleware:
             await self._send_json_error_response(safe_send, 401, auth_error)
             return  # Don't call self.app - request is rejected
 
-        # Reject unauthenticated MCP requests at the transport boundary: with no
-        # user identity, the downstream handler would fall back to the operator's
-        # global credentials, so any unauthenticated caller would transact as the
-        # operator. Refuse unless explicitly opted in (opt-in global-fallback gate).
+        # Without an OAuth provider, reject unauthenticated MCP requests at the
+        # transport boundary so the downstream handler cannot fall back to the
+        # operator's global credentials. When a provider is attached, defer to
+        # FastMCP so it can return the RFC 9728 OAuth discovery challenge.
         if (
             not ignore_header_auth
             and self.mcp_server_ref
             and self._should_process_auth(scope_copy)
             and not scope_copy["state"].get("user_atlassian_auth_type")
+            and self.mcp_server_ref.auth is None
             and not is_env_truthy("ALLOW_GLOBAL_CRED_FALLBACK")
         ):
             logger.warning(
--- a/tests/unit/servers/test_main_server.py
+++ b/tests/unit/servers/test_main_server.py
@@ -7,8 +7,11 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 import httpx
 import pytest
+from key_value.aio.stores.memory import MemoryStore
 
-from mcp_atlassian.servers.main import UserTokenMiddleware, main_mcp
+from mcp_atlassian.servers.main import AtlassianMCP, UserTokenMiddleware, main_mcp
+from mcp_atlassian.servers.oauth_proxy import HardenedOAuthProxy
+from mcp_atlassian.utils.token_verifier import AtlassianOpaqueTokenVerifier
 
 
 @pytest.mark.anyio
@@ -183,6 +186,7 @@ class TestUserTokenMiddleware:
         mock_mcp_server = MagicMock()
         mock_mcp_server.settings.streamable_http_path = "/mcp"
         mock_mcp_server.get_streamable_http_path.return_value = "/mcp"
+        mock_mcp_server.auth = None
         return UserTokenMiddleware(mock_app, mcp_server_ref=mock_mcp_server)
 
     @pytest.fixture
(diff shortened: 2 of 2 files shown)


cblacknygc:
FastMCP v3 comes w/ many oauth changes and mcp spec implementation improvements. Allowing FastMCP to handle the response or failing safely otherwise is a bit more robust and secure while also improving DCR.
As part of this work I had codex generate extra tests including greatly expanding tests/unit/servers/test_oauth_proxy_build.py. This also uncovered another place where mcp-atlassian can utilize code paths in FastMCP v3.

PR #1536, opened 2026-07-28
feat(server): add TLS support for the HTTPS listener

endofcake:
## Summary

Adds native TLS termination to the `sse`/`streamable-http` transports, so the server can
serve HTTPS directly — for deployments where policy requires encryption all the way to
the process and terminating at a reverse proxy/ingress is not enough.

- `--ssl-certfile` / `--ssl-keyfile` CLI options, with `MCP_SSL_CERTFILE` /
  `MCP_SSL_KEYFILE` env fallback (CLI wins, mirroring host/port precedence — including a
  provided CLI pair overriding blank env vars).
- Forwarded to uvicorn through FastMCP's existing `run_http_async(uvicorn_config=...)`
  seam — no new dependencies.
- Under `stdio` the options are ignored entirely with a single warning — from either
  source: the CLI options carry no parse-time validation at all, so neither a missing
  path nor a directory value can abort a stdio start.

## Fail-closed behavior

Every HTTP-transport misconfiguration aborts startup with a clear message rather than
serving plaintext or failing deep inside uvicorn: cert and key must be provided
together; a selected value that is *set but blank* (e.g. a template rendering an unset
value as `""`) is a configuration error from either source, not "TLS off"; resolved
paths must be regular files; and the pair is preflight-loaded so a mismatched pair,
non-PEM content, or an encrypted key (unsupported) surfaces as a clean startup error
instead of an opaque uvicorn failure or an interactive OpenSSL passphrase prompt.

## Cipher pinning

uvicorn's default `ssl_ciphers="TLSv1"` OpenSSL group contains NULL-encryption and
anonymous suites and no AEAD suites, so the negotiated strength varies with the base
image's OpenSSL security level (a `SECLEVEL=0` image will complete an
authenticated-but-unencrypted handshake). The listener pins
`ECDHE+AESGCM:ECDHE+CHACHA20:!aNULL:!eNULL:!MD5`, which also excludes TLS 1.0/1.1 (no
AEAD suites exist for them). TLS 1.3 suites are configured separately by OpenSSL and
stay enabled. Negotiation is regression-tested with real certificates via in-memory
handshakes: TLS 1.3 and TLS 1.2 (AEAD) negotiate; TLS 1.0 and 1.1 offers are proven
to reach the server as real ClientHellos and are refused on the server side of the
handshake (skipped with an explicit reason where the runtime cannot produce a legacy
offer at all).

## Interaction with the OS trust store injection

`truststore.inject_into_ssl()` (default on) replaces `ssl.SSLContext` with a
client-side-only context whose post-handshake peer-cert verification resets **every
inbound TLS connection** — a server's peer sends no certificate — while startup still
logs success. uvicorn builds its listener context from `ssl.SSLContext`, so the two
features cannot coexist in one process. Startup fails fast with the remediation
(`MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE=false`, set as a real env var or in the default
`.env`; `--env-file` is loaded too late to affect injection).

## Private-CA outbound (first commit)

Disabling the trust store injection moves outbound verification to certifi — but the
PAT/OAuth sessions set `trust_env=False` (so `.netrc` cannot override credentials),
which made requests ignore `REQUESTS_CA_BUNDLE`. Worse, `atlassian-python-api` passes
`verify=self.verify_ssl` on **every wrapper request**, overriding any session-level
value — so normal Jira/Confluence API calls needed the fix at the wrapper seam, not
just the session. The first commit resolves `REQUESTS_CA_BUNDLE`/`CURL_CA_BUNDLE` in
`configure_ssl_verification` (single seam, all services): applies it to
`session.verify`, accepts both bundle files and OpenSSL-style hashed CA directories,
returns the effective verify value, and the Jira/Confluence clients propagate it to the
wrapper's `verify_ssl`. Dev-status calls no longer pass a per-request verify bool. A
bundle path that is not a regular file or directory (special files like FIFOs would
block instead of failing) aborts fast, with the path logged server-side only — fetcher
errors can reach the MCP client. Docs cover the full setup: `REQUESTS_CA_BUNDLE` for the requests clients
**and** `SSL_CERT_FILE` for the httpx-based OAuth proxy paths, with a warning against
the `*_SSL_VERIFY=false` workaround.

Note: the listener certificate is unrelated to the client-side `*_CLIENT_CERT` mTLS
options used when connecting *to* Atlassian.

## Known limitation

FastMCP's transport banner hard-codes `http://` in its "Starting MCP server" line
(fastmcp 3.4.4, `server/mixins/transport.py`); our launcher line and uvicorn's own
"running on" line both correctly print `https://`. Candidate for a small upstream
FastMCP fix.

## Testing

- 34 listener tests (`TestServerTLSConfiguration` + `TestServerTLSRealCertificates`):
  precedence (env, CLI, cross-source, CLI-over-blank-env), cipher-pin forwarding,
  half-configured/blank (env and CLI)/missing-path (env and CLI) aborts, preflight
  rejection (garbage pair, mismatched real pair) and acceptance (real pair),
  truststore-guard aborts on both HTTP transports, stdio ignores everything (env, CLI
  flags, missing paths, directory values, would-abort values), https scheme in the
  startup log, truststore-layout detection pin, and the real-certificate protocol
  matrix with server-attributed legacy rejection.
- 12 outbound-CA tests: production-seam regressions through the real `JiraClient` and
  `ConfluenceClient` (an ordinary wrapper request carries the bundle as `verify=`),
  the wrapper-mechanism test, bundle applied on `trust_env=False` sessions,
  CA-directory acceptance, special-file (FIFO) rejection, CURL fallback,
  effective-value return contract, missing-bundle fail-fast, default preserved,
  ignored when `ssl_verify=false`, dev-status covers both call sites.
- Full suite: 3719 passed on this branch. `pre-commit` (ruff-format, ruff, mypy)
  clean.
- Verified live: HTTPS listener serves `/healthz` and MCP initialize; default truststore
  mode aborts with the remediation; a private-CA request through the real `JiraClient`
  (PAT, `trust_env=False`) completes TLS end-to-end with `REQUESTS_CA_BUNDLE`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -168,6 +168,26 @@ CONFLUENCE_API_TOKEN=your_confluence_api_token_here
 # Host for 'sse' transport. Default is '0.0.0.0'.
 #HOST=0.0.0.0
 
+# --- Server TLS (HTTPS listener) ---
+# Serve the 'sse'/'streamable-http' transports over HTTPS by terminating TLS at
+# the MCP server itself. Point these at a certificate + private key (PEM).
+# NOTE: this is the SERVER's own listener cert — the opposite direction from the
+# client-side *_CLIENT_CERT mTLS options above (which authenticate the server TO
+# Atlassian). Both must be set together; setting only one, or setting either to
+# an empty string, aborts startup (empty is treated as a configuration error,
+# not as "TLS off"). Under 'stdio' transport they are ignored (no network
+# listener). Useful for end-to-end TLS where the certificate is mounted into
+# the container from a Secret. The key must be UNENCRYPTED (no passphrase
+# support), and the pair is loaded once at startup — rotation requires a
+# process restart.
+# REQUIRES MCP_ATLASSIAN_USE_SYSTEM_TRUSTSTORE=false (as a real env var or in
+# the default .env; --env-file is loaded too late): the OS trust store
+# injection is client-side only and would reject every inbound TLS handshake,
+# so startup aborts if both are enabled. See the HTTP transport docs for the
+# outbound private-CA setup that replaces the OS trust store.
+#MCP_SSL_CERTFILE=/path/to/server-cert.pem
+#MCP_SSL_KEYFILE=/path/to/server-key.pem
+
 # --- HTTP Timeout ---
 # Timeout in seconds for HTTP requests to Atlassian APIs. Default is 75 seconds.
 #JIRA_TIMEOUT=75
--- a/docs/configuration.mdx
+++ b/docs/configuration.mdx
@@ -185,6 +185,8 @@ openssl rsa -in encrypted.key -out decrypted.key
 | `STATELESS` | Enable stateless mode for streamable-http (`true`/`false`) |
 | `PORT` | Port for HTTP transports (default: `8000`) |
 | `HOST` | Host for HTTP transports (default: `0.0.0.0`) |
+| `MCP_SSL_CERTFILE` | TLS certificate (PEM) for the server's own HTTPS listener (see [HTTP Transport](/docs/http-transport#tls-https-listener)) |
+| `MCP_SSL_KEYFILE` | TLS private key (PEM) for the HTTPS listener; must be set together with the certificate |
 | `READ_ONLY_MODE` | Disable write operations (`true`/`false`) |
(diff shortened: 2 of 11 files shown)


endofcake:
Rebased on current `main`.

PR #1542, opened 2026-07-29
fix(auth): prefer request Cloud ID over DC OAuth config

### LINKED ISSUE #1096: fix(auth): per-request cloud_id overridden by global DC OAuth config
## Summary

When the OAuth proxy runs alongside Data Center environment variables (e.g., `JIRA_URL=http://localhost:8080`), the per-request `X-Atlassian-Cloud-Id` header is silently dropped. This causes Cloud OAuth API calls to fail with 401 because the Jira client hits the wrong endpoint (missing `cloud_id` means it doesn't construct `https://api.atlassian.com/ex/jira/{cloud_id}/...`).

## Root Cause

In `dependencies.py:409-410`, `is_dc_oauth` is determined solely from the **global** OAuth config:

```python
is_dc_oauth = getattr(global_oauth_cfg, "is_data_center", False) is True
```

Then at line 438, `cloud_id` is set to `None` when `is_dc_oauth` is `True`:

```python
cloud_id=effective_cloud_id if not is_dc_oauth else None,
```

The global config comes from `.env` which has DC settings (`JIRA_URL=http://localhost:8080`), so `is_data_center=True`. This overrides the explicit per-request `X-Atlassian-Cloud-Id` header value.

## Request Flow

```
Client → Bearer token + X-Atlassian-Cloud-Id: c72bbc75...
  → Middleware: extracts cloud_id ✅
  → _resolve_bearer_auth_type: returns "oauth" ✅
  → _create_user_config_for_fetcher(cloud_id="c72bbc75...")
    → global_oauth_cfg.is_data_center == True  (from DC .env)
    → is_dc_oauth = True
    → OAuthConfig(cloud_id=None)  ← LOST HERE
  → Jira API call → 401 ❌
```

## Suggested Fix

Per-request `cloud_id` should override the global DC determination:

```python
# dependencies.py:409-410
is_dc_oauth = (
    not cloud_id  # explicit per-request cloud_id → Cloud OAuth
    and getattr(global_oauth_cfg, "is_data_center", False) is True
)
```

This preserves the existing behavior when no per-request `cloud_id` is provided, while correctly handling the case where a client explicitly sends `X-Atlassian-Cloud-Id`.

## Reproduction

1. Start the OAuth proxy with DC env vars in `.env` (`JIRA_URL=http://localhost:8080`) plus Cloud OAuth proxy env vars (`ATLASSIAN_OAUTH_*`)
2. Complete the full OAuth consent flow → obtain proxy JWT
3. Call an MCP tool with `Authorization: Bearer <proxy_jwt>` and `X-Atlassian-Cloud-Id: <cloud_id>`
4. Observe: `configure_oauth_session` receives `cloud_id='None'`, Jira API returns 401

## Impact

- Only affects environments where the OAuth proxy coexists with DC env vars
- Production deployments typically use Cloud-only configuration, so impact is low
- Fix is a single line change

cblacknygc:
## Description

An explicit request `X-Atlassian-Cloud-Id` now takes precedence over a
global Data Center OAuth configuration, so per-request Cloud OAuth routing
uses the requested tenant for Jira and Confluence.

Fixes: #1096

## Changes

- Prefer an explicit request Cloud ID when classifying OAuth routing.
- Retain the existing Data Center base-URL fallback when no request Cloud ID is supplied.
- Add Jira and Confluence regression coverage for both routing paths.

## Testing

- `uv run pytest tests/unit/servers/test_dependencies.py::TestCreateUserConfigForFetcher -xvs` — 19 passed
- `uv run pytest -xvs` — 3675 passed, 240 skipped

## Checklist

- [x] Code follows project style guidelines (ruff/mypy pass).
- [x] All tests pass locally.

## References

- #1096

### DIFF
--- a/src/mcp_atlassian/servers/dependencies.py
+++ b/src/mcp_atlassian/servers/dependencies.py
@@ -723,7 +723,9 @@ def _create_user_config_for_fetcher(
         global_oauth_cfg = base_config.oauth_config
 
         # Determine if this is DC OAuth (base_url set, no cloud_id)
-        is_dc_oauth = getattr(global_oauth_cfg, "is_data_center", False) is True
+        is_dc_oauth = not cloud_id and (
+            getattr(global_oauth_cfg, "is_data_center", False) is True
+        )
 
         # Use provided cloud_id or fall back to global config cloud_id
         effective_cloud_id = cloud_id if cloud_id else global_oauth_cfg.cloud_id
--- a/tests/unit/servers/test_dependencies.py
+++ b/tests/unit/servers/test_dependencies.py
@@ -277,6 +277,49 @@ class TestCreateUserConfigForFetcher:
         )  # Should preserve minimal config
 
     @pytest.mark.parametrize("config_type", ["jira", "confluence"])
+    @pytest.mark.parametrize(
+        "cloud_id,expected_cloud_id,expected_base_url",
+        [
+            ("request-cloud-id", "request-cloud-id", None),
+            (None, None, "https://dc.example.com"),
+        ],
+    )
+    def test_oauth_request_cloud_id_overrides_global_dc_config(
+        self,
+        config_factory,
+        config_type,
+        cloud_id,
+        expected_cloud_id,
+        expected_base_url,
+    ):
+        """Request Cloud IDs take precedence over global Data Center OAuth."""
+        dc_base_url = "https://dc.example.com"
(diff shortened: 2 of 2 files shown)


cblacknygc:
Isolated 4 line diff to fix reported issue. Added parameterized tests for Jira and Confluence, each with and without a request Cloud ID.

Issue #1541, opened 2026-07-29
mcp 2.0 / fastmcp 4 upgrade: resolver is blocked, plus a GHSA-3r68 regression test that fails on the beta


Kayaba-Attribution:
`mcp` 2.0.0 (MCP protocol revision `2026-07-28`) was released on 2026-07-28. This repo pins `mcp>=1.27.0,<2.0.0` and `fastmcp>=3.2.4,<4.0.0`. I did a real dry run of the upgrade to find out what it would take, and hit one hard blocker plus one failure worth flagging early. Filing rather than sending a PR because the blocker is not in your hands.

No urgency implied: the SDK's own migration guide states v2 speaks the 2025-era protocol unless a server explicitly opts in, and the spec carries a minimum twelve-month deprecation window. `mcp` 1.x also continues receiving critical fixes.

## The blocker

Every `fastmcp` release through stable 3.4.5 transitively requires `mcp>=1.24.0,<2.0`. Only the `4.0.0` pre-release line (`4.0.0a1`, `4.0.0a2`, `4.0.0b1`) accepts `mcp>=2.0.0`. So `uv lock` fails outright on bumping `mcp` alone, and only resolves with `--prerelease=allow` once `fastmcp` is also moved to the beta. There is nothing to do here until `fastmcp` 4 ships stable.

## What I found by running it anyway

With `mcp==2.0.0` + `fastmcp==4.0.0b1`, `from mcp_atlassian.servers import main_mcp` imports cleanly, and `pytest tests/unit` gives **3688 passed / 3 failed / 5 skipped**. All three failures share one root cause:

```
AttributeError: 'ErrorPreservingFastMCP' object has no attribute '_call_tool_mcp'
```

`AtlassianMCP._call_tool_mcp` (`src/mcp_atlassian/servers/main.py:361-373`) ends in `return await super()._call_tool_mcp(key, arguments)`, and that method no longer exists on `FastMCP` in 4.0.0b1. Its replacement is a different name **and** a different parameter shape, so it is not a rename-only fix:

```
_on_call_tool(ctx: ServerRequestContext, params: CallToolRequestParams) -> CallToolResult | InputRequiredResult | BaseModel
```

The reason only call-side tests break is that `_list_tools_mcp`'s override never calls `super()`; it drives its own filtering off the still-present public `self.list_tools()`.

[...]

PR #1555, opened 2026-08-02
test(server): add public policy and stdio readiness contracts

cblacknygc:
## Summary

- add a public MCP client contract proving that write tools hidden by read-only, enabled-tool, or toolset policy are absent from discovery and remain uncallable by name
- add an SDK-only stdio probe for MCP initialize and tool discovery, with explicit protocol/tool-visibility gates and expanded package evidence
- add an opt-in, fail-closed Jira DC and Confluence DC PAT staging profile
- document hermetic and live-profile usage in the compatibility guide

## Why

The existing security regression covers the internal `AtlassianMCP` dispatch gate. The new contract exercises the same boundary through a public client, providing an end-to-end signal if framework changes cause hidden tools to become discoverable or directly callable. It strengthens the GHSA-3r68 regression contract with valid write arguments, an executor trap that must remain untouched, comparison with a genuinely unknown tool, and a positive allowed-read control.

Issue #1541 also shows why import success alone is not a sufficient MCP 2 / FastMCP 4 readiness signal. The stdio probe therefore makes the negotiated protocol revision and expected tool presence/absence gateable. It records installed `mcp`, `mcp-types`, `fastmcp`, `fastmcp-slim`, and `cryptography` versions using package metadata. This PR remains compatible with FastMCP 3.4.4, changes no production code or dependency bounds, and does not claim FastMCP 4 compatibility.

## Policy contract

The hermetic public-client regression covers:

- `READ_ONLY_MODE` hiding a valid write operation
- `ENABLED_TOOLS` excluding the write operation
- `TOOLSETS` excluding the write operation's toolset
- separate discovery and dispatch contracts so one failure cannot mask the other
- direct calls returning the same unknown-tool shape as a nonexistent name
- the Jira fetcher never being reached for a denied call
- an allowed `jira_get_issue` call still reaching its mocked executor successfully

## Staging safety

The live profile is explicit and is not run by ordinary tests or CI. It:

- forces read-only mode and an exact `jira_get_issue,confluence_get_page` allowlist
- fails before live calls if required tools are missing or any unexpected tool is exposed
- caps traffic at 30 requests per minute, paces calls, limits concurrency to one, and disables retries
- passes only an allowlisted child-process environment
- records protocol/result shape without credentials, URLs, identifiers, arguments, or response bodies

The live probe intentionally never invokes a hidden write canary: if policy regressed, such a canary could perform the write it was intended to detect. The live staging profile was not invoked during PR validation because no staging credentials or approved test identifiers were supplied.

## Validation

- `uv sync --frozen --all-extras --dev`
- `uv run pytest tests/unit/servers/test_tool_policy_contract.py tests/unit/servers/test_mcp_protocol.py tests/unit/test_mcp_wire_probe.py -q` — 51 passed
- `uv run python scripts/mcp_wire_probe.py --server-command uv --server-arg=run --server-arg=mcp-atlassian --expect-protocol-version=2025-11-25 --expect-tool-absent=jira_create_issue` — MCP 1.28.1 / FastMCP 3.4.4 initialized successfully; no Atlassian tools were called
- `uv run pre-commit run --all-files` — all hooks passed, including Ruff, Ruff format, and mypy
- `uv run pytest -q` — 3,739 passed, 240 skipped

The branch changes four approved test/harness/documentation files (600 CLOC; 755 raw additions) and contains no production, dependency, lockfile, or workflow changes.

### DIFF
--- a/docs/compatibility.mdx
+++ b/docs/compatibility.mdx
@@ -32,9 +32,10 @@ uv run python scripts/mcp_wire_probe.py \
 ```
 
 The default probe does not invoke any Atlassian tool. It emits machine-readable
-JSON with the negotiated protocol version, installed MCP/FastMCP package
-versions, and discovered tool names, making it useful for comparing supported
-SDK and framework versions.
+JSON with the negotiated protocol version, discovered tool names, and installed
+versions of MCP, MCP types, FastMCP, FastMCP slim, and cryptography. Missing
+distributions are reported as `null`, making the output useful for comparing
+supported SDK and framework environments.
 
 Version-specific test lanes can make protocol negotiation an explicit gate:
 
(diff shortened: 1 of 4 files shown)


cblacknygc:
This is driven by prep for MCP v2 / FastMCP v4, but is useful for debug and testing for current release FastMCP v3 branch as well.

cblacknygc:
Converted to draft to add extra fastmcp 4 prep and regression fix mentioned in https://github.com/sooperset/mcp-atlassian/issues/1541 along with a round of negative tests to ensure write tools are not advertised when they should not be (due to read-only mode, ENABLED_TOOLS, TOOLSETS). Provide a mock fetcher for jira_get_issue positive testing.

PR #1568, opened 2026-08-07
fix(jira,confluence): validate attachment URLs against SSRF

### LINKED ISSUE #1567: [Security]: Jira attachment download/fetch skips validate_url_for_ssrf on initial request
## Summary

`AttachmentsMixin.download_attachment` and `AttachmentsMixin.fetch_attachment_content` in `src/mcp_atlassian/jira/attachments.py` call `self.jira._session.get(url, ...)` directly using the attachment's `url` (the `content` field from the Jira attachment API response), without first running it through `utils/urls.py::validate_url_for_ssrf`.

Relevant lines:
- `src/mcp_atlassian/jira/attachments.py:62` (`download_attachment`)
- `src/mcp_atlassian/jira/attachments.py:101` (`fetch_attachment_content`)

The outbound session does mount the DNS-pinning adapter and SSRF redirect hook, so *redirects* are validated/pinned. The *initial* request URL is not checked against `validate_url_for_ssrf` before the first connection.

## Why this matters

Other user/API-controlled URL entry points in the codebase (e.g. header-based Jira/Confluence URLs in `UserTokenMiddleware`) are explicitly validated with `validate_url_for_ssrf` before use. This attachment path is inconsistent with that pattern.

## Exploitability

Not currently confirmed as reachable by an untrusted actor through normal usage — `attachment.url` is server-computed by Jira's own API (`JiraAttachment.from_api_response`), not user-editable through normal issue/attachment creation flows. Reachable only if an attacker can otherwise tamper with the Jira API response content (e.g. compromised/malicious Jira instance, MITM'd API). Flagging as defense-in-depth / consistency gap rather than a confirmed exploit.

## Suggested fix

Call `validate_url_for_ssrf(url)` before the `self.jira._session.get(url, ...)` calls in both `download_attachment` and `fetch_attachment_content`, matching the pattern used elsewhere for externally-sourced URLs.

brennanneoh:
## Summary
- `download_attachment`/`fetch_attachment_content` in both Jira and Confluence attachment mixins hit the attachment `url` directly with `_session.get(...)`, without running it through `validate_url_for_ssrf` first — only redirects were pinned/validated.
- Adds the same `validate_url_for_ssrf` check used elsewhere (e.g. header-based Jira/Confluence URLs) before the initial request in all four affected methods.

Closes #1567

## Test plan
- [x] Added regression tests asserting download/fetch is refused (and the session's `get` is never called) when `validate_url_for_ssrf` blocks the URL, for both Jira and Confluence.
- [x] `uv run pytest tests/unit/jira/test_attachments.py tests/unit/confluence/test_attachments.py` — 129 passed.
- [x] `prek run` on all touched files — ruff, ruff-format, mypy all pass.

### DIFF
--- a/src/mcp_atlassian/confluence/attachments.py
+++ b/src/mcp_atlassian/confluence/attachments.py
@@ -8,7 +8,7 @@ from typing import Any
 
 from ..models.confluence import ConfluenceAttachment
 from ..utils.io import validate_safe_path
-from ..utils.urls import resolve_relative_url
+from ..utils.urls import resolve_relative_url, validate_url_for_ssrf
 from .client import ConfluenceClient
 from .protocols import AttachmentsOperationsProto
 from .v2_adapter import ConfluenceV2Adapter
@@ -348,6 +348,11 @@ class AttachmentsMixin(ConfluenceClient, AttachmentsOperationsProto):
             logger.error("No URL provided for attachment fetch")
             return None
 
+        ssrf_error = validate_url_for_ssrf(url)
+        if ssrf_error:
+            logger.error(f"Refusing to fetch attachment: {ssrf_error}")
+            return None
+
         try:
             logger.info(f"Fetching attachment from {url}")
             response = self.confluence._session.get(url, stream=True)
@@ -380,6 +385,11 @@ class AttachmentsMixin(ConfluenceClient, AttachmentsOperationsProto):
         """
         if not url:
             logger.error("No URL provided for attachment download")
+            return False
+
+        ssrf_error = validate_url_for_ssrf(url)
+        if ssrf_error:
+            logger.error(f"Refusing to download attachment: {ssrf_error}")
             return False
 
         try:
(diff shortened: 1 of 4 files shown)


ebarkhordar (commented):
I ran this against ab1b655 in a clean python:3.11-slim container with an on-prem style config: `JIRA_URL=https://jira.corp.internal`, that host resolving to 10.20.30.40, `MCP_ALLOWED_URL_DOMAINS` unset.

`validate_url_for_ssrf` has no operator-trusted-host exemption, so on a Server/DC instance that lives on a private network the four new guards (jira/attachments.py:38 and :105, confluence/attachments.py:351 and :390) refuse every attachment fetch and download:

```
ab1b655  Refusing to fetch attachment: DNS for jira.corp.internal
         resolves to non-global IP: 10.20.30.40
56b8b93  (main) request proceeds to the session
```

The sessions these methods use already mount `SsrfPinningAdapter` (jira/client.py:191, confluence/client.py:178), which on main refuses the initial request as well, at connect time and without the validate-then-reconnect window: 169.254.169.254 and 10.20.30.40 both come back `SSRF blocked: non-global address`, while the configured host is allowed. That exemption is deliberate, b041733 (#1448) added it as "on-prem DC instances legitimately live on private networks or localhost".

Skipping the new check when the URL host is operator-trusted keeps both properties. Patched in at ab1b655, the internal host reaches the session again and 169.254.169.254 stays refused. `_operator_trusted_hosts` and `_hostname_matches_allowlist` are private, so it likely wants a small public helper.

I did not test the OAuth or per-user fetcher paths.

Issue #1311, opened 2026-04-30
[Bug]: jira_add_comment strips [~username] mentions on Server/DC


pavel-kalmykov:
### Prerequisites

- [x] Searched existing issues
- [x] Checked README

### Bug Description

On JIRA Server/DC, passing `[~username]` in the markdown body of `jira_add_comment` (or `jira_edit_comment`) does not produce a mention. The brackets get stripped during the markdown-to-wiki conversion, and the tilde sometimes too. No notification fires and the rendered comment shows plain text.

Distinct from #1208, which covers the same UX problem on Cloud via the Markdown to ADF path. This issue is on Server/DC via the Markdown to wiki path.

### Steps to Reproduce

1. JIRA Server/DC, authenticated user with mention permissions
2. Call `jira_add_comment` with two mentions in the same paragraph:

```
ending sentence. [~user.one] [~User.Two] could you take a look?
```

3. Read back the comment via REST or check it in the JIRA UI

### Actual

The stored body becomes:

```
ending sentence.user.one ~User.Two could you take a look?
```

First mention: brackets and tilde both stripped, no mention link, no notification.

Second mention: brackets stripped but tilde kept.

Trying to escape with backslashes gives literal backslashes:

Input: `\[~user.one\] \[~User.Two\]`
Stored: `\user.one\ \User.Two\`

Both bracket pairs and both tildes are gone.

### Expected

`[~username]` should pass through the converter unchanged so JIRA wiki markup renders it as a mention link and fires a notification.

### Workaround

Edit the comment via the JIRA REST API directly (`PUT /rest/api/2/issue/KEY/comment/ID`) with the body in raw wiki markup. `[~username]` is preserved verbatim and renders correctly. `jira_add_comment` is only safe for comments without mentions.

### Environment

- mcp-atlassian: latest via `uvx mcp-atlassian`
- JIRA Server/DC

### Notes

The first/second mention asymmetry suggests the Markdown parser treats `[~name]` as a shortcut reference link, and parser state changes after the first one. Worth checking how `markdown_to_jira` handles bracket pairs followed by tilde.

Related: #1208 (Cloud ADF), #842 and #1297 (Confluence mentions).

newhoggy:
You may be interested in this: https://www.reddit.com/r/mcp/comments/1tam8tz/omnidev_atlassian_jiraconfluence_mcp_server_with/

gkintc:
[...]

PR #1570, opened 2026-08-10
fix(jira): only unwrap bare links in jira_to_markdown, not all bracketed text

### LINKED ISSUE #1566: [Bug]: jira_to_markdown strips brackets from all bracketed text, not just bare links (read path)
### Prerequisites

- [x] Searched existing issues
- [x] Checked README

### Bug Description

`jira_to_markdown` strips the square brackets from **any** bracketed run of text when converting Jira wiki markup to Markdown. This is a read-path defect: the stored Jira content is correct, but every MCP read of a description or comment silently rewrites it.

The cause is the second of the two link substitutions in `JiraPreprocessor.jira_to_markdown` (`src/mcp_atlassian/preprocessing/jira.py`, v0.23.0 line 355):

```python
# Links
output = re.sub(r"\[([^|]+)\|(.+?)\]", r"[\1](\2)", output)   # [text|url] -> [text](url)   OK
output = re.sub(r"\[(.+?)\]([^\(])", r"\1\2", output)         # <-- unwraps ANY [x], not just bare links
```

The second line looks intended to unwrap a *bare link* (`[http://example.com]` → `http://example.com`, which Jira renders as a link). But `(.+?)` matches any content, so it also unwraps ordinary bracketed text that carries meaning.

`[~username]` mentions are one instance of this — that symptom is already reported in #1311, where the analysis points at the markdown→wiki (write) direction. This issue is the general case and locates it in the opposite direction: **the write path is not involved.**

### Steps to Reproduce

```python
from mcp_atlassian.preprocessing.jira import JiraPreprocessor
p = JiraPreprocessor(base_url="https://jira.example.com")

p.jira_to_markdown("Emailed [~jdoe] to update the owner.")
p.jira_to_markdown("engagements ([eBay], [Netflix], etc.)")
p.jira_to_markdown("Cost is [redacted] for now")
```

Or over MCP against a Server/DC instance: call `jira_get_issue` with `comment_limit` set on any issue whose comment contains `[~someone]`, and compare against `GET /rest/api/2/issue/<KEY>/comment/<ID>`.

### Actual

| Input (as stored in Jira) | `jira_to_markdown` output |
|---|---|
| `Emailed [~jdoe] to update the owner.` | `Emailed ~jdoe to update the owner.` |
| `engagements ([eBay], [Netflix], etc.)` | `engagements (eBay, Netflix, etc.)` |
| `Cost is [redacted] for now` | `Cost is redacted for now` |

Confirmed end to end on a Jira Server/DC instance (usernames replaced with placeholders here). A comment created in December 2024 and never edited since is stored as `Emailed [~jdoe] to update the product owner.` (verified via `GET /rest/api/2/issue/<KEY>/comment/<ID>`) but is returned by `jira_get_issue` as `Emailed <~jdoe> to update the product owner.` — brackets replaced, so it round-trips to a different string than Jira holds.

### Expected

Bracketed text that is not a link should survive the conversion unchanged. `[~username]` should stay `[~username]`; `[eBay]` should stay `[eBay]`.

Unwrapping should be limited to brackets that actually contain a bare link — roughly, require the content to look like a URI or mail target:

```python
output = re.sub(r"\[((?:[A-Za-z][A-Za-z0-9+.-]*:|mailto:|www\.)[^\]|]+)\]([^\(])", r"\1\2", output)
```

Anything else (`[~user]`, `[eBay]`, `[redacted]`, `[#anchor]`) then passes through untouched. Wiki links with an explicit target (`[text|url]`) are handled by the preceding substitution and are unaffected.

### What is NOT the cause

Worth stating explicitly, since #1311 points the other way. The **write** path preserves brackets correctly:

- `markdown_to_jira("engagements ([eBay], [Netflix], etc.)")` returns the string unchanged. Checked on 0.11.9, 0.13.0, 0.15.0, 0.17.0, 0.19.0, 0.21.1 and 0.23.0 — no version strips brackets on write.
- The Cloud path `markdown_to_adf(...)` also preserves them; the text nodes contain `[eBay]` and `[Netflix]` verbatim.

So a user who sees a bracketed tag "missing" after posting through `jira_add_comment` is most likely looking at a mangled read-back of an intact comment. That distinction matters: the natural response is to "repair" the comment by rewriting it from the MCP output, which replaces good stored content with the corrupted version and causes the real data loss.

### Side note: same-direction underscore mangling

In the same read direction, `AWS_Essential_Services` → `AWS*Essential*Services` via the bold/italic substitution at line ~275 (`re.sub(r"([*_])(.*?)\1", ...)`), which pairs the two underscores as emphasis delimiters. Cosmetic rather than content-losing, and possibly worth a separate issue, but it has the same shape: a markup rule applied to text that was never markup. Happy to file separately if preferred.

### Environment

- mcp-atlassian: reproduced against 0.23.0 (latest at time of filing); write-path check swept 0.11.9 → 0.23.0
- Jira Server/DC (live confirmation), plus direct unit-level reproduction of the converter
- Python 3.12

### Related

- #1311 — same root cause, reported as `[~username]` mentions being stripped, attributed there to the write path

Julien-ser:
Closes #1566.

## Problem

The second link substitution in `JiraPreprocessor.jira_to_markdown` (`src/mcp_atlassian/preprocessing/jira.py`) used an unanchored pattern:

```python
output = re.sub(r"\[(.+?)\]([^\(])", r"\1\2", output)
```

It was intended to unwrap a *bare link* (`[https://example.com]` → `https://example.com`, matching how Jira renders it), but `(.+?)` matches any content, so it also stripped the brackets from ordinary text that carries meaning.

This is a **read-path** defect: the stored Jira content is correct, but every MCP read of a description or comment silently rewrote it.

| Input (as stored in Jira) | Before | After |
|---|---|---|
| `Emailed [~jdoe] to update the owner.` | `Emailed ~jdoe to update the owner.` | unchanged ✅ |
| `engagements ([eBay], [Netflix], etc.)` | `engagements (eBay, Netflix, etc.)` | unchanged ✅ |
| `Cost is [redacted] for now` | `Cost is redacted for now` | unchanged ✅ |
| `See [http://example.com] for details` | `See http://example.com for details` | same ✅ |
| `Link [text\|http://example.com] here` | `Link [text](http://example.com) here` | same ✅ |

## Fix

Anchor the pattern on a URI scheme so only real link targets are unwrapped, and use a negative lookahead instead of consuming a trailing character so already-converted `[text](url)` output from the preceding substitution stays intact.

The lookahead also fixes bare links at end-of-string, which the old trailing-character match structurally could not handle (`Bare at end [https://example.com]` was previously left wrapped).

Note: #1311 reports the `[~username]` symptom and analyses it in the markdown→wiki (write) direction. This PR is the general case in the opposite direction; the write path is not touched.

## Testing

Added 8 parametrized regression tests: 4 pinning that non-link brackets survive, 4 pinning that real bare links still unwrap and labelled links still convert. Verified they **fail against the pre-fix code** (4 failures) and pass after.

- `tests/unit/preprocessing/`: 175 passed
- Full `tests/unit/`: 3743 passed. The 4 unrelated failures (`test_upload_attachment_success`, `test_parse_date_timestamp_boundary_max_valid`, `test_mime_type_detection[zip]`, `test_saved_token_file_is_not_group_or_world_readable`) reproduce identically on unmodified `main` on this Windows machine and are environment-specific, not caused by this change.
- `ruff check` (with the repo's pre-commit ignore list) and `ruff format --check`: clean.

### DIFF
--- a/src/mcp_atlassian/preprocessing/jira.py
+++ b/src/mcp_atlassian/preprocessing/jira.py
@@ -355,7 +355,17 @@ class JiraPreprocessor(BasePreprocessor):
 
         # Links
         output = re.sub(r"\[([^|]+)\|(.+?)\]", r"[\1](\2)", output)
-        output = re.sub(r"\[(.+?)\]([^\(])", r"\1\2", output)
+        # Bare links: [https://example.com] -> https://example.com. Anchored on a
+        # URI scheme so only actual link targets are unwrapped; an unanchored
+        # `\[(.+?)\]` also strips ordinary bracketed text that carries meaning
+        # ([~user] mentions, [eBay], [redacted], ...) — see #1566. The negative
+        # lookahead keeps already-converted `[text](url)` output from the line
+        # above intact.
+        output = re.sub(
+            r"\[([a-zA-Z][a-zA-Z0-9+.-]*://[^\]\s]+)\](?!\()",
+            r"\1",
+            output,
+        )
 
         # Colored text
         output = re.sub(
--- a/tests/unit/preprocessing/test_preprocessing.py
+++ b/tests/unit/preprocessing/test_preprocessing.py
@@ -1479,6 +1479,61 @@ class TestPanelBlocks:
         result = preprocessor.jira_to_markdown("[https://example.com] more text")
         assert "https://example.com" in result, f"URL dropped: {result}"
 
+    @pytest.mark.parametrize(
+        "test_id,input_text",
+        [
+            ("user_mention", "Emailed [~jdoe] to update the owner."),
+            ("bracketed_names", "engagements ([eBay], [Netflix], etc.)"),
+            ("redaction_marker", "Cost is [redacted] for now"),
+            ("bracketed_text_at_end", "Trailing [redacted]"),
+        ],
+    )
+    def test_non_link_brackets_are_preserved(
+        self, preprocessor, test_id: str, input_text: str
+    ):
+        """Bracketed text that is not a link must survive the read path (#1566).
+
+        The bare-link unwrap used to match any `[...]`, so ordinary bracketed
+        content ([~user] mentions, product names, redaction markers) silently
+        lost its brackets on every read of a description or comment.
+        """
+        assert preprocessor.jira_to_markdown(input_text) == input_text, (
+            f"[{test_id}] brackets stripped from non-link text"
+        )
+
+    @pytest.mark.parametrize(
+        "test_id,input_text,expected",
+        [
+            (
+                "http",
+                "See [http://example.com] for details",
+                "See http://example.com for details",
+            ),
+            (
+                "https",
+                "[https://example.com] more text",
+                "https://example.com more text",
+            ),
+            (
+                "at_end",
+                "Bare at end [https://example.com]",
+                "Bare at end https://example.com",
+            ),
+            (
+                "labelled_link_untouched",
+                "Link [text|http://example.com] here",
+                "Link [text](http://example.com) here",
+            ),
+        ],
+    )
+    def test_bare_links_still_unwrapped(
+        self, preprocessor, test_id: str, input_text: str, expected: str
+    ):
+        """Real bare links must still unwrap, and labelled links still convert (#1566)."""
+        assert preprocessor.jira_to_markdown(input_text) == expected, (
+            f"[{test_id}] link conversion regressed"
+        )
+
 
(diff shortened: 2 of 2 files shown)


Julien-ser:
Rebased onto current `main` (no conflicts). 199 preprocessing tests pass, ruff clean. #1566 is still open with no other PR referencing it — happy to address any remaining review feedback.

PR #1591, opened 2026-08-17
fix(confluence): report property delete failures instead of success

### LINKED ISSUE #1557: [Bug]: property delete failures return True; update_page discards the result, so emoji/width removal reports success it never had
### Prerequisites

- [x] I have searched the existing issues to make sure this bug has not already been reported.
- [x] I have checked the README for relevant information.

### Bug Description

`_set_single_property()` returns `True` when a property **delete** fails for any reason at all, and the two callers that consume it throw the result away anyway. So "remove the page emoji" / "reset the page width" can report success while nothing was removed.

Two separate faults stacked on top of each other:

**1. `src/mcp_atlassian/confluence/pages.py:362-369` — the delete branch swallows everything.**

```python
if value is None:
    # Delete the property
    try:
        self.confluence.delete_page_property(page_id, property_key)
    except Exception as e:
        # Property might not exist, which is fine
        logger.debug(f"Could not delete property '{property_key}': {e}")
    return True
```

The comment is right about one case and wrong about the rest. A 404 genuinely is fine. A 401/403 (token lost the scope), a 423 (page locked), a 429, or a dropped connection are all caught by the same bare `except` and land on the same `return True`. *The property does not exist* and *I was not permitted to remove it* come out as the identical value, and the only trace is a `logger.debug` line most people never have on.

**2. `pages.py:871` and `:877` — `update_page()` discards the bool.**

```python
if emoji is not None:
    emoji_to_set = emoji if emoji else None
    self._set_page_emoji(page_id, emoji_to_set)      # <- return value dropped

if page_width is not None:
    width_to_set = page_width if page_width else None
    self._set_page_width(page_id, width_to_set)      # <- return value dropped

return self.get_page_content(page_id)                # <- :880, unconditional success
```

`create_page()` drops the return the same way at `:764` / `:768`, though its guard is truthiness (`if emoji:`) rather than `is not None`, so the empty-string remove path never reaches it there at all.

So even where `_set_single_property` **does** correctly return `False`, nothing upstream looks. `update_page` returns a fully-populated page object either way, and an agent calling `confluence_update_page` gets a success payload and tells its user the emoji was removed.

`_set_page_emoji` itself is careful — it checks `published_ok and draft_ok` and logs warnings — so the fix at the call sites is small and the intent is clearly already there.

### Steps to Reproduce

1. Set an emoji on a page: `confluence_update_page(page_id=..., emoji="🎯")`. Confirm `emoji-title-published` exists.
2. Make the delete fail without making the *page* update fail — easiest reliably is a token whose scope covers content write but not property delete; a locked page or a mid-call 429 does the same thing.
3. `confluence_update_page(page_id=..., emoji="")` — empty string is the documented "remove it" path.
4. Tool returns the page object, no error, no warning above `debug`.
5. `GET /wiki/rest/api/content/<id>/property/emoji-title-published` — still there.

### Expected Behavior

Distinguish *nothing to delete* from *delete refused*. Narrowly: catch the 404 (or whatever `atlassian-python-api` raises for a missing property) and return `True` for that only; let everything else fall through to the existing `except` at `:396` which already logs a warning and returns `False`. Then bind the return at `:871`/`:877`/`:764`/`:768` and surface a partial-success rather than an unconditional page payload.

### Honest scope

This is page **emoji** and **width**, not page content — nobody loses a document to it. I am filing it because the swallow lives in a generic property primitive rather than in the emoji feature, and because there are already several open issues in this repo about writes that report success and do not land (#1547, #1500, #1189, #1230). This is a sibling of those, not the cause of them — I have **not** verified it explains #1547, and I do not think it does; that one is a property being *written* when it shouldn't be, which is the other direction.

Found by reading `pages.py` around #1547. Happy to open a PR for the narrowed exception plus the four call sites if that shape is wanted.

---

Context, since it is relevant to why I was in there: this failure class is what I do — writes that return cleanly while the world does not change. If it is ever useful to have someone go through your action layer looking for the rest of them, that is at unreached.dev. Either way this issue stands on its own and needs nothing from me.

arose26:
Closes #1557

### Change

`_set_single_property()` now treats only a 404 property-delete response as an already-satisfied removal. Permission, lock, throttling, and transport failures return `False`.

`update_page()` now checks the results of `_set_page_emoji()` and `_set_page_width()`. If either metadata update fails, it raises an explicit `_PagePropertyUpdateError` and does not return a refreshed `ConfluencePage`, so callers cannot receive a false success.

### Tests

- `uv run pytest tests/unit -q`: 3821 passed, 5 skipped
- `uv run pytest tests/unit/confluence/test_pages.py -xvs`: 130 passed
- Cloud property-removal E2E: 1 passed, 19 deselected, no skips
- Data Center property-removal E2E: 1 passed, 16 deselected, no skips
- `uv run ruff format --check` and `uv run python scripts/generate_tool_docs.py --check`: passed

The focused unit tests cover failed emoji removal and failed width reset, including the explicit error and the absence of a normal page refresh.

### DIFF
--- a/src/mcp_atlassian/confluence/pages.py
+++ b/src/mcp_atlassian/confluence/pages.py
@@ -6,6 +6,7 @@ from typing import Any
 from urllib.parse import parse_qs, urlparse
 
 import requests
+from atlassian.errors import ApiError
 from bs4 import BeautifulSoup, Tag
 from requests.exceptions import HTTPError
 
@@ -366,9 +367,13 @@ class PagesMixin(ConfluenceClient):
                 # the removal did not happen and must not be reported as done.
                 try:
                     self.confluence.delete_page_property(page_id, property_key)
-                except HTTPError as http_err:
+                except (HTTPError, ApiError) as api_err:
+                    http_err = (
+                        api_err if isinstance(api_err, HTTPError) else api_err.reason
+                    )
                     if (
-                        http_err.response is not None
+                        isinstance(http_err, HTTPError)
+                        and http_err.response is not None
                         and http_err.response.status_code == 404
                     ):
                         logger.debug(
--- a/src/mcp_atlassian/confluence/pages.py
+++ b/src/mcp_atlassian/confluence/pages.py
@@ -899,16 +899,24 @@ class PagesMixin(ConfluenceClient):
             if emoji is not None:
                 # Empty string means remove emoji, otherwise set it
                 emoji_to_set = emoji if emoji else None
-                if not self._set_page_emoji(page_id, emoji_to_set):
-                    error_message = f"Failed to set page emoji for page {page_id}"
+                emoji_updated = self._set_page_emoji(page_id, emoji_to_set)
+                if not emoji_updated:
+                    error_message = (
+                        f"Page content was updated, but page emoji update failed "
+                        f"for page {page_id}"
+                    )
                     raise _PagePropertyUpdateError(error_message)
 
             # Set or remove the page width if provided
             if page_width is not None:
                 # Empty string means reset to default, otherwise set it
                 width_to_set = page_width if page_width else None
-                if not self._set_page_width(page_id, width_to_set):
-                    error_message = f"Failed to set page width for page {page_id}"
+                width_updated = self._set_page_width(page_id, width_to_set)
+                if not width_updated:
+                    error_message = (
+                        f"Page content was updated, but page width update failed "
+                        f"for page {page_id}"
+                    )
                     raise _PagePropertyUpdateError(error_message)
 
             # After update, refresh the page data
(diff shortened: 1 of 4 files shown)


arose26:
Pushed as `7323c64`. `create_page()` now checks the return value of both `_set_page_emoji()`
and `_set_page_width()` and raises `RuntimeError` naming what did and did not happen —
"Page was created, but page emoji update failed for page {page_id}" — so it reads the same way
as the handling already in `update_page()`, and `get_page_content()` is no longer reached on a
rejected property.

Two regressions, next to the existing `update_page` ones and following their shape: one per
helper, each asserting the exact caller-facing message, that the helper was called with the
expected arguments, and that `get_page_content` was **not** called. Reverting `pages.py` alone
fails both.

`tests/unit/confluence/` is 513 passed. `ruff check` on the two touched files reports 53 errors
both before and after — all pre-existing in untouched code, count unchanged.

PR #1597, opened 2026-08-18
Fall back to /rest/api/content/search when primary search returns non-JSON

### LINKED ISSUE #1596: [Bug]: confluence_search returns raw HTML on Confluence Cloud instances where /rest/api/search returns 400
### Prerequisites

### Bug Description

`confluence_search` returns raw HTML instead of search results on some Confluence Cloud instances. The root cause is that the underlying `atlassian-python-api` `cql()` method calls `/rest/api/search`, which on some Confluence Cloud instances returns HTTP 400 with an HTML body (a login/error page) instead of a JSON error for both `siteSearch ~` and `text ~` queries.

The siteSearch fallback logic (line 265-271 in `servers/confluence.py`) catches the initial `siteSearch` failure and retries with `text ~`, but both fail because the issue is with the `/rest/api/search` endpoint itself, not the CQL operator.

### Root Cause

On the affected instance (Confluence Cloud):

- `GET /rest/api/search?cql=siteSearch ~ "foo"` returns **400 + HTML**
- `GET /rest/api/search?cql=text ~ "foo"` returns **400 + HTML**
- `GET /rest/api/content/search?cql=type=page AND text ~ "foo"` returns **200 + JSON** (works correctly)

The `atlassian-python-api` library `cql()` method uses the `/rest/api/search` endpoint. The `/rest/api/content/search` endpoint works but is not used by the MCP tool.

Other Confluence tools (`get_page`, `get_page_children`, etc.) work fine because they use different API endpoints.

### Steps to Reproduce

1. Configure `mcp-atlassian` against a Confluence Cloud instance where `/rest/api/search` returns 400 for CQL queries
2. Call `confluence_search` with any query (simple text or CQL with `text ~`)
3. Observe raw HTML returned as the error message

### Expected Behavior

Search should return results or a clear error message, not raw HTML.

### Suggested Fix

When `cql()` returns a non-dict response (e.g. an HTML string), detect this early and either:
1. Fall back to `/rest/api/content/search` endpoint (which works on affected instances)
2. Detect the HTML content-type in the response and raise a clear error ("Confluence search endpoint returned HTML instead of JSON, the /rest/api/search endpoint may not be available on this instance")

### Environment

- mcp-atlassian version: 0.23.0
- Confluence: Cloud
- Python: via uvx

costajohnt:
Fixes #1596

## Problem

Some Confluence Cloud instances return HTTP 400 with an HTML body from `/rest/api/search` for valid CQL queries. The `/rest/api/content/search` endpoint uses a different backend and works on the affected instances.

`atlassian-python-api` 4.0.7 does not return that HTML body directly: `Confluence.cql()` catches the internal `HTTPError` and raises `ApiValueError("The query cannot be parsed")`, retaining the original HTTP exception in `reason`. The fallback therefore needs to inspect that exception chain before the normal search response validation runs.

## Fix

- On Confluence Cloud, detect an `ApiValueError` whose internal `HTTPError` contains a 400 HTML response and retry through `/rest/api/content/search`.
- Retain support for older client behavior that returns the raw HTML body directly.
- Normalize the flat content-search results to the shape expected by the existing search-result parser.
- Keep ordinary JSON HTTP 400 responses for invalid CQL on the existing error path rather than masking them with a fallback.

## Testing

- `uv run pytest tests/unit/confluence/test_search.py -q` — 71 passed
- `uv run pytest tests/unit/confluence/ -q` — 532 passed
- Focused pinned-client regression cases — 2 passed
- Targeted Ruff format/lint and mypy checks — passed
- `uv run python scripts/generate_tool_docs.py --check` — passed
- Cloud e2e was invoked, but the supplied basic credentials failed the suite health check; all 111 tests skipped and no test resource was created. A read-only OAuth search smoke test also reached Atlassian but returned HTTP 404.
- DC e2e was invoked because this is a shared Confluence client module, but the local Jira/DC fixture was unavailable; all DC tests skipped and no test resource was created.

Regression coverage uses the real `ApiValueError` and nested `requests.HTTPError` structure emitted by client version 4.0.7, and separately verifies that a JSON invalid-CQL 400 does not trigger fallback.

### DIFF
--- a/src/mcp_atlassian/confluence/search.py
+++ b/src/mcp_atlassian/confluence/search.py
@@ -123,6 +123,38 @@ class SearchMixin(ConfluenceClient):
             cql=cql, limit=limit, expand="content.history,content.version"
         )
 
+        # Some Confluence Cloud instances return HTTP 400 with an HTML body
+        # from /rest/api/search instead of a JSON error. When this happens
+        # the library returns the raw HTML string rather than a dict. Fall
+        # back to the /rest/api/content/search endpoint which uses a
+        # different backend and works on affected instances.
+        if not isinstance(results, dict) or "results" not in results:
+            logger.warning(
+                "Primary search endpoint (/rest/api/search) returned a "
+                "non-JSON response; falling back to /rest/api/content/search"
+            )
+            results = self.confluence.get(
+                "rest/api/content/search",
+                params={
+                    "cql": cql,
+                    "limit": limit,
+                    "expand": "history,version",
+                },
+            )
+            # /rest/api/content/search returns results with id, title,
+            # space etc. at the top level, while /rest/api/search nests
+            # them under a "content" key. Normalize to the /rest/api/search
+            # format so the downstream model parser works unchanged.
+            if isinstance(results, dict) and isinstance(
+                results.get("results"), list
+            ):
+                results["results"] = [
+                    {"content": item, "excerpt": item.get("excerpt", "")}
+                    if "content" not in item
+                    else item
+                    for item in results["results"]
+                ]
+
         # Surface malformed responses (missing "results") as errors while
         # allowing a genuine "results": [] to return an empty list.
         self._validate_search_response(results, "search")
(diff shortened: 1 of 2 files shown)


costajohnt:
Checking in on this one. Happy to rebase or take a different approach to the fallback if you'd prefer something else.

PR #1603, opened 2026-08-19
fix(jira): filter Cloud-only tools on Server/Data Center

### LINKED ISSUE #1082: [Bug]: Cloud-Only Tools Exposed to Agents When Connected to Server/Data Center
### Prerequisites

### Bug Description

When `JIRA_URL` is configured with a Jira Server/Data Center instance, the MCP server registers and exposes Cloud-only tools (e.g. `batch_get_changelogs`) to agents. These tools fail at call time instead of being excluded at registration time.

### Steps to Reproduce

1. Configure `JIRA_URL` with a Jira Server/Data Center URL (non-Atlassian Cloud URL).
2. Start the MCP server.
3. Let an agent discover available tools.
4. Agent attempts to call `batch_get_changelogs` (or any other Cloud-only tool).

### Expected Behavior

Tools that are only supported on Jira Cloud should not be registered or exposed when the server is configured with a Jira Server/Data Center URL. The deployment type is known at startup time and should be used to filter available tools during registration.

### Actual Behavior

```shell
Error calling tool 'batch_get_changelogs': Batch get issue changelogs is only available on Jira Cloud.
```

### mcp-atlassian Version

v0.18.1

### Installation Method

Docker

### Operating System

Windows

### Python Version

N/A

### Atlassian Instance Type

Jira Server / Data Center

### Client Application

VS Code

### Additional Context

Image

FullYolk:
## Description

Hide Jira tools backed exclusively by Cloud APIs from MCP discovery when the active Jira deployment is Server or Data Center. Apply the same capability check at dispatch time so a client cannot bypass discovery filtering by calling a hidden tool directly.

Fixes: #1082

Thanks to @AmirF194 for validating the behavior through the production server registration path and documenting the root cause in #1082.

## Changes

- Tag `jira_batch_get_changelogs` and `jira_move_issue` as Cloud-only.
- Derive Jira deployment capability from the startup `JiraConfig` or per-request Jira URL headers.
- Filter deployment-incompatible tools during discovery and reject direct calls before execution.
- Add regression coverage for Cloud and Server/Data Center discovery, header-based deployments, and direct-call defense.

## Testing

- [ ] Integration tests passed (no live Jira instance required for this change)
- [x] Manual checks performed: verified the production `main_mcp` registry lists both tools for Cloud and neither for Server/Data Center

## Checklist

- [ ] All tests pass locally (3,830 passed; 4 unrelated Windows-specific tests fail on path, timestamp, MIME, and POSIX permission behavior).
- [x] Documentation updated (no generated content changed; the documentation consistency check passes).

### DIFF
--- a/src/mcp_atlassian/servers/jira.py
+++ b/src/mcp_atlassian/servers/jira.py
@@ -1899,7 +1899,7 @@ async def batch_create_issues(
 
 
 @jira_mcp.tool(
-    tags={"jira", "read", "toolset:jira_issues"},
+    tags={"jira", "read", "cloud_only", "toolset:jira_issues"},
     annotations={"title": "Batch Get Changelogs", "readOnlyHint": True},
 )
 async def batch_get_changelogs(
@@ -2385,7 +2385,7 @@ async def delete_issue(
 
 
 @jira_mcp.tool(
-    tags={"jira", "write", "toolset:jira_issues"},
+    tags={"jira", "write", "cloud_only", "toolset:jira_issues"},
     annotations={"title": "Move Issue to Project", "destructiveHint": True},
 )
 @check_write_access
--- a/src/mcp_atlassian/servers/main.py
+++ b/src/mcp_atlassian/servers/main.py
@@ -259,11 +259,25 @@ class AtlassianMCP(ErrorPreservingFastMCP[MainAppContext]):
         )
 
         header_based_services = {"jira": False, "confluence": False}
+        service_headers: dict[str, str] = {}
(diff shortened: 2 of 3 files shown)


FullYolk:
@sooperset Thanks for the review and merge! I’m glad the production-registry coverage and direct-call guard matched what the project needed. I appreciate the thoughtful feedback.

AmirF194:
Nice fix, and thanks for building on the diagnosis and adding the dispatch-time check too, that closes the discovery-bypass gap the original report didn't cover.

Issue #1610, opened 2026-08-21
jira_to_markdown pairs escaped emphasis delimiters, rewriting stored \_ as \* on read-back


barrygfox:
## Bug Description

On the **read path**, `JiraPreprocessor.jira_to_markdown` mangles backslash-escaped characters that its write path (`markdown_to_jira`) produced. Stored wiki text containing escaped intraword underscores comes back through every MCP read with the escapes rewritten as asterisks — silently corrupting identifiers, config-variable names, and paths in tool output. e.g. JIRA\_TOKEN --> JIRA\*TOKEN

This is the read-path counterpart of #1361: that PR correctly made `markdown_to_jira` store intraword underscores as `\_` so Jira's renderer would not italicise them, but nothing was added to `jira_to_markdown` to handle those escapes. It is also a second instance of the defect class reported in #1566: *stored content is correct; every MCP read silently rewrites it*.

### Evidence (Jira Data Center)

For a description stored (byte-exact via REST) as:

```
QUALITY\_GATES\_LLM\_ENABLED
```

- Jira's own rendering (`expand=renderedFields`): `QUALITY_GATES_LLM_ENABLED` — correct.
- MCP read-back (`jira_get_issue`): `QUALITY\*GATES\*LLM\_ENABLED` — **corrupted**.

The first two escaped underscores are rewritten to `\*`; an unpaired third survives as `\_`.

### Minimal reproduction

```python
from mcp_atlassian.preprocessing.jira import JiraPreprocessor

print(JiraPreprocessor().jira_to_markdown(r"QUALITY\_GATES\_LLM\_ENABLED"))
# -> QUALITY\*GATES\*LLM\_ENABLED   (expected: QUALITY\_GATES\_LLM\_ENABLED)
```

## Root cause

Two defects combining in `JiraPreprocessor.jira_to_markdown` (`src/mcp_atlassian/preprocessing/jira.py`):

1. The read path never unescapes or otherwise accounts for `\X` escape sequences anywhere.
[...]

PR #1615, opened 2026-08-24
fix(jira): stop prose plus signs corrupting markdown read-back

barrygfox:
## Problem

Unpaired plus signs in Jira prose, especially version numbers such as `Python 3.13+` and `urllib3 2.6+`, could be paired across text by the inserted-text rule. That emitted an `<ins>` tag, triggered whole-document HTML conversion, escaped existing Markdown characters, and removed the plus signs from read responses.

The HTML guard also checked content only after `jira_to_markdown()`. Converter-generated tags therefore still looked like source HTML, so the guard did not prevent the second conversion.

## Solution

1. Inserted-text delimiters adjacent to word characters are treated as prose: `(?<!\w)\+([^+\n]+)\+(?!\w)`.
2. `clean_jira_text()` records whether the original Jira source contains HTML outside protected code spans before Jira markup conversion.
3. Whole-document HTML conversion runs only when HTML came from that original source. Tags emitted by Jira conversion remain valid inline HTML in the Markdown result.

## Changes

- `src/mcp_atlassian/preprocessing/jira.py`
  - Preserve unpaired version-number plus signs.
  - Detect source HTML without mistaking Jira or Markdown code spans for HTML.
  - Keep converter-generated `<ins>`, `<cite>`, `<sup>`, and `<sub>` tags from triggering a second conversion.
- `tests/unit/preprocessing/test_preprocessing.py`
  - Cover stray plus signs alongside Markdown emphasis.
  - Cover literal plus signs in smart-link URLs.
  - Prove generated HTML does not trigger conversion, including when literal HTML appears inside a code span.
  - Preserve genuine inserted-text markup.

## Verification

- `uv run pytest tests/unit/preprocessing/ -q`: 175 passed
- `uv run pytest tests/unit/ -q`: 3842 passed, 5 repository-standard skips
- `uv run pytest 'tests/e2e/test_jira_auth_matrix.py::TestJiraReadOperations::test_get_issue[basic]' --dc-e2e -vv`: 1 passed, no skips
- `uv run python scripts/generate_tool_docs.py --check`: passed
- Ruff check and format check on changed Jira/test files: passed
- mypy hook on changed Jira/test files: passed

A broader DC run reached 102 passed, 1 failed, and 1 skipped. The failure is an unrelated JSM fixture problem: the pre-seeded `JSMDC-2` issue returns 403 from the service-desk comment endpoint.

Related: #1610

### DIFF
--- a/src/mcp_atlassian/preprocessing/jira.py
+++ b/src/mcp_atlassian/preprocessing/jira.py
@@ -309,8 +309,9 @@ class JiraPreprocessor(BasePreprocessor):
             output,
         )
 
-        # Inserted text
-        output = re.sub(r"\+([^+]*)\+", r"<ins>\1</ins>", output)
+        # Inserted text. Delimiters glued to word characters are prose
+        # (e.g. version numbers "3.13+"), not markup.
+        output = re.sub(r"(?<!\w)\+([^+\n]+)\+(?!\w)", r"<ins>\1</ins>", output)
 
         # Superscript
         output = re.sub(r"\^([^^]*)\^", r"<sup>\1</sup>", output)
--- a/src/mcp_atlassian/preprocessing/jira.py
+++ b/src/mcp_atlassian/preprocessing/jira.py
@@ -24,6 +24,20 @@ def _convert_panel(params: str | None, content: str) -> str:
     return f"\n{content}\n"
 
 
+def _contains_html_outside_code(text: str) -> bool:
+    """Return whether source text contains HTML outside protected code spans."""
+    code_patterns = (
+        r"\{code(?::[a-z]+)?\}[\s\S]*?\{code\}",
+        r"\{noformat\}[\s\S]*?\{noformat\}",
+        r"\{\{[^}]+\}\}",
+        r"```[^\n]*\n[\s\S]*?\n```",
+        r"`[^`]+`",
+    )
+    for pattern in code_patterns:
+        text = re.sub(pattern, "", text, flags=re.MULTILINE)
+    return re.search(r"<[^>]+>", text) is not None
+
+
 class JiraPreprocessor(BasePreprocessor):
     """Handles text preprocessing for Jira content."""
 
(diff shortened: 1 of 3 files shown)


src/mcp_atlassian/preprocessing/jira.py
```diff
@@ -24,6 +24,20 @@ def _convert_panel(params: str | None, content: str) -> str:
...
+    code_patterns = (
+        r"\{code(?::[a-z]+)?\}[\s\S]*?\{code\}",
+        r"\{noformat\}[\s\S]*?\{noformat\}",
+        r"\{\{[^}]+\}\}",
+        r"```[^\n]*\n[\s\S]*?\n```",
+        r"`[^`]+`",
+    )
+    for pattern in code_patterns:
+        text = re.sub(pattern, "", text, flags=re.MULTILINE)
+    return re.search(r"<[^>]+>", text) is not None
```
AmirF194, line 38: I ran this against `1864655c2` in a clean container, importing just `preprocessing/jira.py` and `base.py`.

`_contains_html_outside_code` (jira.py:39) reuses the same naive `<[^>]+>` pattern that caused the bug this PR fixes, so it false-positives on any plain-text angle brackets in the source, like an email citation. That flips `source_had_html` to True and the whole-document `_convert_html_to_markdown` pass still runs, corrupting markup `jira_to_markdown` generated elsewhere in the same text, the exact failure this PR is meant to stop.

```
IN : 'Signed-off-by: Jane Doe <jane@example.com>\n\nBumped urllib3 to 2.6+'
OUT: 'Signed-off-by: Jane Doe \nBumped urllib3 to 2.6+'

IN : 'cc <ops-team@example.com> -- please check the +action item+ below'
OUT: 'cc  -- please check the action item below'
```

The email is dropped outright (BeautifulSoup treats `<jane@example.com>` as an unknown tag), and in the second case the genuine `+action item+` markup loses its `<ins>` wrapper too.

Scoping the source-HTML check to real tag shapes (something like `<[a-zA-Z/][^>]*>`) instead of any `<...>` span would probably fix both.

PR #1618, opened 2026-08-26
fix(jira): name the failing argument in update_issue parse errors

barrygfox:
## Problem

`jira_update_issue` parses both its `fields` and `additional_fields` arguments with the same helper (`_parse_additional_fields`), whose `ValueError` always names `additional_fields`. A caller that passes plain text (for example, Markdown) in `fields` is therefore told that the wrong argument failed:

```
Error calling tool 'update_issue': additional_fields is not valid JSON: Expecting value: line 1 column 1 (char 0)
```

The same misleading attribution affects the `fields` argument of `jira_transition_issue`.

## Solution

Parameterize the argument name used by the shared parser and pass `"fields"` from the update and transition call sites.

## Changes

- `_parse_additional_fields(..., param_name="additional_fields")` now uses the actual argument name in validation errors.
- `jira_update_issue` and `jira_transition_issue` identify invalid `fields` input correctly.
- Regression tests cover both tool paths.

## What Does Not Change

- Valid-input behavior is unchanged.
- Empty and whitespace-only strings remain validation errors.
- There are no tool-schema, API, or dependency changes.

## Verification

- [x] Focused Jira parser regressions: 4 passed
- [x] `uv run pytest tests/unit -q`: 3842 passed, 5 skipped
- [x] Ruff format and lint checks on the changed test file
- [x] `uv run python scripts/generate_tool_docs.py --check`

## Related

Fixes the server-side half of internal ticket AIDEV-482. See also #1466 and #1140 for prior work on `update_issue` ergonomics.

### DIFF
--- a/src/mcp_atlassian/servers/jira.py
+++ b/src/mcp_atlassian/servers/jira.py
@@ -160,11 +160,13 @@ def _parse_visibility(
 
 def _parse_additional_fields(
     additional_fields: dict[str, Any] | str | None,
+    param_name: str = "additional_fields",
 ) -> dict[str, Any]:
     """Parse additional_fields from dict or JSON string.
 
     Args:
         additional_fields: Dict, JSON string, or None.
+        param_name: Argument name used in error messages.
 
     Returns:
         Parsed dict of additional fields.
@@ -180,13 +182,11 @@ def _parse_additional_fields(
         try:
             parsed = json.loads(additional_fields)
             if not isinstance(parsed, dict):
-                raise ValueError(
-                    "Parsed additional_fields is not a JSON object (dict)."
-                )
+                raise ValueError(f"{param_name} is not a JSON object (dict).")
             return parsed
         except json.JSONDecodeError as e:
-            raise ValueError(f"additional_fields is not valid JSON: {e}") from e
-    raise ValueError("additional_fields must be a dictionary or JSON string.")
+            raise ValueError(f"{param_name} is not valid JSON: {e}") from e
+    raise ValueError(f"{param_name} must be a dictionary or JSON string.")
 
 
 def _parse_request_field_values(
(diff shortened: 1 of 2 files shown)


src/mcp_atlassian/servers/jira.py
```diff
@@ -160,11 +160,13 @@ def _parse_visibility(
 
 def _parse_additional_fields(
     additional_fields: dict[str, Any] | str | None,
+    param_name: str = "additional_fields",
```
ebarkhordar, line 163: I ran this against `9b33b7e` in a clean container (`python:3.12-slim`, `pip install -e .`). The two call sites you changed behave as described. There is a third caller of this parser that still inherits the default name, and it produces the doubled message rather than a corrected one.

`_parse_request_field_values` at `src/mcp_atlassian/servers/jira.py:197` calls `_parse_additional_fields(request_field_values)` with no `param_name`, then wraps the `ValueError` with its own prefix. That is the parse path for `jira_create_customer_request`'s `request_field_values` argument:

```
>>> _parse_request_field_values("plain markdown text, not JSON")
ValueError: request_field_values is not valid JSON: additional_fields is not valid JSON: Expecting value: line 1 column 1 (char 0)
```

Your own regression assertion is a good yardstick here. `"additional_fields" not in message` is `False` on that path, where it is `True` for the two you fixed.

Passing the name through makes the wrapper's prefix redundant, so line 197 becomes `_parse_additional_fields(request_field_values, param_name="request_field_values")` and the `except ValueError` re-raise around it can go:

```
>>> _parse_additional_fields("plain markdown text, not JSON", param_name="request_field_values")
ValueError: request_field_values is not valid JSON: Expecting value: line 1 column 1 (char 0)
```

I enumerated the callers with `ast` rather than grep, so for what it is worth the full set is five in `src/`: `create_issue:1810` and `update_issue:2137` both pass a genuine `additional_fields` and are correct on the default, your two new ones at 2124 and 3001, and this one at 197.

Up to you whether it belongs in this PR or a follow-up. I did not test against a live Jira instance, only the parser.

PR #1627, opened 2026-08-28
fix(confluence): preserve inline comment replies

huyusong10:
## Description

`confluence_reply_to_comment` currently routes Cloud OAuth/PAT replies through the footer-comment endpoint and does not preserve an inline thread's location on Server/Data Center. Replies to inline review comments can therefore be stored as footer comments even when their parent link is valid.

This PR resolves the thread type before creating a reply and follows legacy reply ancestry back to the thread root. It does not change the MCP tool signature.

Independently mergeable. Complementary read-side retrieval and metadata enhancement: #1626. Related: #1264.

## Changes

- Route Cloud inline replies through `POST /api/v2/inline-comments` with `parentCommentId`.
- Preserve `extensions.location = inline` on Server/Data Center replies.
- Trace nested comment ancestry so legacy children inherit the root thread type.
- Preserve Cloud inline reply parent metadata and cover Cloud/Server paths with focused unit and E2E regressions.

## Testing

- [x] Manual checks performed: created inline and footer replies on Confluence Data Center and verified their locations and parent links.

Verification includes 76 targeted unit tests, the full Confluence unit suite, integration tests, pre-commit, and generated documentation checks. Real-service Cloud and Server/Data Center E2E regressions are included; local execution requires the corresponding test credentials and instances.

## Checklist

- [x] All change-related tests pass locally.
- [x] Documentation updated (not needed; no tool signature or registration changed).

### DIFF
--- a/src/mcp_atlassian/confluence/comments.py
+++ b/src/mcp_atlassian/confluence/comments.py
@@ -284,7 +284,7 @@ class CommentsMixin(ConfluenceClient):
             A tuple containing the owning page ID and ``"inline"`` or ``"footer"``.
 
         Raises:
-            ValueError: If the parent response contains no page reference.
+            ValueError: If no page reference exists or ancestry is cyclic.
         """
         current_id = comment_id
         visited: set[str] = set()
@@ -317,6 +317,9 @@ class CommentsMixin(ConfluenceClient):
             if not comment_parent_id:
                 break
             current_id = comment_parent_id
+        else:
+            message = f"Cyclic comment ancestry detected for '{comment_id}'"
+            raise ValueError(message)
 
         if not page_id:
             message = f"Could not resolve page for parent comment {comment_id}"
--- a/src/mcp_atlassian/confluence/comments.py
+++ b/src/mcp_atlassian/confluence/comments.py
@@ -192,32 +192,49 @@ class CommentsMixin(ConfluenceClient):
 
             v2_adapter = self._v2_adapter
             if v2_adapter:
-                response = v2_adapter.create_footer_comment(
-                    parent_comment_id=comment_id, body=content
-                )
+                location = v2_adapter.get_comment_thread_location(comment_id)
+                if location == "inline":
+                    response = v2_adapter.create_inline_comment(
+                        parent_comment_id=comment_id, body=content
+                    )
+                else:
+                    response = v2_adapter.create_footer_comment(
+                        parent_comment_id=comment_id, body=content
+                    )
                 space_key = ""
             else:
-                # v1 API: thread replies under the parent page with ancestors,
-                # not as a direct child of the parent comment.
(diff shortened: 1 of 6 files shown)


src/mcp_atlassian/confluence/comments.py
```diff
@@ -245,34 +262,66 @@ def _page_id_from_reference(reference: Any) -> str | None:
...
             ValueError: If the parent response contains no page reference.
         """
-        parent = self.confluence.get_page_by_id(
-            page_id=comment_id, expand="container,ancestors"
-        )
+        current_id = comment_id
+        visited: set[str] = set()
+        page_id: str | None = None
+
+        while current_id not in visited:
```
AmirF194, line 293: I ran this against a50bf13e3714a33c94da72b30b5885bf218cb8bc in a clean container, importing `comments.py` and `v2_adapter.py` directly with a fake `confluence` client.

`_resolve_parent_comment_context` guards against a cyclic ancestry with a `visited` set, but when the while loop exits because `current_id` is already in `visited`, execution falls through to the same code path as a legitimate `break` and silently returns `(page_id, "footer")`. `get_comment_thread_location` in `v2_adapter.py`, which you wrote in this same PR for the identical loop shape, raises `ValueError("Cyclic comment ancestry detected...")` instead.

```
v1 _resolve_parent_comment_context('A') on a cyclic A<->B ancestry returned ('999', 'footer') -- NO exception raised
v2 get_comment_thread_location('A') on the identical cycle raised ValueError: Cyclic comment ancestry detected for 'A'
```

I built the cycle by hand (A's ancestor is B, B's ancestor is A), so I have not seen this shape from a real Confluence response. But your v2 code already treats it as invalid data worth failing loudly on, so the v1 twin silently falling back to "footer" on the same shape looks like an oversight rather than a deliberate choice, and it would misclassify an inline thread as footer in exactly the case this PR is meant to fix.

A fix matching the v2 twin: after the while loop, check whether it ended because `current_id` was already in `visited` (vs a clean `break`), and raise the same way instead of returning `page_id, "footer"`.

huyusong10: Thanks for catching this — you are right. The v1 resolver now raises on cyclic ancestry, matching the v2 path, and an A/B cycle regression test was added in f39f2f6. All checks pass.

huyusong10:
Thanks for taking a look. I noticed this was closed just after #1626 merged. Is the write-side change not something you want to pursue, or would you be open to a smaller PR rebased onto current main with real Cloud and Server/DC verification? Happy to adjust accordingly.

src/mcp_atlassian/confluence/comments.py
AmirF194: Confirmed, the regression test covers the case I hand built. Thanks for the quick fix.

PR #1636, opened 2026-09-02
Fix/confluence page tree position expand

nold2:
### Summary
- `get_space_page_tree` requested `expand="ancestors"` only. Confluence's `/rest/api/content` endpoint doesn't return `extensions.*` fields by default, so `page["extensions"]["position"]` was always missing — every page's `position` silently came back `None`.
- Added `extensions.position` to the expand param. Fixed a stale comment that incorrectly claimed position was "auto-included."
- Updated `test_expand_uses_ancestors_only` (name kept, assertion corrected) to assert the real expand string sent to the API.

### Why it matters
`position` drives sibling ordering in the returned page tree. Without it, `get_space_page_tree` fell back to the `999999` default for every page, so ordering silently degraded to alphabetical-by-title instead of actual Confluence page order — with no error to signal it.

### Test plan
- [x] `uv run pytest tests/unit/confluence/` — 363 passed
- [x] `uv run pytest -q` (full suite) — 2578 passed, 161 skipped
- [x] `uv run pre-commit run --files src/mcp_atlassian/confluence/pages.py tests/unit/confluence/test_pages.py` — ruff, ruff-format, mypy clean

### Out of scope
Folders (Confluence Cloud) are excluded from the page tree entirely — `get_all_pages_from_space_raw` only supports `type=page`/`blogpost`, so a page nested under a folder gets a `parent_id` pointing to a node absent from the result. Tracked as follow-up.

### DIFF
--- a/register-mcp.sh
+++ b/register-mcp.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env sh
+# Registers mcp-atlassian as a global Claude Code MCP server using credentials from .env
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+ENV_FILE="$SCRIPT_DIR/.env"
+
+if [ ! -f "$ENV_FILE" ]; then
+  echo "Error: $ENV_FILE not found. Copy .env.example to .env and fill in credentials." >&2
+  exit 1
+fi
+
+# Parse .env — skip comments and blank lines, no subshell needed
+while IFS='=' read -r key value; do
+  case "$key" in
+    '#'*|'') continue ;;
+  esac
+  # Strip inline comments and leading/trailing whitespace from value
+  value="${value%%#*}"
+  value="${value#"${value%%[! ]*}"}"
+  value="${value%"${value##*[! ]}"}"
+  export "$key=$value"
+done < "$ENV_FILE"
+
+# Validate required vars
+missing=""
+for var in JIRA_URL CONFLUENCE_URL; do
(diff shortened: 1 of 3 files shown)


register-mcp.sh
```diff
@@ -0,0 +1,56 @@
...
+  CONFLUENCE_SPACES_FILTER JIRA_PROJECTS_FILTER; do
+  eval "val=\$$var"
+  [ -n "$val" ] && env_flags="$env_flags -e $var=$val"
+done
+
+echo "Registering mcp-atlassian (user scope)..."
+
+# shellcheck disable=SC2086
+claude mcp add mcp-atlassian \
+  --scope user \
```
AmirF194, line 52: I read this diff and the PR description against head 27641d3.

register-mcp.sh is new (it does not exist on main) and is not mentioned anywhere in the PR body, which only describes the extensions.position expand fix in pages.py. It also lives at the repo root while the project already has a scripts/ directory for this kind of helper.

Separately, the script builds `env_flags` from the Atlassian credential variables and passes them as `claude mcp add ... -e VAR=value ...` arguments, which puts tokens on the process command line (visible via ps and typically saved to shell history) rather than keeping them in the .env file the script already reads.

If this file was meant to ship with the fix, it probably belongs in scripts/ and should read credentials without exposing them as CLI arguments. If it was a local helper that got included by mistake, dropping it from this PR would keep the diff to the stated fix.

PR #1646, opened 2026-09-03
fix(confluence): limit simple-search fallback

### LINKED ISSUE #1644: [Bug]: confluence_search fallback catches ReadTimeout and doubles the blocking time
### Prerequisites 

- [x] I have searched existing issues
- [x] I have read the README

### Bug Description

`confluence_search` converts a plain-text query into `siteSearch ~ "..."` and, if that fails, falls
back to `text ~ "..."`. The fallback catches **any** exception:

https://github.com/sooperset/mcp-atlassian/blob/v0.23.0/src/mcp_atlassian/servers/confluence.py#L258-L270

```python
try:
    query = f'siteSearch ~ "{original_query}"'
    logger.info(f"Converting simple search term to CQL using siteSearch: {query}")
    pages = confluence_fetcher.search(query, limit=limit, spaces_filter=spaces_filter)
except Exception as e:                       # <-- also catches requests ReadTimeout
    logger.warning(f"siteSearch failed ('{e}'), falling back to text search.")
    query = f'text ~ "{original_query}"'
    logger.info(f"Falling back to text search with CQL: {query}")
    pages = confluence_fetcher.search(query, limit=limit, spaces_filter=spaces_filter)
```

(The code is identical on `main` as of 2026-08-31.)

When the first search fails with `requests.exceptions.ReadTimeout`, the server immediately issues a
**second** heavy search against an instance that has just demonstrated it is too slow to answer. A
single `confluence_search` call can therefore occupy the worker for up to `2 × CONFLUENCE_TIMEOUT`.

A timeout is not a "this CQL is unsupported" signal. The fallback was introduced in #270 to recover
from queries Confluence cannot parse, and that is the only case where retrying with different CQL
makes sense.

This matters more than it looks because these calls are synchronous and share the event loop with
`/healthz` (filed separately as #1643), so a doubled call blocks health checks for twice as long.

### Steps to Reproduce

1. Set `CONFLUENCE_TIMEOUT=25` (or leave the 75s default — the effect just takes longer to show).
2. Call `confluence_search` with a plain-text query (no `=`, `~`, `AND`, `OR` — so the siteSearch
   path is taken), e.g. `rate-correction template date format csv`.
3. Have the Confluence instance be slow enough that `siteSearch` hits the read timeout — on a large
   Server/DC instance under load this happens on its own.
4. Observe that the tool does not return the timeout: it starts a second search and only then
   answers.

### Expected Behavior

A read timeout propagates to the caller as an error. The fallback fires only for query-parsing
failures (`ApiValueError` / HTTP 400), which is what it was built for.

### Actual Behavior

```shell
# Two consecutive log lines from production, same call:
ERROR - mcp_atlassian.utils.decorators - Network error during search:
  _PinnedHTTPSConnectionPool(host='confluence.example', port=443): Read timed out. (read timeout=25)
WARNING - mcp_atlassian.servers.confluence - siteSearch failed
  ('Network error during search: _PinnedHTTPSConnectionPool(host='confluence.example', port=443):
  Read timed out. (read timeout=25)'), falling back to text search.

# The call did not fail at the timeout: it went on to run the fallback search and only
# then returned, so its wall-clock cost exceeded CONFLUENCE_TIMEOUT by the duration of
# the second search. In our traffic this is the only pattern that produces calls longer
# than the configured timeout.
```

### mcp-atlassian Version

```
0.23.0
```

### Installation Method

Docker

### Operating System

Linux

### Python Version

```
3.13
```

### Atlassian Instance Type

Confluence Server/Data Center

### Client Application

Claude Code

### Additional Context

**Scope of the fallback as introduced.** PR #270 added the siteSearch-with-text-fallback behaviour to
recover from queries Confluence cannot parse — in that case the retry with different CQL is what
makes the tool work. The current `except Exception` also covers transport failures, which are a
different class of error: the query itself was fine, the instance simply did not answer in time.

**Relation to the other issue we filed.** These Confluence calls are synchronous and share the event
loop with the `/healthz` handler, so a doubled call keeps health checks waiting for twice as long.
That coupling is reported in #1643; the doubling described here is observable on its own,
independently of how the server schedules the calls.

atirna:
## Description

I limited the simple-query `confluence_search` fallback to HTTP 400 responses. Read timeouts and other transport failures now return after the first `siteSearch` request instead of starting a second search.

## Testing

- `uv run pytest tests/unit/servers/test_confluence_server.py -q`
- `uv run pre-commit run --files src/mcp_atlassian/servers/confluence.py tests/unit/servers/test_confluence_server.py`

Fixes #1644

### DIFF
--- a/src/mcp_atlassian/servers/confluence.py
+++ b/src/mcp_atlassian/servers/confluence.py
@@ -13,6 +13,7 @@ from urllib.parse import parse_qs, urlsplit
 from fastmcp import Context
 from mcp.types import BlobResourceContents, EmbeddedResource, ImageContent, TextContent
 from pydantic import BeforeValidator, Field
+from requests.exceptions import HTTPError
 
 from mcp_atlassian.exceptions import MCPAtlassianAuthenticationError
 from mcp_atlassian.models.confluence import ConfluenceAttachment
@@ -262,7 +263,9 @@ async def search(
             pages = confluence_fetcher.search(
                 query, limit=limit, spaces_filter=spaces_filter
             )
-        except Exception as e:
+        except HTTPError as e:
+            if e.response is None or e.response.status_code != 400:
+                raise
             logger.warning(f"siteSearch failed ('{e}'), falling back to text search.")
             query = f'text ~ "{original_query}"'
             logger.info(f"Falling back to text search with CQL: {query}")
--- a/src/mcp_atlassian/servers/confluence.py
+++ b/src/mcp_atlassian/servers/confluence.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from typing import Annotated
 from urllib.parse import parse_qs, urlsplit
 
+from atlassian.errors import ApiError
 from fastmcp import Context
 from mcp.types import BlobResourceContents, EmbeddedResource, ImageContent, TextContent
 from pydantic import BeforeValidator, Field
@@ -263,8 +264,21 @@ async def search(
             pages = confluence_fetcher.search(
                 query, limit=limit, spaces_filter=spaces_filter
             )
-        except HTTPError as e:
-            if e.response is None or e.response.status_code != 400:
+        except (HTTPError, RuntimeError) as e:
+            # Confluence.cql() converts an HTTP 400 (unparseable CQL) into
+            # ApiValueError, which handle_atlassian_api_errors then wraps as
+            # RuntimeError, so the 400 can arrive here as either type.
+            status_error = e
+            if isinstance(e, RuntimeError):
+                status_error = e.__cause__
+                if not isinstance(status_error, ApiError):
+                    raise
+                status_error = status_error.reason
+            if (
+                not isinstance(status_error, HTTPError)
+                or status_error.response is None
+                or status_error.response.status_code != 400
+            ):
                 raise
             logger.warning(f"siteSearch failed ('{e}'), falling back to text search.")
             query = f'text ~ "{original_query}"'
--- a/tests/unit/servers/test_confluence_server.py
+++ b/tests/unit/servers/test_confluence_server.py
@@ -511,17 +511,26 @@ async def test_search(client, mock_confluence_fetcher):
 
 
 @pytest.mark.anyio
-async def test_search_returns_error_details(client, mock_confluence_fetcher):
+@pytest.mark.parametrize(
+    ("query", "expected_query"),
+    [("type=page", "type=page"), ("test search", 'siteSearch ~ "test search"')],
+)
+async def test_search_returns_error_details(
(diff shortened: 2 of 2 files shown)


src/mcp_atlassian/servers/confluence.py
```diff
@@ -262,7 +263,9 @@ async def search(
             pages = confluence_fetcher.search(
                 query, limit=limit, spaces_filter=spaces_filter
             )
-        except Exception as e:
+        except HTTPError as e:
```
AmirF194: Ran this against 6df65c9c7aaef92f56b662047ff154be1c6a894c in a clean container against the real atlassian-python-api dependency, not a mock.

On the malformed-CQL case this fallback exists for, `except HTTPError` here never fires. `atlassian.Confluence.cql()` catches the 400 itself and raises `ApiValueError`, not `HTTPError`. This repo's `handle_atlassian_api_errors` decorator does not recognize `ApiValueError` either (it is not `HTTPError`/`KeyError`/`RequestException`/`ValueError`/`TypeError`), so it falls through and re-raises as `RuntimeError`. That doesn't match `except HTTPError`, so the fallback never runs. The bare `except Exception` this PR replaces did catch it, so this regresses the exact case the log line below still describes.

```
STEP1: Confluence.cql() on HTTP 400 raised: atlassian.errors.ApiValueError
STEP1: isinstance of requests HTTPError? False
STEP2: decorated search() raised: RuntimeError: Unexpected error during search: ...
STEP3: NOT caught by 'except HTTPError' -- fallback never fires
```

The new test mocks `confluence_fetcher.search` with `ValueError` directly, so it never exercises the real exception type.

`raise RuntimeError(...) from http_err` chains the original error, so `isinstance(e.__cause__, HTTPError)` would catch it, or catch `atlassian.errors.ApiValueError` alongside `HTTPError`.

atirna: thx for the container run, that pinned it exactly.

pushed c238af1. the 400 now reaches the fallback through the chain you traced: the except also catches the RuntimeError the decorator wraps the ApiValueError in, unwraps to the ApiError cause, and checks its reason for the 400. catching ApiValueError directly cannot fire at this layer since handle_atlassian_api_errors re-raises it as RuntimeError before the tool sees it, and the wrapped cause is the ApiValueError itself rather than the HTTPError, so the status check goes through .reason.

the tests now drive that real chain (RuntimeError from ApiValueError with a 400 response) instead of a bare ValueError, plus no-fallback cases for a direct 500 HTTPError, a transport ValueError, and a plain RuntimeError.

AmirF194:
That matches the chain I traced. Makes sense that ApiValueError cannot be caught directly at this layer since the decorator already wraps it in RuntimeError before the tool sees it.

PR #1647, opened 2026-09-03
fix(jira): describe wiki markup in tool schemas when translation is disabled

### LINKED ISSUE #1594: [Feature]: When `DISABLE_JIRA_MARKUP_TRANSLATION` is set, change tool schemas to say "Wiki" instead of "Markdown"

- `jira_create_issue` - tool text currently includes "Issue description in Markdown format. On Jira Cloud, use '{expand:Title}...{expand}' for a collapsible section."
- `jira_batch_create_issues` - tool text currently includes "description (optional): Issue description in Markdown format"
- `jira_update_issue` - tool text currently includes "For 'description', provide text in Markdown format; on Jira Cloud, use '{expand:Title}...{expand}' for a collapsible section." (example still uses `"## Updated\nMarkdown text"`)
- `jira_add_comment` - tool text currently includes "Comment text in Markdown format"
- `jira_edit_comment` - tool text currently includes "Updated comment text in Markdown format"
- `jira_add_worklog` - tool text currently includes "(Optional) Comment for the worklog in Markdown format"
- `jira_transition_issue` - tool text currently includes "(Optional) Comment to add during the transition in Markdown format."

 When `DISABLE_JIRA_MARKUP_TRANSLATION` is set, tool text becomes:

- `jira_create_issue` - "Issue description. Server/DC: Jira wiki markup, passed through with no conversion. Cloud: Markdown (converted to ADF). On Jira Cloud, use '{expand:Title}...{expand}' for a collapsible section."
- `jira_batch_create_issues` - "description (optional): Server/DC: Jira wiki markup, passed through. Cloud: Markdown."
- `jira_update_issue` - "For 'description': Server/DC: Jira wiki markup, passed through with no conversion. Cloud: Markdown (converted to ADF)."
- `jira_add_comment` - "Comment text. Server/DC: Jira wiki markup, passed through. Cloud: Markdown."
- `jira_edit_comment` - "Updated comment text. Server/DC: Jira wiki markup, passed through. Cloud: Markdown."
- `jira_add_worklog` - "(Optional) Worklog comment. Server/DC: Jira wiki markup, passed through. Cloud: Markdown."
- `jira_transition_issue` - "(Optional) Transition comment. Server/DC: Jira wiki markup, passed through. Cloud: Markdown."

Or, even better, if possible: it figures out whether it's a Server/DC or Cloud on startup and tool text is specific to that.

_No response_

### Use Case

_No response_

### Additional Context

_No response_

Julien-ser:
Closes #1594.

## Problem

With `DISABLE_JIRA_MARKUP_TRANSLATION` set, text is written to Jira exactly as supplied. On Server/DC the format Jira actually stores is then wiki markup, not Markdown, but seven write tools still described their text fields as *"in Markdown format"*.

The model writes whatever the schema asks for, so it kept sending Markdown into a field that stores the bytes verbatim, and the Markdown ended up in the ticket as literal text. The schema was describing a format the server had been configured not to produce.

Affected: `jira_create_issue`, `jira_batch_create_issues`, `jira_update_issue`, `jira_add_comment`, `jira_edit_comment`, `jira_add_worklog`, `jira_transition_issue`.

## Change

Descriptions are built through a small `_body_format()` helper that switches on the setting, using the wording from the issue:

| Setting | `jira_add_comment.body` |
|---|---|
| default | `Comment text in Markdown format.` |
| `DISABLE_JIRA_MARKUP_TRANSLATION=true` | `Comment text. Server/DC: Jira wiki markup, passed through. Cloud: Markdown.` |

Two things beyond the literal ask, both in the same failure mode:

- **`jira_update_issue`'s worked example switches too.** It previously showed `"description": "## Updated\nMarkdown text"` regardless of the setting, and a worked example is the part a model is most likely to copy verbatim. It now shows `h2. Updated` when translation is off.
- **The flag moved into a shared `markup_translation_disabled()` helper** that both `JiraConfig` and the schemas call, so the format the schemas advertise cannot drift from what the conversion layer actually does. It deliberately keeps `JiraConfig`'s existing strict `== "true"` match rather than the looser `is_env_truthy()`, since changing that would alter runtime behaviour; a test pins that the two stay in agreement.

Docstring `Args:` lines that named Markdown are now format-neutral, since FastMCP surfaces the docstring as the tool description; the precise format lives in the parameter schema. The value is read once at import time, matching `ISSUE_KEY_PATTERN` directly above it.

The issue also floats detecting Server/DC vs Cloud at startup. I left that out: it needs a live connection, and these descriptions are built at import. The wording covers both deployments instead.

**No change in default behaviour** — with the flag unset every description is byte-identical to before. `scripts/generate_tool_docs.py --check` enforces this: `git diff origin/main -- docs/` is empty. (My first commit did not hold to that and shifted some default wording; the docs check caught it and it is reverted in a follow-up commit — see the comment below.)

## Testing

8 regression tests. Verified against the pre-fix code: 7 fail, and the 1 that passes is the default-behaviour test, which should pass either way.

- `tests/unit/servers/` + `tests/unit/jira/`: **2271 passed**. The one unrelated failure, `test_upload_attachment_success`, reproduces identically on unmodified `main` on this Windows machine (path separators) and is not related to this change.
- `ruff check` (with the repo's pre-commit ignore list) and `ruff format --check`: clean.
- `scripts/generate_tool_docs.py --check` passes, with the generated docs unchanged from `main`.

### DIFF
--- a/docs/tools/jira-comments-worklogs.mdx
+++ b/docs/tools/jira-comments-worklogs.mdx
@@ -14,7 +14,7 @@ Add a comment to a Jira issue.
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
 | `issue_key` | `string` | Yes | Jira issue key (e.g., 'PROJ-123', 'ACV2-642') |
-| `body` | `string` | Yes | Comment text in Markdown format. |
+| `body` | `string` | Yes | Comment text in Markdown format |
 | `visibility` | `string` | No | (Optional) Comment visibility as JSON string (e.g. '`{"type":"group","value":"jira-users"}`') |
 | `public` | `boolean` | No | (Optional) For JSM/Service Desk issues only. Set to true for customer-visible comment, false for internal agent-only comment. Posted via the ServiceDesk API as a raw string; Jira Cloud renders it server-side and stores ADF (markdown observed to render on Cloud, without the client-side markdown-to-ADF guarantees of the regular comment path). Cannot be combined with visibility. If the issue's project is listed in JIRA_INTERNAL_ONLY_PROJECTS, only public=false is accepted — public=true or omitting this field is rejected. Issues in such a project that are not JSM customer requests (e.g. an agent-created Task) have no portal audience and are exempt: they post through the ordinary comment path and ignore this field. |
 **Example:**
@@ -41,7 +41,7 @@ Edit an existing comment on a Jira issue.
 |-----------|------|----------|-------------|
 | `issue_key` | `string` | Yes | Jira issue key (e.g., 'PROJ-123', 'ACV2-642') |
 | `comment_id` | `string` | Yes | The ID of the comment to edit |
-| `body` | `string` | Yes | Updated comment text in Markdown format. |
+| `body` | `string` | Yes | Updated comment text in Markdown format |
 | `visibility` | `string` | No | (Optional) Comment visibility as JSON string (e.g. '`{"type":"group","value":"jira-users"}`') |
 
 ---
@@ -70,7 +70,7 @@ Add a worklog entry to a Jira issue.
 |-----------|------|----------|-------------|
 | `issue_key` | `string` | Yes | Jira issue key (e.g., 'PROJ-123', 'ACV2-642') |
 | `time_spent` | `string` | Yes | Time spent in Jira format. Examples: '1h 30m' (1 hour and 30 minutes), '1d' (1 day), '30m' (30 minutes), '4h' (4 hours) |
-| `comment` | `string` | No | (Optional) Worklog comment in Markdown format. |
+| `comment` | `string` | No | (Optional) Comment for the worklog in Markdown format |
 | `started` | `string` | No | (Optional) Start time in ISO format. If not provided, the current time will be used. Example: '2023-08-01T12:00:00.000+0000' |
 | `original_estimate` | `string` | No | (Optional) New value for the original estimate |
 | `remaining_estimate` | `string` | No | (Optional) New value for the remaining estimate |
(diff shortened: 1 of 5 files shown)


Julien-ser:
Heads-up on why I reverted part of the bot's `docs(jira): refresh markup schema references` commit, since it looks like I undid a fix.

The `generate_tool_docs.py --check` failure was a genuine defect in my first commit, and it caught something my own testing did not. Building the descriptions through the helper had also **shifted the default wording**, with the flag unset:

- `Comment text in Markdown format` gained a trailing period
- `(Optional) Comment for the worklog in Markdown format` became `(Optional) Worklog comment in Markdown format.`
- `(Optional) Comment to add during the transition...` became `(Optional) Transition comment...`
- and `jira_update_issue` read `...provide text in Markdown format. On Jira Cloud,` after my splice changed the original comma/semicolon clause

That contradicted the "no change in default behaviour" claim in my own PR description, so it was my error, not a docs problem.

The bot resolved the check by regenerating the docs to match that drift, which would have shipped the wording change. I have instead removed the drift: `_body_format()` now takes the existing description explicitly and returns it **unchanged** whenever translation is on, so the new wiki wording exists only on the disabled path. The generated docs then go back to exactly what they are on `main`.

`git diff origin/main -- docs/` is now empty, which is the actual proof that default schemas are untouched. `--check` passes.

I kept the bot's other fix (it removed a duplicate `@pytest.mark.anyio` I had left on one test) and rebased on top of its commit rather than force-pushing over it.

1129 server tests pass.

Julien-ser:
Rebased onto current `main`, resolving two real conflicts against the parent/Epic-Link guidance added there since — kept that new guidance and layered this PR's format-neutral wording on top of it rather than dropping either side. Docs regenerated (`scripts/generate_tool_docs.py`), diff vs. `main` is empty as before. Full `test_jira_server.py` suite (234 tests) passes, ruff clean.

PR #1648, opened 2026-09-03
fix(confluence): apply CONFLUENCE_SPACES_FILTER to space listings

amitvijapur:
## Problem

`CONFLUENCE_SPACES_FILTER` is enforced in `confluence_search`, but `get_spaces` and `get_user_contributed_spaces` return every space the credentials can see, ignoring the configured allowlist.

**Scope, corrected after review (thanks @AmirF194):** no MCP tool calls either method today, so this is not a live tool disclosure. `SpacesMixin` is mixed into `ConfluenceFetcher`, which is exported in `__all__`, so the gap is reachable by library consumers calling the fetcher directly. This is public-API hardening, not a fix for an exposed tool path.

## Change

Adds `ConfluenceClient._get_allowed_spaces()`, which parses the configured allowlist once and returns `None` when no filter is set, and applies it in both listing methods. When no filter is configured the behaviour and the request count are unchanged.

## Verification

- `pytest tests/unit/confluence/` — 530 passed.
- Two new tests, both mutation-checked: disabling either filter call makes its test fail, so neither is decorative.
- `mypy` and `ruff format` clean on the touched files.

## Scope and sequencing

This replaces #1519, which tried to cover all Confluence tools in one 25-file change and was too large to review.

Review also showed my original split used the wrong seam. I cut where the code was convenient (methods already holding a space key, needing no extra API call) rather than where the impact is. Around 30 fetcher methods are actually reached by tools, and they are the page-ID and content-ID addressed ones: `get_page_content`, `get_page_by_title`, `get_page_labels`, `get_page_restrictions`, `get_page_comments`, `get_space_permissions`, `list_page_templates`. Enforcement there requires resolving an ID to a space first, which changes the request count for filtered clients, and that is the next slice.

If a maintainer would rather see the page-ID work first, I am happy to close this and lead with that instead.

Disclosure: written with AI assistance, reviewed by me before opening.

### DIFF
--- a/src/mcp_atlassian/confluence/client.py
+++ b/src/mcp_atlassian/confluence/client.py
@@ -306,3 +306,14 @@ class ConfluenceClient:
         return self.preprocessor.process_html_content(
             html_content, space_key, self.confluence
         )
+
+    def _get_allowed_spaces(self) -> set[str] | None:
+        """Return the configured ``CONFLUENCE_SPACES_FILTER`` allowlist.
+
+        Returns:
+            A set of allowed space keys, or None when no filter is configured.
+        """
+        filter_to_use = getattr(self.config, "spaces_filter", None)
+        if not isinstance(filter_to_use, str) or not filter_to_use:
+            return None
+        return {s.strip().upper() for s in filter_to_use.split(",") if s.strip()}
--- a/src/mcp_atlassian/confluence/spaces.py
+++ b/src/mcp_atlassian/confluence/spaces.py
@@ -25,8 +25,26 @@ class SpacesMixin(ConfluenceClient):
             Dictionary containing space information with results and metadata
         """
         spaces = self.confluence.get_all_spaces(start=start, limit=limit)
-        # Cast the return value to the expected type
-        return cast(dict[str, object], spaces)
+        spaces_result = cast(dict[str, object], spaces)
+        allowed_spaces = self._get_allowed_spaces()
+        if allowed_spaces is None or not isinstance(spaces_result, dict):
+            return spaces_result
+
+        results = spaces_result.get("results")
+        if not isinstance(results, list):
+            return spaces_result
+
+        filtered_results = [
+            space
+            for space in results
+            if isinstance(space, dict)
+            and str(space.get("key", "")).strip().upper() in allowed_spaces
+        ]
+        filtered_response = dict(spaces_result)
+        filtered_response["results"] = filtered_results
+        if "size" in filtered_response:
+            filtered_response["size"] = len(filtered_results)
+        return filtered_response
 
     def get_user_contributed_spaces(self, limit: int = 250) -> dict:
         """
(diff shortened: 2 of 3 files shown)


AmirF194 (commented):
I checked this against `e476fe5` in a local clone (no live Atlassian instance).

The filter logic in `get_spaces` and `get_user_contributed_spaces` itself looks correct for the cases the new tests cover. My question is about reachability, not the filtering code.

I could not find a caller of either `get_spaces` or `get_user_contributed_spaces` anywhere in `src/mcp_atlassian`, so I could not tell which tool actually discloses the space list this PR is closing off:

```
$ grep -rn "\.get_spaces(\|\.get_user_contributed_spaces(" --include="*.py" . | grep -v tests/
(no output)
$ grep -n "@confluence_mcp.tool" src/mcp_atlassian/servers/confluence.py | wc -l
35
$ git log --oneline -- src/mcp_atlassian/confluence/spaces.py
e476fe5 fix(confluence): apply CONFLUENCE_SPACES_FILTER to space listings
47a9e25 Major Refactoring of Atlassian Integration Components (#72)
```

Both methods exist only in `src/mcp_atlassian/confluence/spaces.py`, called from the e2e auth-matrix tests directly against the Python class, and none of the 35 `@confluence_mcp.tool()` functions in `servers/confluence.py` call either one. `spaces.py` has had exactly two commits ever, so this isn't a case of a tool being removed later and the method left behind.

Is there a tool path I'm missing that reaches these, or is this hardening ahead of a tool that exposes them later? If it's the latter, worth saying so in the PR description, since "disclosed through those two tools" reads as describing a live exposure.

amitvijapur:
You're right, and thank you for checking reachability rather than just the diff.

I verified it independently: no references to either method anywhere in `src/` outside their definitions, no dynamic dispatch onto the fetcher, and no space-listing tool among the 35 in `servers/confluence.py`. There is no tool path. My description was wrong to say "disclosed through those two tools".

What is true is narrower. `SpacesMixin` is mixed into `ConfluenceFetcher`, which is exported in `__all__`, so a library consumer calling `ConfluenceFetcher(...).get_spaces()` directly does bypass a configured allowlist. That is a real gap, but it is public-API hardening rather than a live MCP exposure, and I have rewritten the description to say only that.

The more useful thing your question surfaced is that I split the original #1519 along the wrong seam. I cut it where the code was convenient, methods that already hold a space key and need no extra API call, and then described that seam as though it tracked security impact. It does not. Grepping the tool surface, roughly 30 fetcher methods are actually called by tools, and they are the page-ID and content-ID addressed ones I deferred: `get_page_content`, `get_page_by_title`, `get_page_labels`, `get_page_restrictions`, `get_page_comments`, `get_space_permissions`, `list_page_templates`. That is where the live exposure is, and the next slice will cover it.

I am leaving this open since the public-API gap is genuine and the change is small, but I would understand entirely if a maintainer would rather see the page-ID work first. Happy to close it on request.

Disclosure: written with AI assistance, reviewed by me before posting.

AmirF194:
Thanks for checking that independently and for the follow-up detail on the page-ID/content-ID methods. Good to know the next slice targets the actual tool-reachable surface.

amitvijapur:
Merged current main to clear the `BEHIND` state. `pytest tests/unit/confluence/` is 531 passed on `afe9c04`.

Also, thank you for the fail-closed fix in `c452b84` — that was a real hole and I should have caught it. My version returned the response unfiltered when `results` was not a list:

```python
results = spaces_result.get("results")
if not isinstance(results, list):
    return spaces_result   # every space leaks past the allowlist
```

A filter that passes everything through when it cannot parse the input is worse than no filter, because the operator believes it is enforced. Raising is right. Hoisting `_get_allowed_spaces()` out of the per-result loop is a good catch too.

I mutation-tested the happy path when I opened this, confirming each filter call was load-bearing, but never tested a malformed response. For a security control that was the obvious case to cover and I missed it.

Still happy to follow with the page-ID and content-ID enforcement, which is where the tool-reachable surface actually is, whenever this one is settled.

PR #1656, opened 2026-09-07
fix(jira): keep a Markdown link title out of the href

### LINKED ISSUE #1650: [Bug]: A Markdown link title ends up inside the href, so the stored link URL is broken
### Prerequisites

### Bug Description

A Markdown inline link can carry an optional title: `[docs](https://example.com/page "Docs home")`. On the Jira Cloud write path the title is not separated from the URL, so the link that gets built points at `https://example.com/page "Docs home"`. The link text survives, the URL does not.

Both quote styles do it, and the parenthesised `(Title)` form does too (it also swallows the closing bracket). A link written without a title is fine.

This is the conversion behind `jira_create_issue` and `jira_update_issue` (description), `jira_add_comment`, `jira_edit_comment` and `jira_add_worklog`.

### Steps to Reproduce

1. `pip install mcp-atlassian==0.23.1`
2. Save as `repro.py`:

```python
from mcp_atlassian.models.jira.adf import markdown_to_adf

def hrefs(n, out):
    if isinstance(n, list):
        for x in n:
            hrefs(x, out)
    elif isinstance(n, dict):
        for m in n.get("marks") or []:
            if m.get("type") == "link":
                out.append((n.get("text"), m["attrs"].get("href")))
        hrefs(n.get("content") or [], out)
    return out

for line in [
    'See [the runbook](https://wiki.example.com/rb "Runbook") first',
    "See [the runbook](https://wiki.example.com/rb 'Runbook') first",
    "See [the runbook](https://wiki.example.com/rb) first",
]:
    print(line)
    print("  ->", hrefs(markdown_to_adf(line)["content"], []))
```

3. `python repro.py`

### Expected Behavior

`href` is `https://wiki.example.com/rb` in all three, and the title is either carried as the link title or dropped.

### Actual Behavior

```shell
See [the runbook](https://wiki.example.com/rb "Runbook") first
  -> [('the runbook', 'https://wiki.example.com/rb "Runbook"')]
See [the runbook](https://wiki.example.com/rb 'Runbook') first
  -> [('the runbook', "https://wiki.example.com/rb 'Runbook'")]
See [the runbook](https://wiki.example.com/rb) first
  -> [('the runbook', 'https://wiki.example.com/rb')]
```

### mcp-atlassian Version

0.23.1

### Installation Method

### Operating System

### Python Version

3.12.14

### Atlassian Instance Type

Jira Cloud

### Client Application

custom script

### Additional Context

Same on main at 22369b76. The `[d](url (Title))` form gives `url (Title` with the closing bracket gone.

I have not pointed this at a live Cloud site, so what I measured is the ADF the conversion produces before the REST call, not what Jira then renders.

ConnorMoss02:
Closes #1650

## Problem

The link pattern captures everything up to the closing paren as the href:

```python
r"|\[(?P<link_text>[^\]]+)\]\((?P<link_href>[^)]+)\)"
```

An optional Markdown title is stored as part of the URL. Reproduced with the script in the issue:

```
[docs](https://example.com/p "Docs home")   -> href='https://example.com/p "Docs home"'
[docs](https://example.com/p 'Docs home')   -> href="https://example.com/p 'Docs home'"
[docs](https://example.com/p (Docs home))   -> href='https://example.com/p (Docs home'  + stray ')' as text
[docs](https://example.com/p)               -> href='https://example.com/p'
```

The paren form also stops at the title's own closing paren, so the trailing `)` is left behind as text. Reaches `jira_create_issue`, `jira_update_issue`, `jira_add_comment`, `jira_edit_comment` and `jira_add_worklog`.

## Change

The href becomes non-greedy and an optional quoted or parenthesised title is matched after it:

```python
r"|\[(?P<link_text>[^\]]+)\]\((?P<link_href>[^)]+?)"
r"(?:\s+(?P<link_title>\"[^\"]*\"|'[^']*'|\([^)]*\)))?\s*\)"
```

The title is carried on the link mark rather than dropped, which ADF supports.

After:

```
[docs](https://example.com/p "Docs home")   -> href='https://example.com/p'  title='Docs home'
[docs](https://example.com/p 'Docs home')   -> href='https://example.com/p'  title='Docs home'
[docs](https://example.com/p (Docs home))   -> href='https://example.com/p'  title='Docs home'
```

## What still behaves the same

A URL containing a space resolves as one href, since the title group only matches a quoted or parenthesised suffix and the href expands when neither is present:

```
[docs](https://example.com/a b)   -> href='https://example.com/a b'
```

Checked query strings, fragments, `mailto:`, and two links in one line; all unchanged.

## Tests

`tests/unit/models/test_adf.py`: one test covering the three title forms including the stray paren, one asserting a space-containing URL still resolves as one href. Both fail with the source change reverted.

```bash
uv run pytest tests/unit/models/ tests/unit/jira/test_comments.py -q
uv run ruff format --check src/mcp_atlassian/models/jira/adf.py tests/unit/models/test_adf.py
uv run ruff check --ignore=BLE001,EM102,FBT001,FBT002,E501,F841,S112,S113,B904 src/mcp_atlassian/models/jira/adf.py tests/unit/models/test_adf.py
uv run pre-commit run mypy --all-files
uv run python scripts/generate_tool_docs.py --check
```

439 passed, 5 skipped. Formatting, lint, mypy and generated-doc checks pass.


src/mcp_atlassian/models/jira/adf.py
```diff
@@ -112,7 +112,8 @@ def _parse_inline_formatting(
         r"title=(?P<status_title>[^}]+)\}"
         r"|\*\*(?P<bold_inner>.+?)\*\*"
         r"|~~(?P<strike_inner>.+?)~~"
-        r"|\[(?P<link_text>[^\]]+)\]\((?P<link_href>[^)]+)\)"
+        r"|\[(?P<link_text>[^\]]+)\]\((?P<link_href>[^)]+?)"
+        r"(?:\s+(?P<link_title>\"[^\"]*\"|'[^']*'|\([^)]*\)))?\)"
```
ebarkhordar: I filed #1650, so I ran this at 8b14af4 in a clean container (`python:3.12-slim`; the module is stdlib only, so `adf.py` imports standalone).

All three title forms are fixed when the title sits flush against the closing paren. Put a space before that paren and they all fall back to the old behaviour, because the new optional group is followed directly by `\)`:

```
[docs](https://example.com/p "Docs home")    -> href='https://example.com/p' title='Docs home'
[docs](https://example.com/p "Docs home" )   -> href='https://example.com/p "Docs home" ' title=None
[docs](https://example.com/p 'Docs home' )   -> href="https://example.com/p 'Docs home' " title=None
[docs](https://example.com/p (Docs home) )   -> href='https://example.com/p (Docs home' title=None, plus a stray ' )' text node
```

That space is allowed: markdown-it-py 3.0.0 in commonmark mode renders all three of them as `href="https://example.com/p" title="Docs home"`.

An `\s*` before the closing paren covers it:

```python
r"(?:\s+(?P<link_title>\"[^\"]*\"|'[^']*'|\([^)]*\)))?\s*\)"
```

I ran your three new title cases, your `[docs](https://example.com/a b)` case and the existing `test_link` input through a patched copy: identical output to this branch on all five, and the space inside that href is still kept. That is five inputs through `markdown_to_adf`, not the test suite, which I did not run.

ConnorMoss02:
Confirmed, and the `\s*` is in as `6c62431`.

```
[docs](https://example.com/p "Docs home" )   -> href='https://example.com/p'  title='Docs home'
[docs](https://example.com/p 'Docs home' )   -> href='https://example.com/p'  title='Docs home'
[docs](https://example.com/p (Docs home) )   -> href='https://example.com/p'  title='Docs home'
```

The paren form no longer leaves the stray ` )` node. `[docs](https://example.com/a b)` still resolves as one href, which was the case I was most worried about breaking.

Added your three spaced forms to `test_link_title`; they fail without the `\s*`. 439 passed, 5 skipped, and the lint, mypy and generated-doc checks are clean.

ebarkhordar:
Read 6c62431. The `\s*` landed where I had it and the three spaced forms are pinned in test_link_title. I did not rerun the suite.

Issue #877, opened 2026-01-26
bug: DEFAULT_READ_JIRA_FIELDS set comparison fails due to hash randomization


major:
## Problem

`DEFAULT_READ_JIRA_FIELDS` is a set, but the code compares joined strings:

```python
if fields_param == ",".join(DEFAULT_READ_JIRA_FIELDS) or fields_param == "*all":
```

Python's hash randomization causes different iteration order across processes, so this comparison fails when the MCP server and the library compute the join separately.

## Impact

The block that auto-adds `changelog`, `properties`, and `comment` fields gets skipped intermittently.

## Fix

Compare as sets instead of strings. Included in #876.

major:
Definitely not stale. ;)

PR #1661, opened 2026-09-09
fix(jira): make default field schemas deterministic

### LINKED ISSUE #1662: [Bug]: Jira tool schemas are nondeterministic because default fields are serialized from an unordered set
### Prerequisites

### Bug Description

Four Jira read tools build their `fields` default by joining `DEFAULT_READ_JIRA_FIELDS`, which is a set:

```python
default=",".join(DEFAULT_READ_JIRA_FIELDS),
```

Python randomises string hashing per process, so set iteration order differs between processes. Two workers running the same version therefore advertise two different `fields.default` values for the same tool:

```text
created,labels,updated,description,reporter,assignee,issuetype,priority,status,versions,summary
labels,description,summary,status,updated,assignee,versions,created,issuetype,reporter,priority
```

Affected tools:

- `jira_get_issue`
- `jira_search`
- `jira_get_board_issues`
- `jira_get_sprint_issues`

The order carries no meaning for Jira, but it changes the emitted tool schema, so the tool catalog gets a different fingerprint per process. Clients that cache on the catalog (prompt caching in LLM clients, schema-hash caches, catalog diffing) see the tool set as having changed when nothing changed, and replicas behind one endpoint disagree with each other.

This is the same root cause as #877, which fixed the *consumer* side by comparing sets instead of joined strings. The *producer* side still serializes the set into the schema.

### Steps to Reproduce

1. Clone the repo and install: `git clone https://github.com/sooperset/mcp-atlassian && cd mcp-atlassian && uv sync --all-extras`
2. Save this as `fingerprint.py` in the repo root. No credentials or network access are needed, since it only lists tools:

```python
import asyncio
import hashlib
import json
import os

import mcp_atlassian.servers.jira as jira_server


async def main() -> None:
    tools = await jira_server.jira_mcp.list_tools()
    catalog = [{"name": t.name, "schema": t.parameters} for t in tools]
    payload = json.dumps(catalog, separators=(",", ":"))
    digest = hashlib.sha256(payload.encode()).hexdigest()
    fields_default = next(
        t.parameters["properties"]["fields"]["default"]
        for t in tools
        if t.name == "get_issue"
    )
    print(
        f"seed={os.environ.get('PYTHONHASHSEED'):>3} "
        f"tools={len(catalog)} {digest[:16]} {fields_default[:46]}"
    )


asyncio.run(main())
```

3. Run it under several hash seeds, which is what separate worker processes get:

```bash
for s in 0 1 2 3 4 5; do PYTHONHASHSEED=$s uv run python fingerprint.py; done
```

### Expected Behavior

The tool catalog is a function of the code version, so every process running the same version should emit byte-identical schemas and therefore one fingerprint, whatever the hash seed.

### Actual Behavior

```shell
seed=  0 tools=63 da9e9323a79f6875 labels,status,updated,reporter,assignee,versio
seed=  1 tools=63 6e06849aa88845b8 reporter,updated,summary,labels,versions,statu
seed=  2 tools=63 51fb9509dbd0443c assignee,labels,reporter,versions,updated,crea
seed=  3 tools=63 510b201051421f33 issuetype,versions,summary,updated,created,des
seed=  4 tools=63 af5ffaaf15d510e9 versions,labels,status,reporter,description,cr
seed=  5 tools=63 df9a0095c1e41270 status,summary,issuetype,reporter,description,
```

Six seeds, six different catalog fingerprints, with `fields.default` reordered each time.

### mcp-atlassian Version

0.23.1.dev (`main` at 74bdaa8)

### Installation Method

From source (git clone)

### Operating System

Linux

### Python Version

3.11.14

### Atlassian Instance Type

Jira Cloud

### Client Application

Custom script calling `list_tools`. Originally observed through an MCP client that caches on the tool catalog, where two replicas of one endpoint produced two fingerprints.

### Additional Context

The bug is independent of the Atlassian instance and of credentials, since it happens while emitting the schema and before any Atlassian call.

The Confluence catalog is not affected: its tools default `fields` to `None`, and it is already stable across seeds. Within the Jira server, other set-typed constants such as `_GET_ISSUE_INCLUDE_SECTIONS` are only used for membership tests and never reach a schema, so these four defaults appear to be the only place set ordering escapes into the tool catalog.

VladanRakic:
## Description

Makes the Jira default `fields` serialization stable, so every process emits identical tool schemas.

The order is pinned where the string is built rather than normalized in clients, because the emitted schema is what gets fingerprinted, and otherwise every client would have to compensate for the same drift. One precomputed CSV now feeds both the tool schema defaults and the backend `fields` requests, so the two can no longer disagree either.

This is purely a serialization change. The field list, its meaning to Jira, and the public constant's type are all unchanged, so no importer or caller behaves differently.

Fixes: #1662

## Changes

- Fix the serialization order at the source with a module-private ordered tuple, `_DEFAULT_READ_JIRA_FIELDS_ORDERED`.
- Precompute `DEFAULT_READ_JIRA_FIELDS_CSV` once, and use it for the four tool schema defaults and for the backend `fields` requests in `jira/issues.py` and `jira/search.py`.
- Leave the public `DEFAULT_READ_JIRA_FIELDS` exactly as it is today, a `set[str]` with the same members, so this is not an API change for any importer. It is now derived from the tuple, and the order-independent comparison added for #877 in `jira/issues.py` is untouched.

## Testing

- [x] Relevant integration tests passed
- [x] Manual checks performed

Maintainer verification on the current `main`:

- `417 passed` across the focused Jira and server unit suites.
- `8 passed, 15 skipped` in the integration suite; the skipped cases are the separately gated real-API scenarios.
- `6 passed` for the affected Jira reads against Data Center.
- Jira Cloud basic-auth `search_issues` and `get_issue` both passed against the live API.
- Six fresh processes with distinct hash seeds emitted one identical schema-default fingerprint.
- Generated tool documentation is current (`98` tools across `14` categories and `25` toolsets).

The repository-wide Cloud E2E command was also invoked, but its shared fixture skipped because the configured Confluence credential is currently rejected. The affected Jira Cloud paths were therefore verified directly with the valid basic-auth credential.

Two regression tests are included:

- `test_csv_preserves_field_order` pins the exact CSV literal, so any reordering of the tuple fails the build.
- `test_default_jira_field_schemas_are_deterministic` asserts all four tool schemas expose that same CSV, so a fifth tool added later cannot quietly reintroduce a set join.

## Checklist

- [x] Relevant tests pass locally.
- [x] Documentation checked; regeneration produces no diff because generated tables do not include parameter defaults.


VladanRakic:
Hi @sooperset,
Can you take a look at the PR? 
I would rely appreciate some version of this fix making it to the next release 🤗

VladanRakic:
Addressed with: #1685

Issue #1634, opened 2026-09-01
[Bug]: Jira Server/DC attachment download fails with PAT — /secure/attachment redirects to login


trynocoding:
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/sooperset/mcp-atlassian/issues) to make sure this bug has not already been reported.
- [x] I have checked the [README](https://github.com/sooperset/mcp-atlassian/blob/main/README.md) for relevant information.

### Bug Description

With Jira Server 8.14.1 and mcp-atlassian v0.23.0, PAT authentication works for normal Jira REST APIs, but attachment downloads fail.

`jira_get_issue` / `jira_search` work normally, and attachment metadata can be retrieved. However, `jira_download_attachments` / `jira_get_issue_images` return `Fetch failed`.

The attachment `content` URL returned by Jira points to:

`/secure/attachment/<id>/<filename>`

On the Jira Server 8.14.1 instance tested here:

- REST API + Bearer PAT → `200 OK`
- `/secure/attachment/*` + Bearer PAT → `302` → Jira login page
- The same `/secure/attachment/*` URL + Basic Auth → `200 OK` with the actual attachment bytes

For private-network Jira deployments, the 302 redirect can also be blocked by the SSRF redirect hook because the configured Jira hostname resolves to a private IP. This masks the authentication problem and surfaces only as `Fetch failed`.

### Steps to Reproduce

1. Configure mcp-atlassian with Jira Server 8.14.1 and PAT authentication:

   `JIRA_URL=https://jira.example.internal`

   `JIRA_PERSONAL_TOKEN=<PAT>`

2. Verify that the REST API works:

   `curl -H "Authorization: Bearer <PAT>" https://jira.example.internal/rest/api/2/issue/TEST-1`

   Result: `200 OK`

3. Get an attachment `content` URL, for example:

   `https://jira.example.internal/secure/attachment/12345/report.pdf`

4. Request it using the same PAT:

   `curl -i -H "Authorization: Bearer <PAT>" https://jira.example.internal/secure/attachment/12345/report.pdf`

   Result: `302` redirect to the Jira login page.

5. Request the exact same URL using Basic Auth:

   `curl -L -u "<user>:<password>" https://jira.example.internal/secure/attachment/12345/report.pdf -o report.pdf`

   Result: `200 OK` with the actual attachment.

6. Call `jira_download_attachments` or `jira_get_issue_images`.

   Result: `Fetch failed` / `Download failed`.

[...]

PR #1667, opened 2026-09-12
fix(server): allow redirects to the session's own configured host

unwnu:
## Description

Fix redirects to configured private-network Jira/Confluence instances and relative `Location` handling with a shared, session-bound redirect-hook implementation.

Addresses the SSRF part of #1634. PAT attachment limitations and login-page detection remain out of scope. The initial-URL validation in #1568 still needs a separate adjustment for private instances.

## Changes

Exempt these transitions from non-global-address rejection:

- Configured origin (scheme, hostname, port) → itself.
- Configured HTTP origin, on any port → HTTPS:443 on the same hostname.
- That HTTPS:443 origin → itself, including relative redirects.

This deliberately adds a second trusted origin for HTTP sessions, without requiring a preceding upgrade. Both ends of each redirect are checked; downgrades, other ports and unrelated source origins receive normal SSRF validation. Scheme, backslash-authority and domain-allowlist checks remain enforced. The DNS-pinning adapter is unchanged.

Requests' forwarding rules are unchanged: HTTP:80 → HTTPS:443 preserves `Authorization`; HTTP:8080 → HTTPS:443 removes it. Custom headers, matching cookies and POST bodies on 307/308 can reach the additional HTTPS service.

## Testing

- [x] Unit tests added/updated. New positive regression checks fail with the previous hook.
- [x] Integration tests passed as part of the local suite. Full pytest with `--integration`: 3963 passed, 235 skipped. Real-API/E2E tests requiring additional flags or environment configuration remain skipped.
- [x] Manual checks performed against a live Jira 10.3.17 instance, as detailed below.

- Another 55 isolated probes covered redirect chains, forwarding behavior and SSRF boundaries.
- Live Jira 10.3.17: unauthenticated HTTP:80 → HTTPS:443 → login was blocked at the first 301 with `31be007`; the candidate followed 301 → 302 → 200. The final response was login HTML.
- HTTPS with a valid PAT returned the same 31-byte attachment and SHA-256 in both versions.

Live checks sent no PAT over HTTP. Port 8080 was unreachable, so its upgrade case was tested locally only.

## Checklist

- [x] Code follows project style guidelines. Ruff, formatting and mypy passed for the latest hook and test updates using the repository's pre-commit options.
- [x] All tests pass locally, with 235 skipped as reported above.
- [x] Documentation updated where needed: redirect-hook docstring reflects the supported transitions.


src/mcp_atlassian/utils/urls.py
```diff
@@ -8,18 +8,80 @@
...
+                ) from e
+
+            # Waive only when the hop stays inside the session's own origin.
+            # Requiring the source too stops an off-origin hop from pivoting back
+            # in with the exemption applied.
+            trusted_host = None
+            if (
+                trusted_origin is not None
+                and _origin(response.url) == trusted_origin
+                and _origin(redirect_url) == trusted_origin
```
ebarkhordar: I ran this branch's hook at `31be007d` in a clean `python:3.12-slim` container, with `jira.internal` in `/etc/hosts` on `10.0.0.5` and no DNS mocking.

The summary names three shapes this lets through: session expiry to `/login.jsp`, canonical base-URL redirects, and reverse-proxy redirects. Only the first is same-origin. The other two normally change scheme or port, and the comparison here is on the whole origin, so they are still refused whenever `JIRA_URL` is not already the canonical origin:

```
base = http://jira.internal:8080
  Location: /login.jsp                       -> ALLOWED
  Location: https://jira.internal/login.jsp  -> BLOCKED (SSRF): DNS for jira.internal resolves to non-global IP: 10.0.0.5
  Location: http://jira.internal/login.jsp   -> BLOCKED (SSRF): same
```

That base URL is the ordinary shape when the proxy terminates TLS and `JIRA_URL` points at the app server port.

`test_scheme_downgrade_on_the_base_host_is_not_trusted` pins https to http, and the reason it gives (session headers and non-Secure cookies survive the change) is specific to the downgrade. The upgrade is pinned in neither direction.

The narrow form, if you want it, is one predicate beside the origin check:

```python
def _same_or_upgraded(dst, trusted):
    if dst == trusted:
        return True
    # http -> https on the same host, default port only
    return (trusted[0], dst[0], dst[1], dst[2]) == ("http", "https", trusted[1], 443)
```

That adds one port, reached over TLS, on a host the session already talks to. It leaves the downgrade and the other-port cases exactly as they are. `ssrf_adapter.py` already trusts that host on every port, since `_pinned_create_connection` passes only the host to `_hostname_matches_allowlist`, so this hook is the only layer refusing the hop.

I did not test against a real Jira DC. Whether your 10.3.17 emits a canonical redirect in that configuration is something you can check and I cannot.

unwnu: Agreed, thanks for checking this and suggesting the narrower exception.

Added the same-host HTTPS:443 upgrade, including subsequent redirects within that HTTPS origin. Downgrades and unrelated source origins still go through normal SSRF validation.

Verified against a live Jira instance: HTTP:80 → HTTPS:443 → login follows the full 301 → 302 → 200 chain. Port 8080 was unreachable, so that case is covered by local tests.

Requests’ auth handling is unchanged: 80 → 443 preserves Authorization; 8080 → 443 strips it. Updated the description to match the supported redirects.

ebarkhordar: Re-ran the boundary cases against `f0207f4f` in the same container, real DNS, `jira.internal` on `10.0.0.5`:

```
base = http://jira.internal:8080
  http://jira.internal:8080/rest  -> https://jira.internal/login.jsp   ALLOWED
  https://jira.internal/login.jsp -> /next                             ALLOWED
  https://jira.internal/x         -> http://jira.internal:8080/y       BLOCKED
  http://jira.internal:8080/x     -> https://jira.internal:8443/y      BLOCKED
  http://jira.internal:9090/x     -> https://jira.internal/y           BLOCKED
  http://jira.internal:8080/x     -> https://evil.jira.internal/y      BLOCKED
  https://evil.example.com/x      -> https://jira.internal/y           BLOCKED
```

Same set at `31be007d` blocks the first two and agrees on the rest, so the widening is those two transitions and nothing else. The `https://jira.internal:8443` base is still strict about `:443`, which is what the description now says it supports.

Thanks for the forwarding note. The `8080 -> 443` auth strip was the part I could not check from here.

PR #1671, opened 2026-09-13
fix(confluence): offload blocking fetcher calls to worker threads (#1643)

### LINKED ISSUE #1643: [Bug]: Blocking sync calls share the event loop with /healthz — Kubernetes restarts healthy pods
### Prerequisites 

- [x] I have searched existing issues
- [x] I have read the README

### Bug Description

Every Confluence tool is declared `async def`, but its body makes a **blocking** call into
`atlassian-python-api` (which uses `requests`) without handing it off to a worker thread. The
`/healthz` route is an ordinary async handler on that same event loop:

- https://github.com/sooperset/mcp-atlassian/blob/v0.23.0/src/mcp_atlassian/servers/main.py#L894
- https://github.com/sooperset/mcp-atlassian/blob/v0.23.0/src/mcp_atlassian/servers/confluence.py#L262

While a tool waits on Atlassian, the pod cannot answer its liveness probe. It is not broken — it is
busy. Kubernetes cannot tell the difference, so it restarts the container and every in-flight MCP
session on that pod dies with it.

This is not a theoretical race: on our deployment it produces a steady stream of restarts under
completely normal traffic.

The Jira tools behave differently here: `run_jira_fetcher_call` in
`src/mcp_atlassian/servers/async_utils.py` wraps blocking fetcher calls in `anyio.to_thread.run_sync`
with a `CapacityLimiter` and a `JIRA_FETCHER_MAX_WORKERS` knob. It is used in two places in
`servers/jira.py` and **not at all** in `servers/confluence.py` — `grep -c to_thread
src/mcp_atlassian/servers/confluence.py` returns 0 on `main`.

### Steps to Reproduce

1. Run the server with `--transport streamable-http` and `--stateless` behind Kubernetes.
2. Point it at a Confluence Server/DC instance large enough that a typical tool call takes several
   seconds rather than milliseconds.
3. Configure a liveness probe on `/healthz` — for example the one from the project's own docs
   (`docs/advanced/docker-production.mdx`): `periodSeconds: 30`, default `timeoutSeconds: 1` and
   `failureThreshold: 3`, i.e. a 90-second window.
4. Send ordinary agent traffic: a handful of tool calls per minute per pod.
5. Watch pods get restarted by the kubelet while they are healthy and serving.

### Expected Behavior

`/healthz` answers regardless of how long a tool call takes, so an orchestrator can distinguish
"the process is wedged" from "the process is busy waiting on Atlassian".

### Actual Behavior

```shell
# Container exit codes seen on restarts:
#   reason=Error     exitcode=137   (SIGKILL after the grace period)
#   reason=Completed exitcode=0     (uvicorn shut down cleanly on SIGTERM)
#
# Nothing is exhausted at the moment of the restarts:
#   memory working set is a small fraction of the container limit
#   CPU usage is far below the limit, container_cpu_cfs_throttled_seconds_total == 0
#   container_oom_events_total == 0
#
# Restarts happen daily under ordinary traffic, on pods that are serving requests normally.
```

We enabled `MCP_VERBOSE=true` and paired `Processing request of type CallToolRequest` with the
uvicorn access-log line on the same pod to see how long individual tool calls actually take. On our
instance a routine call takes seconds, and a noticeable share of calls run well into tens of
seconds — long enough that a handful of consecutive calls keep the loop busy past any reasonable
liveness window. Load is spread evenly across pods and each pod receives only a few calls per
minute, so this is execution time, not queueing.

Since `/healthz` shares that loop, the pod stops answering probes while it is working, and the
kubelet kills it.

### mcp-atlassian Version

```
0.23.0
```

### Installation Method

Docker

### Operating System

Linux

### Python Version

```
3.13
```

### Atlassian Instance Type

Confluence Server/Data Center

### Client Application

Claude Code

### Additional Context

**Asymmetry with the Jira side.** Jira tools do not have this property, because blocking fetcher
calls there go through `run_jira_fetcher_call` in `src/mcp_atlassian/servers/async_utils.py`, which
runs them via `anyio.to_thread.run_sync` under a `CapacityLimiter` sized by
`JIRA_FETCHER_MAX_WORKERS` (default 8). On `main` that helper is referenced twice in
`servers/jira.py` and zero times in `servers/confluence.py`.

**What we tried on our side.** We capped `CONFLUENCE_TIMEOUT` at 25s (the default is 75s) and widened
the liveness window. That reduced the restart rate noticeably but did not stop it: probe thresholds
only change how long a busy pod may stay silent before it is killed, they do not change the fact that
`/healthz` cannot answer while a tool call is in flight.

**Related.** #1644 describes a separate defect in `confluence_search` that makes a single call cost up to twice `CONFLUENCE_TIMEOUT`, which lengthens exactly the window described here.

sanjibani:
Closes #1643.

## Bug

Confluence tool handlers are `async def` but call the blocking `atlassian-python-api` client directly on the event loop (search, page CRUD, comments, labels, attachments, templates, permissions). While any tool waits on Atlassian, the same loop serves `/healthz`, so Kubernetes liveness probes time out under normal traffic and healthy pods get restarted. The Jira side already offloads through `run_jira_fetcher_call` in `servers/async_utils.py`; `servers/confluence.py` had no offloading at all.

## Repro

Run the server with `--transport streamable-http`, call any Confluence tool against a slow or unreachable host, and request `/healthz` while the tool is in flight: the health request sits behind the blocking call until it completes. With enough concurrent tool calls the liveness probe times out and the pod restarts.

## Fix

Add `run_confluence_fetcher_call` in `servers/async_utils.py`, mirroring the existing Jira helper: a bounded `anyio.to_thread.run_sync` offload with its own event-loop-local `CapacityLimiter` and a `CONFLUENCE_FETCHER_MAX_WORKERS` knob (default 8). The shared env-parsing and limiter logic is factored into private helpers; the Jira functions keep their exact behavior and signatures.

Every blocking Confluence fetcher call in `servers/confluence.py` (search, pages, comments, labels, attachments including the v2-adapter and raw-session fallbacks in the download paths, templates, restrictions, permissions) now goes through the worker pool.

## Tests

`tests/unit/servers/test_async_utils.py` gains the Confluence mirrors of the Jira suite: env-var parsing for `CONFLUENCE_FETCHER_MAX_WORKERS`, bounded concurrency, single-worker serialization, and limiter isolation across asyncio/Trio loops. The full `test_confluence_server.py` suite passes unchanged (76 tests), confirming handler behavior is preserved when fetcher calls move to worker threads.


AmirF194 (commented):
I ran this against f060a697a0ad0e8800964d5a5f8c33c61ddce344 in a clean container.

confluence.py:2571's delete_attachment tool calls confluence_fetcher.delete_attachment(attachment_id=attachment_id) directly, without run_confluence_fetcher_call. That fetcher method makes a real blocking HTTP call (v2_adapter.delete_attachment, or self.confluence._session.delete(url) in confluence/attachments.py around line 800), so this one handler still stalls the event loop the way #1643 describes. Every other Confluence tool touched by this PR goes through the worker pool; this one does not.

I reproduced the mechanism with this PR's own run_confluence_fetcher_call, substituting a 0.4s sleep for the network call and racing it against three healthz style probes on the same event loop:

python3 repro.py (anyio, imports async_utils.py directly from the PR branch)
unpatched (current delete_attachment handler, no offload):
[('healthz-0', 0.4007), ('healthz-1', 0.4007), ('healthz-2', 0.4007)]
patched (same call wrapped in run_confluence_fetcher_call):
[('healthz-0', 0.0005), ('healthz-1', 0.0005), ('healthz-2', 0.0005)]

Fix would be the same pattern used everywhere else here: await run_confluence_fetcher_call(confluence_fetcher.delete_attachment, attachment_id=attachment_id).

PR #1683, opened 2026-09-17
fix(server): release request-scoped Atlassian sessions to stop fd/CLOSE_WAIT leak

### LINKED ISSUE #1682: [Bug]: HTTP transport leaks a file descriptor and a CLOSE_WAIT socket for every Atlassian fetcher it builds
### Prerequisites

- [x] I have searched the existing issues to make sure this bug has not already been
  reported.
- [x] I have checked the README for relevant information.

### Bug Description

In HTTP mode (`--transport streamable-http`) every Jira/Confluence fetcher built for a
request keeps its `requests.Session` open. Nothing releases the request-scoped
fetchers:

* `JiraClient` / `ConfluenceClient` own a `requests.Session` (created by
  `atlassian-python-api`) and expose no `close()`.
* `_create_and_validate()` builds the fetcher before validating the credentials and
  raises without releasing it when validation fails.
* `UserTokenMiddleware.__call__` never closes the fetchers cached on `request.state`
  when the response is done, and the global-credential fallback fetcher is not cached
  on `request.state` at all.
* The four `stream=True` attachment downloads never close their responses.

Each fetcher therefore keeps a pooled socket. When the peer closes its side (idle
keep-alive timeout on the Atlassian gateway), the client socket stays in `CLOSE_WAIT`
forever because that session is never reused and never closed, so file descriptors grow
linearly with the number of fetchers built (at least one per authenticated request, two
when a request uses both Jira and Confluence).

### Steps to Reproduce

1. Run the HTTP transport with per-request credentials (Bearer/PAT headers) or with
   `ALLOW_GLOBAL_CRED_FALLBACK=true`, behind a gateway that closes idle connections.
2. Send N authenticated MCP requests.
3. Watch `ls /proc/$(pgrep -f mcp-atlassian)/fd | wc -l` and the `CLOSE_WAIT` count grow
   by roughly one per fetcher built.

Measured with the real `UserTokenMiddleware`, a real `JiraClient` (real
`requests.Session`) and a stub gateway that closes the connection after each response,
retaining each fetcher to emulate the production reference:

| requests | file descriptors | CLOSE_WAIT |
| --- | --- | --- |
| 0 (warm) | 15 | 10 |
| 100 | 115 | 110 |
| 200 | 215 | 210 |
| 300 | 315 | 310 |
| 400 | 415 | 410 |
| 500 | 515 | 510 |

### Expected Behavior

Request-scoped sessions and streaming responses are closed once the request is
finished, so the file-descriptor and `CLOSE_WAIT` counts stay flat under sustained
multi-user traffic.

### Actual Behavior

File descriptors and `CLOSE_WAIT` sockets grow without bound while the server runs. In
production this surfaced as roughly 16k `CLOSE_WAIT` sockets and 16k open file
descriptors, ending in `OSError: [Errno 24] Too many open files` and an MCP endpoint
that stopped answering until the fd limit was raised and the process was restarted.

```
$ ls /proc/1/fd | wc -l
16179
$ ss -tan | grep -c CLOSE-WAIT
16080
```

### mcp-atlassian Version

main @ 2e25fb4f

### Installation Method

Docker

### Operating System

Linux (Debian-based container image)

### Python Version

3.12

### Atlassian Instance Type

- Jira Server / Data Center
- Confluence Server / Data Center

### Client Application

Custom MCP HTTP clients (streamable-http), one per user, each sending its own
Bearer/PAT header.

### Additional Context

A fix with regression tests is prepared and will be linked as a pull request. The stdio
transport is not affected: the lifespan creates the global credential configuration
once, and the stdio path builds a fetcher per call without caching it on
`request.state`, so it cannot accumulate request-scoped sessions.

blakeouy:
## Description

Fixes #1682

The HTTP transport builds a Jira/Confluence fetcher per request and caches it on
`request.state`, but nothing ever releases the underlying `requests.Session`, so one
pooled socket per fetcher is left behind as `CLOSE_WAIT` until the process hits its fd
limit. See the linked issue for the reproduction and measurements.

## Changes

- Add `close()` to `JiraClient` and `ConfluenceClient` that closes the underlying
  `requests.Session` connection pool.
- Release an already-created fetcher when `_create_and_validate()` fails during creation
  or credential validation.
- Close the request-scoped Jira/Confluence fetchers in `UserTokenMiddleware` in a
  `finally` block, so normal responses, tool exceptions and client disconnects all
  release them; drop them from `request.state` once closed.
- Cache the global-credential fallback fetcher on `request.state` so the same cleanup
  owns its lifecycle. A request that falls back to the global credentials now builds one
  fetcher and reuses it for the remainder of the request, instead of one per call.
- Close the four Jira/Confluence `stream=True` attachment responses with `try/finally`,
  and assert that those responses are closed on the HTTP-error and write-error paths
  too (the two Jira download failure tests previously stopped at
  `validate_safe_path`'s cwd-containment check before reaching the request, so they now
  really exercise the failure path).
- Regression tests for client close, validation-failure cleanup, middleware cleanup,
  the dependency-layer/`request.state` contract and the streaming attachment responses.

The stdio transport is intentionally out of scope: the lifespan creates the global
credential configuration once, and the stdio path builds a fetcher per call without
caching it on `request.state`, so it cannot accumulate request-scoped sessions.

## Testing

- [ ] Integration tests passed (the `tests/integration` and `tests/e2e` suites need live
  Atlassian credentials and are skipped without them)
- [x] Manual checks performed: `fd/CLOSE_WAIT growth measurement with a stub gateway`

```
full suite:   uv run pytest -q            -> 3937 passed, 243 skipped
unit suite:   uv run pytest -q tests/unit -> 3937 passed, 5 skipped
focused:      jira/confluence client + attachments + servers
              (dependencies, mcp_protocol) -> 302 passed
manual fd:    control (cleanup disabled) -> fd 15 -> 515, CLOSE_WAIT 10 -> 510 over 500
              requests (linear growth of 1 fd + 1 CLOSE_WAIT per fetcher)
              patched (cleanup enabled)  -> fd 5, CLOSE_WAIT 0, flat over 500 requests
```

The manual check drives the real `UserTokenMiddleware` over Starlette's test client; each
request builds a real `JiraClient` (real `requests.Session`) against a stub gateway that
closes the connection after the response, then retains the fetcher to emulate the
long-lived reference seen in production.

## Checklist

- [x] All tests pass locally.


catidegla:
The `finally` placement here is the right shape, and dropping the fetchers from `request.state` once closed is doing more work than it looks like.

Worth being explicit about why, because it is the part a future refactor can quietly undo: `requests.Session.close()` closes the adapters and their connection pools, and that is all it does. It does not touch `session.auth` or `session.headers`. So after `close()` the socket is gone but the Bearer token or PAT is still sitting on the session object, and it stays reachable for as long as anything holds a reference to it.

Which means the credential half of #1682 is fixed by the `del` from `request.state` rather than by `close()`. That is fine, and it works, but it makes the fix load-bearing on there being no *other* reference. Anything that later stashes the fetcher for reuse, memoises it, or captures it in a closure for a background task or a logger would restore the retention without reopening the socket leak, so the symptom everyone measures would stay fixed while the part that matters multi-tenant came back silently.

Two small things that would make that hard to regress:

* A line in the `finally` saying *why* the reference is dropped, not just that it is. Otherwise it reads like tidiness rather than a control.
* `test_mcp_protocol.py` is already asserting the sessions get closed. Asserting the fetcher is gone from `request.state` afterwards would pin the other half, and it is a one-liner next to what is there.

Neither changes the diff's behaviour. It is just that the socket leak is visible in `lsof` and the credential retention is not, so the invariant that fixes the invisible one is the one worth writing down.

blakeouy:
Thanks — this is the part I'd want pinned, and it's the half the diff can silently lose.

Where the release actually happens: `_create_and_validate()` hands the fetcher to `request.state` (`setattr(request.state, spec.state_key, fetcher)`), and the middleware's `finally` pops both fetcher keys, so the Session and the `auth` / `headers` it carries become unreachable with the request. `dict(scope)` is a shallow copy, so the state the middleware pops is the same dict the app wrote into — the assertions check the request's own state, not a copy of it.

Two places your framing reaches further than the diff does:

- The pop covers the fetcher reference. The plaintext token entries the middleware writes to that same dict (`user_atlassian_token`, `user_atlassian_api_token`) are not popped; they go away with the per-request scope, so that isn't a retention path today, but it does mean the "no credential outlives the request" invariant is carried by the scope rather than by the pop.
- On the validation-failure path the raised `ValueError` is the other candidate holder — it is raised inside the `except` block, so it carries `__context__` and the frame that built the config. I have not chased whether anything retains that exception past the handler, so I'm flagging it rather than claiming it's clean. What I can say is that the partially built fetcher is closed before the raise.

On your two suggestions: the state assertion was already in the PR — `test_middleware_closes_request_fetchers` asserts both `close()` calls and `"jira_fetcher" not in scope["state"]` / `"confluence_fetcher" not in scope["state"]`, and `test_middleware_closes_fetcher_cached_on_request_state` covers the fetcher cached by `get_jira_fetcher()`. The why-line is now at the pop site (head 72e4a36d, rebased onto main), phrased around what `close()` does and does not do, because that is the part a future refactor can quietly undo.

One note on `finally` around `_create_and_validate()`: the `except` branch already covers the failure case — a `finally` there would also close the fetcher the request is about to use on the success path. And you are right that the tests pin the pop rather than the absence of every other holder, so a memoisation or closure capture added later would pass them while restoring retention. The comment is where that reasoning lives now, since the thread does not travel with the merge.

catidegla:
The exception path is worth chasing, and I think it's the one place the invariant is genuinely not carried by the scope.

A traceback holds frame objects, and frames hold their locals. So for as long as anything retains that `ValueError`, it retains the frame that built the config, and the credential is a local in that frame. The fetcher being closed before the raise doesn't help: `close()` releases the pool, not `session.auth`.

One correction worth having before anyone reaches for it, because it's the obvious move and it doesn't work: `raise ... from None` sets `__suppress_context__`, which changes how the chain is *displayed*. It does not clear `__context__`, so the reference survives. Same for `__cause__`. If the goal is dropping the frame, that isn't the lever.

What actually decides it is whether anything retains the exception past the handler. Two candidates I'd look at rather than reason about:

* Any APM or error-reporting integration in the deployment. Those capture exceptions with frame locals attached, on purpose, and that is the whole feature. Somewhat pointed for this repo, but a Sentry SDK in the app would hold exactly the frame we're discussing, for as long as the event is buffered.
* Starlette's own `ServerErrorMiddleware` in debug, which renders locals into the response.

Cheapest robust fix is to not have the credential be a local in the raising frame at all — validate in a helper that takes and returns what it needs, so the frame that raises has nothing interesting in it. That sidesteps the whole question rather than answering it per-integration.

And fair on the assertion, `test_middleware_closes_request_fetchers` was already doing it — I read the diff for the close and not for the pop.

blakeouy:
Thanks for the correction on `from None`: it suppresses display of the context, but does not clear `__context__` (`__cause__` is `None` in that form).

Moving validation into a helper alone would not necessarily remove the reachable sensitive state here. If `_create_and_validate()` raises, the traceback also includes `_get_fetcher()`. In the relevant auth branches, that frame retains `request` and the token/config it built; `request` in turn reaches the current state, including the plaintext credential entries. The helper would need to be part of a broader error-boundary design, not just a move of the validation call.

There is also a concrete retention path before any external reporter: FastMCP wraps this `ValueError` in a `ToolError` using `raise ... from e`, so it remains reachable as `ToolError.__cause__` while that error is live. The project declares no APM/error-reporting dependency and this `http_app()` call does not explicitly enable debug, but I have not traced whether the framework or host retains that `ToolError` after handling it.

I agree this is a worthwhile hardening target. This PR's claim should remain limited to deterministic session/response cleanup; a follow-up should define and test the stronger invariant that credentials are not reachable through surviving traceback frames or exception links.

catidegla:
You're right, and my helper suggestion was too shallow. A traceback carries every frame in the chain, not just the raising one, so moving the validation call does nothing while `_get_fetcher()` sits above it holding `request` — and from `request` everything on the state is reachable anyway. Restructuring the raising frame is the wrong lever.

The `ToolError.__cause__` chain is a better find than anything I offered, because it's a retention path inside the stack you already depend on rather than a hypothetical reporter.

Given both of those, the tractable lever looks like the state entries rather than the frames. The middleware's `finally` already pops the two fetcher keys; popping `user_atlassian_token` and `user_atlassian_api_token` in the same place would mean that a surviving traceback reaching `request` finds a state that no longer holds the plaintext. That converts the invariant from "carried by the scope" to "carried by the pop", which is your phrasing and it's the right distinction — a scope is a lifetime, a pop is an action you can assert on.

Being honest about what that does and does not buy, since it's the standard you've been holding this to: it does not close the hole. The frame that read the raw header still has the token as a local, and that frame is in the traceback too. What it does is shrink the reachable set from "anything that can get to `request`" to "the specific frames that touched the raw header" — a much smaller and more auditable surface, and one a follow-up can reason about concretely.

Agreed on scoping. Deterministic session and response cleanup is a claim this diff actually proves, and the traceback-reachability invariant needs its own tests to mean anything. Worth writing the follow-up issue against `ToolError.__cause__` specifically rather than the general shape, since that's the one you can demonstrate today.

blakeouy:
That's a cleaner statement of the trade than mine, and I agree with the distinction: a scope is a lifetime, a pop is an action you can assert on. Two things a follow-up would still have to cover, both of which I checked after reading your comment:

- One remaining path we can point at in this stack is `_get_fetcher()`'s traceback frame. It holds `request`, and the shallow-copied request scope still references the original headers list, so the raw `Authorization` value stays reachable through `request.scope["headers"]`, independently of `state`. Popping credential entries in the middleware `finally` would remove them from the `request.scope["state"]` mapping that `request.state` exposes, but it would not rewrite the headers or clear locals the traceback already captured.
- The set to pop is also larger than the two keys you named: `atlassian_service_headers` holds the per-service personal tokens the same middleware parsed from `x-atlassian-jira-personal-token` / `x-atlassian-confluence-personal-token`.

Agreed on scoping the follow-up to the demonstrated `ToolError.__cause__` path, with its own assertions, rather than framing it as the general retention question. For this PR, I'll keep the claim limited to deterministic session and response cleanup.

catidegla:
The headers finding changes the conclusion, and I think it kills the popping idea as a security measure rather than merely qualifying it.

If the raw `Authorization` value stays reachable through `request.scope["headers"]`, then any frame holding `request` reaches the credential no matter what `state` contains. Popping `state` entries would then buy tidiness and nothing else — the reachable set doesn't shrink, because the transport is holding the same secret one attribute over. Good to know before someone writes that patch believing it closes something.

Which suggests the follow-up should change target rather than scope. "Credentials are not reachable through surviving traceback frames" is not achievable at the application layer for header-borne credentials: the ASGI scope owns the raw headers, the framework hands you `request`, and every frame in the chain that took a request has it. You cannot write a passing test for that invariant without rewriting the scope, which is not this project's to rewrite.

What is achievable is the other half: **nothing retains the exception**. That one is testable, it is bounded by code in this repo plus FastMCP, and you already have the concrete instance in `ToolError.__cause__`. An assertion that the `ToolError` is not referenced after the handler returns is a real test; an assertion that no frame holds a secret is not.

So: reachability is the wrong invariant here, retention is the right one, and your earlier instinct to scope the follow-up to the demonstrated `__cause__` path was the correct call for a better reason than either of us gave at the time.

And thanks for the `atlassian_service_headers` catch — I'd named two keys from reading the middleware and missed that the per-service personal tokens land somewhere else entirely. That is the second time the thing I proposed was narrower than the actual surface, which is itself an argument for the retention framing: the retention set is enumerable, the reachability set apparently isn't.

blakeouy:
Agreed on the reframing — and it is a better target than the one we started with.

For the record, the pop idea is off the table as a control: with the raw `Authorization` still in `request.scope["headers"]`, any frame holding `request` reaches the credential regardless of what `state` contains, so popping entries would tidy the state dict without shrinking the reachable set.

What I would want the follow-up to state is a retention invariant, not a reachability one: under a controlled logging configuration that does not retain `LogRecord.exc_info`, after the tool-call handler returns — test-owned strong references dropped, collection run — weakrefs to both the `ToolError` and a weakrefable custom cause should be dead. The follow-up should first measure whether any repo- or FastMCP-owned object retains either of them; external reporters and deliberately buffering handlers are out of scope, since holding the exception is their job.

Two notes from checking how that would actually be tested:

- In the checked runtime, a plain Python-defined exception subclass and FastMCP 3.4.4's `ToolError` support weakrefs, while the builtin `ValueError` raised in `_create_and_validate()` does not (`TypeError: cannot create weak reference to 'ValueError' object`). So the assertion has to hang off `ToolError` and a weakrefable cause, and both need to be checked — collecting only the `ToolError` would not prove the cause is not retained separately.
- The test must pin a non-retaining handler set.

Still unmeasured on our side: whether anything in or around FastMCP holds the `ToolError` after the handler returns, so I would write the follow-up as "measure first, then harden" rather than claiming it is already released.

For this PR, the claim stays at deterministic session and response cleanup. If I open the follow-up issue, I will write it with the criterion above and these two testing notes.

catidegla:
That's the right criterion, and "measure first, then harden" is the honest order. Two small things from thinking through how it would actually be written:

* **The weakrefable cause should subclass `ValueError`**, not just be any Python-defined exception. Retention is decided by who holds references, not by type — but anything in FastMCP or the app that branches on `except ValueError` or `isinstance(e, ValueError)` would route an arbitrary custom exception down a different path. If one of those branches is where the retaining reference lives, a test using an unrelated class passes while production still leaks. A `ValueError` subclass keeps every type-dispatched path identical and still gets a `__weakref__` slot.

* **`pytest.raises(...) as excinfo` is itself a strong reference.** It's the most natural way to write the trigger, and `excinfo.value` plus `excinfo.tb` keep both the exception and its frames alive for as long as `excinfo` is in scope. So the test needs an explicit `del excinfo` before the collection and the weakref assertion, or it fails for its own reason and looks like a real retention finding. Easy to lose inside "test-owned strong references dropped", so worth naming in the issue.

Good work pinning this down. The weakref-on-builtin detail in particular is the kind of thing that would have cost a confused afternoon.

blakeouy:
Both points taken; the first one is a correction I would otherwise have learned the hard way.

On the cause type: agreed that it has to be a `ValueError` subclass. I had been treating the weakref target as "some exception we can hold weakly", but the test is also exercising the path that produced it, and the stack does branch on the type — `servers/dependencies.py` has `except ValueError` branches in the same request stack, and FastMCP wraps the failure as `ToolError(...) from e`. A custom class could take a different branch, so a passing test would say nothing about the path we actually care about. A `ValueError` subclass keeps every `except` / `isinstance` dispatch identical, and it is still weakrefable: I re-checked locally with CPython, and a Python-defined `ValueError` subclass supports `weakref.ref()`, while the builtin does not.

On `pytest.raises(...) as excinfo`: that is a real trap here, because the most natural way to write the trigger is also the thing that defeats the assertion. So the requirement is that when the weakrefs are checked, nothing in the test frame still binds the exception or its traceback — `del excinfo`, plus dropping any local that kept `excinfo.value` or the traceback alive. Ending the `with` block is not enough on its own.

Nothing here changes what the PR claims; it stays at deterministic session and response cleanup. If/when a follow-up happens, it keeps "measure first, then harden", with these two as requirements on how the measurement is written.

PR #1688, opened 2026-09-20
chore: declare the Forked MCP Server profile

fatherlinux:
This repo is a fork of `sooperset/mcp-atlassian`. Constitution Section III exempts forked projects from the GHCR dual-push — they publish to Quay only, same as `transcriptor`.

But it had **no per-repo constitution at all**, so `parse_header` found no profile and the repo fell through to the default container-image expectations. That made it read as a Section III violation during the RT #1480 fleet sweep when it was compliant the whole time.

Adds `.specify/memory/constitution.md` following the shape Constitution VIII prescribes for this profile, modelled on `transcriptor`:

- **Upstream** — source, MIT license, forked at v0.22.1 (last sync 2026-07-12)
- **Deployment** — port 8021, env file path, credential *names* only
- **Patches** — the distroless Hummingbird Python multi-stage build, the security hardening across attachment/transport/SSRF/authorization/filter/OAuth, proxy routing for trusted transports
- **Registry** — records the Quay-only exemption explicitly so the next audit doesn't re-flag it

`validate-constitution.py` now parses the profile correctly.

**Not fixed here:** the repo still has no `CHANGELOG.md` (Constitution II, universal). That's RT #1484's scope. Worth knowing that this repo is missing from #1484's list for the same reason it was missing from #1480's — it was never cloned on breetai, so a filesystem-based audit couldn't see it.

https://claude.ai/code/session_01NqzfUS8fsJHYeXZRmViorn

### DIFF
--- a/src/mcp_atlassian/confluence/client.py
+++ b/src/mcp_atlassian/confluence/client.py
@@ -41,6 +41,7 @@ class ConfluenceClient:
             MCPAtlassianAuthenticationError: If OAuth authentication fails
         """
         self.config = config or ConfluenceConfig.from_env()
+        transport_url = self.config.url
 
         # Initialize the Confluence client based on auth type
         if self.config.auth_type == "oauth":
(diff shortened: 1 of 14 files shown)


fatherlinux:
Closing — opened against the wrong repository by mistake. This is an internal packaging change for a downstream fork (crunchtools/mcp-atlassian) and was never intended for upstream. Apologies for the noise.

PR #1691, opened 2026-09-21
fix(auth): verify credentials at startup instead of trusting their presence

### LINKED ISSUE #1690: [Bug]: Server starts with unusable credentials and reports empty results as success
## What happens

With a wrong or stale Jira credential, the server starts normally, exposes its
tools, and answers reads with empty collections and `isError: False`. Nothing in
the response or the logs indicates an authentication problem.

Observed with 0.23.1, stdio transport, `TOOLSETS=jira_links`, basic auth against
Jira Cloud:

```
jira_get_link_types -> isError: False, content: []
```

The same site returns four link types once the credential is corrected.

## Why

Two things combine.

**1. Atlassian answers some endpoints anonymously with HTTP 200 and an empty
collection.** Verified with no credentials at all:

```
GET /rest/api/3/issueLinkType   -> 200  {"issueLinkTypes":[]}
GET /rest/api/3/myself          -> 401  Client must be authenticated...
```

So a tool built on such an endpoint cannot distinguish "not authenticated" from
"genuinely empty" by inspecting the response.

**2. The server never checks whether its credentials work.**
`JiraConfig.is_auth_configured()` only checks that credentials are *present*, and
`_validate_authentication()` is gated behind debug logging, with failure caught
and downgraded (`src/mcp_atlassian/jira/client.py`):

```python
# Test authentication during initialization (in debug mode only)
if logger.isEnabledFor(logging.DEBUG) and self.config.auth_type != "external":
    try:
        self._validate_authentication()
    except MCPAtlassianAuthenticationError:
        logger.warning(
            "Authentication validation failed during client initialization - "
            "continuing anyway"
        )
```

In normal (non-debug) operation the check never runs; under debug it runs but
cannot fail the startup. `confluence/client.py` has the identical block.

Worth noting: the `auth_type != "external"` guard already excludes the multi-user
case where credentials arrive per request and startup validation would be
meaningless. So the debug-only gating is not protecting that case - it only
disables the check for single-user deployments, which is where failing fast is
most useful.

## Impact

A user whose token has expired, been revoked, or was minted on the wrong
Atlassian account sees a working-looking server and an empty backlog. In my case
the account email in `JIRA_USERNAME` was wrong; every read returned empty
successes and it took a direct REST comparison to identify the cause.

## Reproduction

1. Configure Jira with a valid URL and a deliberately wrong `JIRA_USERNAME`.
2. Start the server (stdio, `TOOLSETS=jira_links`).
3. Call `jira_get_link_types`.

Expected: an authentication error, or a refusal to start.
Actual: `isError: False` with `[]`.

## Suggested fix

Verify credentials once at server startup, in `main_lifespan`, next to the
existing `is_auth_configured()` check - not in `JiraClient.__init__`, since the
mixins subclass the client and constructor validation would turn every mixin
instantiation into a network call.

Happy to open a PR; I have one prepared.

## Environment

- mcp-atlassian 0.23.1 (behaviour unchanged on `main` at time of writing)
- Jira Cloud, basic auth (API token)
- stdio transport, Python 3.13

kernel-oops:
Fixes #1690

## Problem

The server starts happily with an unusable credential and then reports empty
results as successes. `jira_get_link_types` returns `isError: False` with `[]`
rather than an authentication error.

Two causes combine:

1. Atlassian serves some endpoints to anonymous callers with HTTP 200 and an
   empty collection (`/rest/api/3/issueLinkType` is one), so the response alone
   cannot distinguish "not authenticated" from "genuinely empty".
2. Nothing verifies the credential. `is_auth_configured()` only checks presence,
   and `_validate_authentication()` ran only under debug logging with failure
   caught and downgraded to "continuing anyway".

## Change

`main_lifespan` now verifies credentials once per start, before loading a
service config. A rejected credential logs the reason and leaves that service's
tools unavailable instead of silently serving empty results.
`ATLASSIAN_SKIP_AUTH_VALIDATION=true` restores the previous behaviour.

## Why the server lifespan and not `JiraClient.__init__`

I tried the constructor first. It is the obvious place, but the mixins
(`WorklogMixin`, `TransitionsMixin`, ...) subclass `JiraClient`, so validating
there turns every mixin instantiation into a network call - 984 errors across
the unit suite, most of them fixtures building mixins. The lifespan runs once
per process, covers the case the bug is actually about, and leaves client
construction pure.

## Also included

Several fixtures patched `atlassian.Jira`, but the client binds the name with
`from atlassian import Jira`, so the patch never applied and real clients were
constructed against `test.atlassian.net` and `jira.example.com`. Corrected to
patch the name in the client module. This is a latent bug independent of this
change - it is only visible once anything in construction touches the network -
but leaving it would make the suite's behaviour hard to reason about.

## Open questions for maintainers

- **Failure mode.** This leaves the service's tools unavailable and logs an
  error. A hard startup failure may be preferable: an empty tool list is itself
  a quiet failure, which is the class of problem this PR is about. Happy to
  change it.
- **Confluence probe endpoint.** `ConfluenceClient._validate_authentication()`
  uses `get_all_spaces`, which on a site with public spaces may succeed
  anonymously - the same trap as `issueLinkType`. Left alone here to keep the
  diff focused, but it likely wants a `/user/current`-style probe.

## Testing

- `tests/unit/test_startup_credential_validation.py` - 6 new tests covering
  accepted credentials, rejected credentials, client construction failure, the
  opt-out, and the Confluence path.
- `tests/unit/servers/conftest.py` - autouse fixture so server wiring tests do
  not depend on the network.
- Full unit suite: 3943 passed, 1 failed, 5 skipped. The single failure
  (`test_stdio_homebrew_probe_exits_after_stdin_close`) also fails on an
  unmodified checkout in this environment; baseline was 3937 passed, 1 failed.
- `ruff check` reports no new findings on the changed files; changed files are
  `ruff format` clean.

---

https://claude.ai/code/session_01JNJzjZ4PCbyNoGSEq4Nu2V


src/mcp_atlassian/servers/main.py
```diff
@@ -135,6 +135,45 @@ async def health_check(request: Request) -> JSONResponse:
...
+            "Skipping %s credential validation (ATLASSIAN_SKIP_AUTH_VALIDATION set)",
+            service,
+        )
+        return True
+
+    try:
+        if service == "Jira":
+            from mcp_atlassian.jira import JiraFetcher
+
+            JiraFetcher(config=config)._validate_authentication()
```
AmirF194, line 159: I ran this against d96c3a7 in a clean container, with `atlassian.Jira.myself` patched to count calls and `MCP_VERY_VERBOSE=true` set the way this repo's own CLI startup resolves debug logging (mirrors `mcp_atlassian/__init__.py`'s `-vv`/`MCP_VERY_VERBOSE` path into `setup_logging(DEBUG)`).

`_credentials_are_usable` calls `myself()` twice per Jira service on startup when debug logging is on, not once. `JiraFetcher.__init__` (jira/client.py:236-241) already runs its own `_validate_authentication()` whenever `logger.isEnabledFor(logging.DEBUG)`, and this function's explicit call right after construction (line 159) runs it again. Same shape for Confluence at line 163 (`ConfluenceFetcher.__init__`, confluence/client.py:219-226). Without `MCP_VERY_VERBOSE` only 1 call happens, so the doubling is specific to the debug mode a user reaches for when an auth problem is exactly what they're chasing.

```
$ python repro.py   # MCP_VERY_VERBOSE=true, atlassian.Jira.myself patched
DEBUG - mcp-jira - Testing Jira authentication by retrieving current user info...
INFO - mcp-jira - Jira authentication successful. Current user: Fake User (fake@example.com)
DEBUG - mcp-jira - Testing Jira authentication by retrieving current user info...
INFO - mcp-jira - Jira authentication successful. Current user: Fake User (fake@example.com)
myself() call count: 2
```

`tests/unit/test_startup_credential_validation.py` mocks `JiraFetcher` itself (`patch("mcp_atlassian.jira.JiraFetcher", ...)`), so it never constructs a real client and doesn't see this.

Since this function now validates unconditionally, the client's own debug-mode self-check (jira/client.py:236-241, confluence/client.py:219-226) is redundant from here and could be dropped, saving a second live call against Atlassian on every debug-mode restart.

